Issue #1, opened 2020-05-04
statistics calculations taking too much CPU


jeremyandrews:
Especially when running lengthy/large load tests, the statistics calculations are taking up a large amount of CPU power. Also the sheer amount of data being collected by clients and passing to the parents is slowing things down.

Need to optimize response times similar to how Locust does it, with some rounding, and maintaining running counters.

jeremyandrews:
Fixed by #2

PR #2, opened 2020-05-04
optimize statistics

jeremyandrews:
- remove response_times from unordered HashMap to an ordered BTreeMap
 - round response_times to compress the data
 - maintain running min/max/total counters to minimize CPU overhead
 - rework median() to use grouped BTreeMap response time counters
 - rework calculation of percentiles to use grouped BTreeMap response time counters


jeremyandrews:
Fixes #1

PR #11, opened 2020-05-15
[WIP] wrap gaggle code in compile-time feature

jeremyandrews:
Status: This is mostly working. Which is to say, it properly makes Gaggle's a compile-time only option. The problem is, this means we can no longer use the examples to run a gaggle.

Offhand, it seems this isn't possible, ie https://github.com/rust-lang/cargo/issues/2911


jeremyandrews:
Actually, it's trivial, just pass in `--features gaggle`. I've updated the documentation with instructions.

PR #13, opened 2020-05-16
async

jeremyandrews:
I've rebased #8 onto latest `master` so we can continue this effort.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ license = "Apache-2.0"
 ctrlc = "3.1"
 http = "0.2"
 log = "0.4"
+macro_rules_attribute = "*"
 num_cpus = "1.0"
 num-format = "0.4"
 rand = "0.7"
@@ -25,6 +26,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_cbor = "0.11"
 simplelog = "0.7"
 structopt = "0.3"
+tokio = "0.2.20"
 url = "2.1"
 
 # optional dependencies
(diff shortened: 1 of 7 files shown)


jeremyandrews:
I spent some time trying to wrap `async` in a feature flag, and trying to avoid too much code-duplication in the process. It just kept getting uglier, and didn't even work -- I'm abandoning the effort, just leaving it here for my future reference: https://github.com/tag1consulting/goose/pull/15

Cargo.toml
```diff
@@ -16,6 +16,7 @@ license = "Apache-2.0"
 ctrlc = "3.1"
 http = "0.2"
 log = "0.4"
+macro_rules_attribute = "*"
```
jeremyandrews, line 19: @alsuren  I'm hesitant to add this dependency. Perhaps instead we can implement this as a macro in `goose_codegen`, and generally add other macros in this dependency to improve goose ergonomics? (The code currently in there can be removed).

I don't like that it's a 0.0.1 release, that I don't see any other modules using it on crates.io, and that we have to invoke it before every load test function.

jeremyandrews:
@alsuren I'm working in yet-another-branch (created as I wasn't sure this was going to work), and seeing some good progress. I've been able to remove the macro_rules_attribute macro which I wanted to do, though currently ergonomics are worse not better. If you have time/interest, please see:
https://github.com/tag1consulting/goose/pull/22

jeremyandrews:
Closing in favor of #22

PR #22, opened 2020-05-23
Convert to Async

jeremyandrews:
Re-implement async support without using `macro_rules_attribute`

With help from this thread:
https://www.reddit.com/r/rust/comments/goh2be/how_to_store_an_async_future_to_a_function_with_a/


examples/drupal_loadtest.rs
```diff
@@ -99,25 +103,34 @@ fn drupal_loadtest_front_page(client: &mut GooseClient) {
         },
     }
 }
+fn drupal_loadtest_front_page_async<'r>(client: &'r mut GooseClient) -> Pin<Box<dyn Future<Output = ()> + Send + 'r>> {
+    Box::pin(drupal_loadtest_front_page(client))
+}
```
alsuren, line 108: This is a bit of a baptism of fire into rust async. This is a problem that none of the async web frameworks have a good answer to (as far as I can see).

I've been screwed by this ergonomics problem repeatedly and not found a good answer. I wanted to write GooseTask::new() as a generic function that could accept `drupal_loadtest_front_page` directly, but I couldn't manage it.

This wrapper function seems to be in the privileged position that it is only generic over one thing (the lifetime 'r), and not also over the returned future. (I have previously managed to be generic over returned future in another project ( https://github.com/gotham-rs/gotham/pull/281/files#diff-7ea3539fe8dc99083b2e3fb06a9bcb13R525 ), but not over both at the same time).

I'm wondering if we could cheat by making a macro called task!(), so you do    
```
    GooseAttack::initialize()
        .register_taskset(GooseTaskSet::new("AnonBrowsingUser")
            .register_task(task!(drupal_loadtest_post_comment)
        )
```

It could then generate a closure that does the same boxing job that this wrapper function does, and pass that into GooseTask::new(). This might side-step the problem in the same way.

I might follow up on a few of your threads and see if I can find a solution that avoids macros (or a satisfying explanation of why it's not possible).

examples/drupal_loadtest.rs
```diff
@@ -99,25 +103,34 @@ fn drupal_loadtest_front_page(client: &mut GooseClient) {
         },
     }
 }
+fn drupal_loadtest_front_page_async<'r>(client: &'r mut GooseClient) -> Pin<Box<dyn Future<Output = ()> + Send + 'r>> {
```
alsuren, line 106: In an ideal world, we'd get rid of this wrapper function entirely before merging, but if we can't we probably want to rename it.

I suspect that the suffix `_async` comes from Darksonn/Alice's playground here https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=938b5daebbfe9b86125f8731aeb2302a . There are a bunch of things that are misleadingly named in that playground, so I renamed a few of them: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ddeda28da92f8e0425536af9bf3800a9

examples/drupal_loadtest.rs
jeremyandrews: My apologies -- I have nothing against using Macros. I was just not eager to build on top of `macro_rules_attribute` as it's a 0.0.1 release and has no other software depending on it. I'm actually very eager to add some custom macros to Goose to simplify the creation of load tests, generally speaking.

Your idea of making a task!() macro would be great.

examples/drupal_loadtest.rs
jeremyandrews: Yes, these wrappers would be very confusing to anyone using Goose, I'm 100% committed to at least hiding them behind a macro before we merge this. And you are correct, I copied Alice's naming but agree it's backwards.

jeremyandrews:
@alsuren I've added a simple macro to remove the code-duplication required to wrap each task. I couldn't figure out how to put `task!()` inside `.register_task()` as you'd proposed, any ideas are very welcome. I'd also appreciate any thoughts on how to improve the macro ergonomics.

While it's not perfect, as it gives a 2x boost in performance I think this is usable/mergable.

examples/drupal_loadtest.rs
```diff
@@ -98,26 +98,26 @@ fn drupal_loadtest_front_page(client: &mut GooseClient) {
             client.set_failure();
         },
     }
-}
+} task!(drupal_loadtest_front_page, drupal_loadtest_front_page_task);
```
alsuren, line 101: I feel like rustfmt is going to have something to say about this.

If we are going for something like a python decorator then attribute macros are the answer. It would also be familiar to anyone who has used the async-trait macro. We could write one in this create if you want: it's just a bit of a pain.

Let me have breakfast and then I'll see what I can do about making a version that expands inline.

examples/drupal_loadtest.rs
```diff
@@ -22,48 +22,48 @@
 //! See the License for the specific language governing permissions and
 //! limitations under the License.
 
+use goose::boilerplate;
+boilerplate!();
```
jeremyandrews, line 26: @alsuren What do you think this is just confusing? Or is there a cleaner way to do this? It's just a big block of `use` that has to go at the start of every load test: I figure as long as I document this people can choose which they prefer, this macro or manually declaring everything.

examples/drupal_loadtest.rs
```diff
@@ -22,48 +22,48 @@
...
+
```
alsuren, line 27: The traditional way to write this is `use goose::prelude::*`

examples/drupal_loadtest.rs
jeremyandrews, line 101: @alsuren It would be ideal if we only had to include the name of the wrapped task here, and could somehow auto-derive/extract the name of the function it's wrapping. Any ideas?

jeremyandrews: Thanks!

There's a `goose_codegen/` that was intended for this purpose -- it probably makes sense to use that library, replacing the broken code that was there.

alsuren: Sorry for the delay: I forgot that I had a family video call scheduled ;-)

I was originally thinking about something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=efa9597001e725286a62b5e4ed9746b7

you could potentially move more code into the macro, depending on your appetite for magic.

jeremyandrews: Yes, that's fantastic! Much cleaner/simpler. I committed this change.

I plan to fix up the documentation, comments, comment tests, and use a proper prelude, then I'll get this merged -- I welcome further improvements, but at this point I think it's as usable as pre-async so no need to block it longer. (I'm mostly away from my computer today, so I likely won't get to it until early next week)

I'd be curious what other magic you feel is possible? If we don't have to add any unusual dependencies I'm definitely happy to add macro-magic that simplifies the writing ergonomics of load tests.

src/lib.rs
```diff
@@ -1049,7 +1050,7 @@ impl GooseAttack {
                     info!("waiting for clients to exit");
                 }
                 for client in clients {
-                    let _ = client.join();
+                    let _ = client.await;
```
LionsAd, line 1053: Should you not join all clients together instead of executing them in serial?

examples/drupal_loadtest.rs
```diff
@@ -101,23 +101,23 @@ fn drupal_loadtest_front_page(client: &mut GooseClient) {
 }
 
 /// View a node from 1 to 10,000, created by preptest.sh.
-fn drupal_loadtest_node_page(client: &mut GooseClient) {
+async fn drupal_loadtest_node_page<'r>(client: &'r mut GooseClient) {
```
LionsAd, line 104: Maybe we want to add a macro for that as well - it’s always the same:

- `goose_fn! drupal_loadtest_node_page() {
}`

is what I would suggest.

src/goose.rs
```diff
@@ -134,17 +134,17 @@
 //!     let mut c_task = GooseTask::new(c_task_function);
 //!
 //!     /// A very simple task that simply loads the "a" page.
-//!     fn a_task_function(client: &mut GooseClient) {
+//!     fn a_task_function<'fut>(client: &'fut mut GooseClient) {
```
LionsAd, line 137: These still use 'fut while the rest used 'r — might be confusing

src/lib.rs
```diff
@@ -902,9 +903,9 @@ impl GooseAttack {
...
 
                 // Launch a new client.
-                let client = thread::spawn(move || {
+                let client = tokio::spawn(
                     client::client_main(thread_number, thread_task_set, thread_client, thread_receiver, thread_sender, is_worker)
-                });
+                );
 
                 clients.push(client);
                 self.active_clients += 1;
```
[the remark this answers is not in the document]
LionsAd: Agree - I did a quick check and the other one in worker.rs is outside of the async part.

examples/drupal_loadtest.rs
alsuren: I made a sketch of an attribute style macro in a PR against this branch: https://github.com/jeremyandrews/goose/pull/1
This would make it look like:
```
#[goose_client_callback]
async fn website_login<'r>(client: &'r mut GooseClient) {
    ...
}
```

It may be possible to cut out the arguments and lifetime annotations, but it seems too magical so I'd rather not. Most people will be happy to copy-paste from the examples, so as long as the error messages are easy to understand, I think the verbosity is probably fine.

Honestly, I think of both approaches as temporary workarounds until we can fix the language to be more expressive around fn lifetimes (at which point we can make ::new() accept a bare async fn directly).

examples/drupal_loadtest.rs
```diff
@@ -72,13 +72,13 @@ fn main() {
 }
 
 /// View the front page.
-fn drupal_loadtest_front_page(client: &mut GooseClient) {
-    let response = client.get("/");
+async fn drupal_loadtest_front_page<'r>(client: &'r mut GooseClient) {
```
LionsAd, line 75: It turns out the lifetime is not needed when using the macro to call this:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=804f51aa07a240299de50b5211506a5c

So this can be simplified back to:

```
async fn drupal_loadtest_front_page(client: &mut GooseClient) {
```

examples/drupal_loadtest.rs
LionsAd: Given the lifetime is not needed, this might still be okay, but not as necessary now.

examples/drupal_loadtest.rs
alsuren: Oh right. Sneaky! This is because async fns can do lifetime elision for you.

I guess that settles it in favour of a `task!()`-style macro then. I'm not in the mood for writing lifetime elision in an attribute macro.

alsuren: This is really exciting guys. I love it when a plan comes together :D.

examples/drupal_loadtest.rs
```diff
@@ -22,48 +22,48 @@
...
     GooseAttack::initialize()
         .register_taskset(GooseTaskSet::new("AnonBrowsingUser")
             .set_weight(4)
-            .register_task(GooseTask::new(drupal_loadtest_front_page)
+            .register_task(GooseTask::new(task!(drupal_loadtest_front_page))
                 .set_weight(15)
                 .set_name("(Anon) front page")
             )
-            .register_task(GooseTask::new(drupal_loadtest_node_page)
+            .register_task(GooseTask::new(task!(drupal_loadtest_node_page))
```
LionsAd, line 39: I think there should be a goose_ task! macro that also calls ::new. Then it’s way less redundant and also task! cannot be accidentally forgotten.

examples/drupal_loadtest.rs
jeremyandrews: This is now working wonderfully with the `task!()` macro.

examples/drupal_loadtest.rs
jeremyandrews: The wrapper issue is now resolved thanks to `task!()`.

examples/drupal_loadtest.rs
jeremyandrews: removed nonstandard `boilerplate!()` macro

examples/drupal_loadtest.rs
jeremyandrews: this has been resolved with the `task!()` macro.

jeremyandrews: doc-tests are finally working again locally -- I'll get this merged and then worry about the documentation and further optimizations/cleanups.

src/lib.rs
jeremyandrews: we're now using `futures::future::join_all()`

src/goose.rs
jeremyandrews: I fixed this as part of getting doc-tests working

examples/drupal_loadtest.rs
jeremyandrews: Agreed!! Thanks for all the time and effort!

examples/drupal_loadtest.rs
jeremyandrews: I agree we can improve ergonomics like this. However, I'm not sure we need a `task!()` and a `goose_task!()` -- once we have the outer wrapper I'm not seeing a need for the inner wrapper -- so maybe it should just all be folded into `task!()`?

Let's explore once what we have is merged.

src/lib.rs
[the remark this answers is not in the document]
jeremyandrews: There were three occurrences of this in `lib.rs`, I just pushed a fix for all three.

examples/drupal_loadtest.rs
jeremyandrews: We can add prelude in a followup PR

examples/drupal_loadtest.rs
LionsAd: What I meant was:

`.register_task(task!(drupal_loadtest_simple))`

Or

`.register_task(goose_task!(drupal_loadtest_simple))`

PR #23, opened 2020-05-25
simplify writing load tests with macros

jeremyandrews:
- include `GooseTask::new()` in `task!()`
 - replace `GooseTaskSet::new()` with `taskset!()`

It would also be possible to create an `attack!()` macro that wraps all task sets and tasks, removing the need for `GooseAttack::initialize()` and `.execute()`. Is it worth it?


alsuren:
I would probably stop here, but it depends on your tolerance for magic.

After this patch, task!() and taskset!() feel quite similar to vec![]. They just wrap simple objects/identifiers and return a new object. This doesn't sound too crazy.

(Task is doing a bunch of magic to work around a compiler limitation, but it should eventually be possible to remove this, and make it just call ::new() directly.)

attack!() sounds like it might be doing all sorts of magic under the hood, and I wouldn't trust that I was writing vanilla rust within its walls. It also wraps the execute() call (tying the declarative "here are my task definitions" bit of your to the thing that runs it, in a way that you can't easily separate), so it might make debugging harder.

In general I'm anti-magic and pro-boilerplate. I would even push for making GooseAttack:: execute () implement Future, so you call it as .execute().await, and force all test scripts to use tokio's async main attribute on their main() function. I might be getting ahead of myself there though.

jeremyandrews:
Thanks for the feedback, agreed. I wasn't even certain taskset!() was a good idea as it's not doing anything particularly useful, but in the end I decided I prefer the consistency of task!() and taskset!() together.

I do like the idea of load tests needing to run .execute().await and using tokio's async main attribute. One one hand it's slightly more complex, but it feels more standard. I'll consider this in a followup.

src/goose.rs
```diff
@@ -302,8 +306,18 @@ static APP_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_P
 #[macro_export]
```
alsuren, line 306: This doc string probably wants changing to something like

    task!(foo) conceptually expands to GooseTask::new(foo), but also does some boxing, to work around a limitation in the compiler.

jeremyandrews:
Whoops; I missed these comments. I’ll address in a follow up PR. Thanks!

PR #37, opened 2020-05-27
push per-thread statistics to parent immediately

jeremyandrews:
In the current implementation, each Goose client thread processes requests and maintains statistics, which are pushed up to the client every 15 seconds. 

The intent of this PR is to instead push the raw requests/response details to the parent immediately, offloading the overhead of calculating statistics so clients can run as quickly as possible. This will primarily benefit a Goose Attack if the load test server isn't already using 100% of the CPU.

### Performance:
**Crossbeam**
- With a non-starved CPU we're seeing a ~2% performance improvement
- With a starved CPU we're seeing a ~2-3% performance improvement

**Tokio mpsc**
- With a non-starved CPU we're seeing a ~2.1% performance improvement
- With a starved CPU we're seeing a ~1-2% performance improvement

The raw numbers:

- https://docs.google.com/spreadsheets/d/18qW_ZTWeY5apSSIeGhT298DzVA0IIs2kHxq9wAbkgGE/edit#gid=0

I'm leaning toward using Tokio mpsc as Tokio is already a dependency, and it's definitely preferred you don't starve the load test server for the most reliable numbers.

### DIFF
--- a/examples/drupal_loadtest.rs
+++ b/examples/drupal_loadtest.rs
@@ -213,27 +213,27 @@ async fn drupal_loadtest_post_comment(client: &mut GooseClient) {
                         ("form_id", &form_id[1]),
                         ("op", "Save"),
                     ];
-                    let request_builder =
-                        client.goose_post(&comment_path);
+                    let request_builder = client.goose_post(&comment_path);
                     let response = client.goose_send(request_builder.form(&params)).await;
                     match response {
-                        Ok(r) => {
-                            match r.text().await {
-                                Ok(html) => {
-                                    if !html.contains(&comment_body) {
-                                        eprintln!("no comment showed up after posting to {}", &comment_path);
-                                        client.set_failure(&GooseMethod::POST, &comment_path);
-                                    }
-                                }
-                                Err(e) => {
+                        Ok(r) => match r.text().await {
+                            Ok(html) => {
+                                if !html.contains(&comment_body) {
                                     eprintln!(
-                                        "unexpected error when posting to {}: {}",
-                                        &comment_path, e
+                                        "no comment showed up after posting to {}",
+                                        &comment_path
                                     );
                                     client.set_failure(&GooseMethod::POST, &comment_path);
                                 }
                             }
-                        }
+                            Err(e) => {
+                                eprintln!(
+                                    "unexpected error when posting to {}: {}",
+                                    &comment_path, e
+                                );
+                                client.set_failure(&GooseMethod::POST, &comment_path);
+                            }
+                        },
                         // Goose will catch this error.
                         Err(_) => (),
                     }
--- a/examples/drupal_loadtest.rs
+++ b/examples/drupal_loadtest.rs
@@ -97,12 +97,14 @@ async fn drupal_loadtest_front_page(client: &mut GooseClient) {
             }
             Err(e) => {
                 eprintln!("failed to parse front page: {}", e);
-                client.set_failure();
+                // @TODO
+                //client.set_failure();
             }
         },
         Err(e) => {
             eprintln!("unexpected error when loading front page: {}", e);
-            client.set_failure();
+            // @TODO
+            //client.set_failure();
         }
     }
 }
@@ -131,7 +133,8 @@ async fn drupal_loadtest_login(client: &mut GooseClient) {
                         Some(f) => f,
                         None => {
                             eprintln!("no form_build_id on page: /user page");
-                            client.set_failure();
+                            // @TODO
+                            //client.set_failure();
                             return;
                         }
                     };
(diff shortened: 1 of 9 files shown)


examples/drupal_loadtest.rs
```diff
@@ -97,12 +97,12 @@ async fn drupal_loadtest_front_page(client: &mut GooseClient) {
             }
             Err(e) => {
                 eprintln!("failed to parse front page: {}", e);
-                client.set_failure();
+                client.set_failure(&GooseMethod::GET, "/");
```
jeremyandrews, line 100: @LionsAd Because we're pushing statistics directly to the parent thread, we have no way to know what request we're marking as failed. The previous method for determining this was a bit of an ugly kludge anyway (and not 100% correct). This complicates writing load-tests a little as you now have to specify the Method and Path (or name) that failed -- but, now `set_failure()` and `set_success()` should actually be accurate. I think this is a worthwhile change?

examples/drupal_loadtest.rs
```diff
@@ -164,7 +164,9 @@ async fn drupal_loadtest_login(client: &mut GooseClient) {
 /// Post a comment.
 async fn drupal_loadtest_post_comment(client: &mut GooseClient) {
     let nid: i32 = rand::thread_rng().gen_range(1, 10_000);
-    let response = client.get(format!("/node/{}", &nid).as_str()).await;
+    let node_path = format!("node/{}", &nid);
+    let comment_path = format!("/comment/reply/{}", &nid);
+    let response = client.get(&node_path).await;
```
jeremyandrews, line 169: In a slightly more complicated example, we can reduce the added overhead by calculating `node_path` and `comment_path` one time, then using the variable to set errors.

src/client.rs
```diff
@@ -109,15 +102,8 @@ pub async fn client_main(
...
             while message.is_ok() {
                 match message.unwrap() {
-                    // Sync our state to the parent.
-                    GooseClientCommand::SYNC => {
-                        thread_sender.send(thread_client.clone()).unwrap();
-                        // Reset per-thread counters, as totals have been sent to the parent
-                        thread_client.requests = HashMap::new();
-                    }
-                    // Sync our state to the parent and then exit.
+                    // Time to exit.
```
jeremyandrews, line 105: This reduces some complexity: we no longer need the parent to ask the client threads to send statistics, as now the client thread just sends statistics in real-time.

src/lib.rs
```diff
@@ -1449,137 +1397,10 @@ fn is_valid_host(host: &str) -> bool {
         }
     }
 }
-
```
jeremyandrews, line 1452: We still need this complexity for Gaggles, as in that case each Goose Worker has a parent thread tracking statistics then pushes them to the Gaggle Manager where they have to merged. It's just moving to another file as it's not needed here anymore.

src/goose.rs
```diff
@@ -424,19 +425,6 @@ impl GooseTaskSet {
     }
 }
 
-/// Tracks the current run-mode of a client.
```
jeremyandrews, line 427: I copied this from Locust -- but, it served no useful purpose so I'm removing it as part of this PR.

examples/drupal_loadtest.rs
LionsAd: To be honest I would just encapsulate in the response, so:

`client.set_failure(response);`

might be wanted to still do it manually so having the `<Option>, string` interface as well might be good, but response should have all data on the request.

src/goose.rs
```diff
@@ -462,15 +448,54 @@ pub enum GooseMethod {
...
-        Method::PUT => Some(GooseMethod::PUT),
-        _ => None,
+        Method::DELETE => GooseMethod::DELETE,
+        Method::GET => GooseMethod::GET,
+        Method::HEAD => GooseMethod::HEAD,
+        Method::PATCH => GooseMethod::PATCH,
+        Method::POST => GooseMethod::POST,
+        Method::PUT => GooseMethod::PUT,
+        _ => {
+            error!("unsupported metehod: {}", method);
```
LionsAd, line 460: Nit - typo

examples/drupal_loadtest.rs
jeremyandrews: Ok, I've been hesitant to add a lot of abstraction, but it would clean this up nicely, and work with any client and/or protocol. I like this plan, I will implement in a followup once this PR is merged.

src/goose.rs
jeremyandrews: Thanks! Fixed.

examples/drupal_loadtest.rs
jeremyandrews: Issue created: https://github.com/tag1consulting/goose/issues/40

jeremyandrews:
I had hoped for a more clear winner between Crossbeam and Tokyio sync, but since they're both offering similar performance and we already depend on Tokio, I went with that.

PR #41, opened 2020-05-29
reimplement set_success and set_failure, fix #40

### LINKED ISSUE #40: Abstract client response
Currently if you need to change whether a test succeeded or failed, Goose assumes you're referring to the previous request. (And in PR #37 it requires you to tell it the method and path you're changing).

Per @LionsAd 's feedback in #37, a better solution is for calls to Client to return GooseResponse which includes the Reqwest Response, but also includes the request. This will allow doing things like:
`client.set_failure(response);` much simpler, cleaner, and more reliable.

jeremyandrews:
Requests made by Goose return a custom GooseResponse object which currently contains but the request made, and the response.

Subsequent calls to `set_success` and `set_failure` must include a mutable reference to the request that was made. For example:

```rust
    use goose::prelude::*;

    let mut task = task!(get_function);

    /// A simple task that makes a GET request.
    async fn get_function(client: &mut GooseClient) {
        let mut response = client.get("/404").await;
        match &response.response {
            Ok(r) => {
                // We expect a 404 here.
                if r.status() == 404 {
                    client.set_success(&mut response.request);
                }
            },
            Err(_) => (),
        }
    }
````

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
  - don't shuffle order of weighted task sets when launching clients
  - remove GooseClientMode as it serves no useful purpose
  - push statistics from client threads to parent in real-time
+ - simplify `set_failure` and `set_success` to pass in request
 
 ## 0.7.1 May 26, 2020
  - no longer compile Reqwest blocking client
--- a/examples/drupal_loadtest.rs
+++ b/examples/drupal_loadtest.rs
@@ -85,7 +85,7 @@ async fn drupal_loadtest_front_page(client: &mut GooseClient) {
     let response = client.get("/").await;
 
     // Grab some static assets from the front page.
-    match response {
+    match response.response {
         Ok(r) => match r.text().await {
             Ok(t) => {
                 let re = Regex::new(r#"src="(.*?)""#).unwrap();
@@ -97,12 +97,12 @@ async fn drupal_loadtest_front_page(client: &mut GooseClient) {
             }
             Err(e) => {
                 eprintln!("failed to parse front page: {}", e);
-                client.set_failure(&GooseMethod::GET, "/");
+                client.set_failure(&response);
             }
         },
         Err(e) => {
             eprintln!("unexpected error when loading front page: {}", e);
-            client.set_failure(&GooseMethod::GET, "/");
+            client.set_failure(&response);
         }
     }
 }
@@ -122,7 +122,7 @@ async fn drupal_loadtest_profile_page(client: &mut GooseClient) {
 /// Log in.
 async fn drupal_loadtest_login(client: &mut GooseClient) {
     let response = client.get("/user").await;
-    match response {
+    match response.response {
         Ok(r) => {
             match r.text().await {
                 Ok(html) => {
@@ -131,13 +131,13 @@ async fn drupal_loadtest_login(client: &mut GooseClient) {
                         Some(f) => f,
                         None => {
                             eprintln!("no form_build_id on page: /user page");
-                            client.set_failure(&GooseMethod::GET, "/user");
+                            client.set_failure(&response);
                             return;
                         }
                     };
 
                     // Log the user in.
-                    let uid = rand::thread_rng().gen_range(3, 5_002);
+                    let uid: usize = rand::thread_rng().gen_range(3, 5_002);
                     let username = format!("user{}", uid);
                     let params = [
                         ("name", username.as_str()),
@@ -152,7 +152,7 @@ async fn drupal_loadtest_login(client: &mut GooseClient) {
                 }
                 Err(e) => {
                     eprintln!("unexpected error when loading /user page: {}", e);
-                    client.set_failure(&GooseMethod::GET, "/user");
+                    client.set_failure(&response);
                 }
             }
         }
@@ -167,7 +167,7 @@ async fn drupal_loadtest_post_comment(client: &mut GooseClient) {
     let node_path = format!("node/{}", &nid);
     let comment_path = format!("/comment/reply/{}", &nid);
     let response = client.get(&node_path).await;
-    match response {
+    match response.response {
         Ok(r) => {
             match r.text().await {
                 Ok(html) => {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/goose.rs [hunk 1]
```diff
@@ -625,6 +626,21 @@ impl GooseRequest {
...
+#[derive(Debug)]
+pub struct GooseResponse {
+    pub request: GooseRawRequest,
+    pub response: Result<Response, Error>,
+}
+impl GooseResponse {
+    pub fn new(request: GooseRawRequest, response: Result<Response, Error>) -> Self {
+        GooseResponse {
+            request,
+            response,
```
LionsAd, line 639: Need to take a shared reference to the result of reqwest.

src/goose.rs
```diff
@@ -1109,26 +1124,26 @@ impl GooseClient {
...
         }
 
-        // @TODO: (improve comment) Consume request_name, if set.
+        // Consume request_name if set.
         if self.request_name != None {
             self.request_name = None;
         }
 
-        response
+        GooseResponse::new(raw_request, response)
```
LionsAd, line 1135: Probably need &response as you want a shared ref to it and not move it?

jeremyandrews: we need to return it here, or we lose it

src/goose.rs
```diff
@@ -1221,14 +1267,14 @@ impl GooseClient {
...
-        update_request.update = true;
-
-        self.send_to_parent(update_request);
+    pub fn set_failure(&mut self, request: &GooseRawRequest) {
+        // Only send update if this was previously a success.
+        if request.success {
+            let mut update_request = request.clone();
+            update_request.success = false;
+            update_request.update = true;
+            self.send_to_parent(&update_request);
```
LionsAd, line 1276: Is it correct that this function and send_to_parent are not async? ie send_to_parent can never block?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
  - don't shuffle order of weighted task sets when launching clients
  - remove GooseClientMode as it serves no useful purpose
  - push statistics from client threads to parent in real-time
+ - simplify `set_failure` and `set_success` to pass in request
```
LionsAd, line 7: 👍

examples/drupal_loadtest.rs
```diff
@@ -97,12 +97,12 @@ async fn drupal_loadtest_front_page(client: &mut GooseClient) {
             }
             Err(e) => {
                 eprintln!("failed to parse front page: {}", e);
-                client.set_failure(&GooseMethod::GET, "/");
+                client.set_failure(&response.request);
```
LionsAd, line 100: I am not yet understanding why we cannot get a shared reference for the express purpose of only sending this struct member. That should be safe.

examples/drupal_loadtest.rs
```diff
@@ -131,13 +131,13 @@ async fn drupal_loadtest_login(client: &mut GooseClient) {
...
                             eprintln!("no form_build_id on page: /user page");
-                            client.set_failure(&GooseMethod::GET, "/user");
+                            client.set_failure(&response.request);
                             return;
                         }
                     };
 
                     // Log the user in.
-                    let uid = rand::thread_rng().gen_range(3, 5_002);
+                    let uid: usize = rand::thread_rng().gen_range(3, 5_002);
```
LionsAd, line 140: nit - unrelated change (pedantic I know 🙃)

src/goose.rs
```diff
@@ -1163,31 +1208,32 @@ impl GooseClient {
...
-        let request_name = self.get_request_name(path);
-        let mut update_request = GooseRawRequest::new(method.clone(), &request_name);
-        update_request.success = true;
-        // This is an updaate to a previously recorded statistic.
-        update_request.update = true;
-
-        self.send_to_parent(update_request);
+    pub fn set_success(&mut self, request: &GooseRawRequest) {
+        // Only send update if this was previously not a success.
+        if !request.success {
```
LionsAd, line 1224: Hmmm - would need to update the request as well then afterwards, because else .success() .fail() .success() would not do the right thing ...

src/goose.rs
jeremyandrews: Correct, we're using UnboundSender which does not block:
https://docs.rs/tokio/0.2.21/tokio/sync/mpsc/struct.UnboundedSender.html

examples/drupal_loadtest.rs
jeremyandrews: Whoops, this slipped in when I was testing various solutions -- in one, this was required.

src/goose.rs
jeremyandrews: Excellent point! It also is doing an unnecessary copy because of this. Fixed to require a mutable pointer so we can modify in place, and just resend the modified request object to the parent.

examples/drupal_loadtest.rs
jeremyandrews: The earlier call to `t.text().await` takes partial ownership of the request.request object. If we try and pass the entire GooseRequest object to `set_failure(&request)` we get the following compilation error:

```
move occurs because value has type `reqwest::async_impl::response::Response`, which does not implement the `Copy` trait
```

I believe this is because internally it's calling `to_bytes`, and it works similar to `into_bytes` in Rust's std String:
https://doc.rust-lang.org/std/string/struct.String.html#method.into_bytes

> This consumes the String, so we do not need to copy its contents.

examples/drupal_loadtest.rs
```diff
@@ -97,12 +97,12 @@ async fn drupal_loadtest_front_page(client: &mut GooseClient) {
             }
             Err(e) => {
                 eprintln!("failed to parse front page: {}", e);
-                client.set_failure(&GooseMethod::GET, "/");
+                client.set_failure(&mut response.request);
```
LionsAd, line 100: While I like the long form this feels like an excellent use case for another macro - `!fail(response)`

etc.

Of course it might also hide things.

src/goose.rs
```diff
@@ -1221,14 +1267,13 @@ impl GooseClient {
...
-        update_request.success = false;
-        // This is an updaate to a previously recorded statistic.
-        update_request.update = true;
-
-        self.send_to_parent(update_request);
+    pub fn set_failure(&mut self, request: &mut GooseRawRequest) {
+        // Only send update if this was previously a success.
+        if request.success {
+            request.success = false;
+            request.update = true;
```
LionsAd, line 1274: Unless I miss something the update property gets never reset back - this points to a potential logic error — either it’s unnecessary or at least from set_success / set_failure it’s always true — which means it could save some memory by putting it into an argument and not persisting in the request struct.

PR #43, opened 2020-05-31
make GooseClient immutable

jeremyandrews:
Make GooseClient immutable, and generally cleanup the API for consistency (storing things like the request name in a mutable GooseClient had unintended consequences: for example, we weren't loading static assets but the previous implementation hid this as we'd end up naming the wrong request)

Mutable parts are moved into a global Vector of mutable GooseClientState objects (one per client thread), all wrapped in a RwLock().

The Vector is written to only once at startup, after which we're always just grabbing the lock with a `.read()` minimize the performance impact.

Inside `GooseClientState` are currently three mutable values:

```
struct GooseClientState {
    /// A Reqwest client, wrapped in a Mutex as a read-write copy is always needed to
    /// manage things like sessions and cookies.
    client: Mutex<Client>,
    /// Integer value indicating which sequenced bucket the client is currently running
    /// tasks from.
    weighted_bucket: AtomicUsize,
    /// Integer value indicating which task within the current sequenced bucket is currently
    /// running.
    weighted_bucket_position: AtomicUsize,
}
```

### Performance
 - There's no big performance change from this change
 - Raw results: https://docs.google.com/spreadsheets/d/18qW_ZTWeY5apSSIeGhT298DzVA0IIs2kHxq9wAbkgGE/edit#gid=2009779167

### DIFF
--- a/examples/drupal_loadtest.rs
+++ b/examples/drupal_loadtest.rs
@@ -82,20 +82,26 @@ fn main() {
 
 /// View the front page.
 async fn drupal_loadtest_front_page(client: &GooseClient) {
-    let mut response = client.get("/").await;
+    let mut response = client.get("/", None).await;
 
     // Grab some static assets from the front page.
     match response.response {
         Ok(r) => match r.text().await {
             Ok(t) => {
                 let re = Regex::new(r#"src="(.*?)""#).unwrap();
+                // Collect copy of URLs to run them async
+                let mut urls = Vec::new();
+                let mut futures = Vec::new();
                 for url in re.captures_iter(&t) {
                     if url[1].contains("/misc") || url[1].contains("/themes") {
-                        // @TODO: implement `set_request_name` without mutable client
-                        //let _response = client.set_request_name("static asset").get(&url[1]);
-                        let _response = client.get(&url[1]);
+                        urls.push(url[1].to_string());
                     }
                 }
+                for index in 0..urls.len() {
+                    futures.push(client.get(&urls[index], Some("static asset")));
+                }
+                // Asynchronously load all static assets together.
+                futures::future::join_all(futures).await;
             }
             Err(e) => {
                 eprintln!("failed to parse front page: {}", e);
@@ -112,18 +118,18 @@ async fn drupal_loadtest_front_page(client: &GooseClient) {
 /// View a node from 1 to 10,000, created by preptest.sh.
 async fn drupal_loadtest_node_page(client: &GooseClient) {
     let nid = rand::thread_rng().gen_range(1, 10_000);
-    let _response = client.get(format!("/node/{}", &nid).as_str()).await;
+    let _response = client.get(format!("/node/{}", &nid).as_str(), None).await;
 }
 
 /// View a profile from 2 to 5,001, created by preptest.sh.
 async fn drupal_loadtest_profile_page(client: &GooseClient) {
     let uid = rand::thread_rng().gen_range(2, 5_001);
-    let _response = client.get(format!("/user/{}", &uid).as_str()).await;
+    let _response = client.get(format!("/user/{}", &uid).as_str(), None).await;
 }
 
 /// Log in.
 async fn drupal_loadtest_login(client: &GooseClient) {
-    let mut response = client.get("/user").await;
+    let mut response = client.get("/user", None).await;
     match response.response {
         Ok(r) => {
             match r.text().await {
@@ -149,7 +155,7 @@ async fn drupal_loadtest_login(client: &GooseClient) {
                         ("op", "Log+in"),
                     ];
                     let request_builder = client.goose_post("/user").await;
-                    let _response = client.goose_send(request_builder.form(&params)).await;
+                    let _response = client.goose_send(request_builder.form(&params), None).await;
                     // @TODO: verify that we actually logged in.
                 }
                 Err(e) => {
@@ -168,7 +174,7 @@ async fn drupal_loadtest_post_comment(client: &GooseClient) {
     let nid: i32 = rand::thread_rng().gen_range(1, 10_000);
     let node_path = format!("node/{}", &nid);
     let comment_path = format!("/comment/reply/{}", &nid);
-    let mut response = client.get(&node_path).await;
+    let mut response = client.get(&node_path, None).await;
     match response.response {
         Ok(r) => {
             match r.text().await {
@@ -216,7 +222,7 @@ async fn drupal_loadtest_post_comment(client: &GooseClient) {
                         ("op", "Save"),
                     ];
                     let request_builder = client.goose_post(&comment_path).await;
-                    let mut response = client.goose_send(request_builder.form(&params)).await;
+                    let mut response = client.goose_send(request_builder.form(&params), None).await;
                     match response.response {
                         Ok(r) => match r.text().await {
                             Ok(html) => {
--- a/examples/drupal_loadtest.rs
+++ b/examples/drupal_loadtest.rs
@@ -146,7 +146,7 @@ async fn drupal_loadtest_login(client: &mut GooseClient) {
                         ("form_id", "user_login"),
                         ("op", "Log+in"),
                     ];
-                    let request_builder = client.goose_post("/user");
+                    let request_builder = client.goose_post("/user").await;
                     let _response = client.goose_send(request_builder.form(&params)).await;
                     // @TODO: verify that we actually logged in.
                 }
@@ -213,7 +213,7 @@ async fn drupal_loadtest_post_comment(client: &mut GooseClient) {
                         ("form_id", &form_id[1]),
                         ("op", "Save"),
                     ];
-                    let request_builder = client.goose_post(&comment_path);
+                    let request_builder = client.goose_post(&comment_path).await;
                     let mut response = client.goose_send(request_builder.form(&params)).await;
                     match response.response {
                         Ok(r) => match r.text().await {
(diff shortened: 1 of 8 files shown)


examples/drupal_loadtest.rs
```diff
@@ -81,19 +81,27 @@ fn main() {
 }
 
 /// View the front page.
-async fn drupal_loadtest_front_page(client: &mut GooseClient) {
-    let mut response = client.get("/").await;
+async fn drupal_loadtest_front_page(client: &GooseClient) {
+    let mut response = client.get("/", None).await;
```
jeremyandrews, line 85: @LionsAd Now that `GooseClient` is immutable, I had to add a new way to set per-request names. Ultimately this meant adding a new parameter to all requests, which is an optional name. Is this too unwieldy? (ie, you either pass in `None` if you're not setting a custom request name, or you pass in `Some("request name")` if you are).

Another idea I had was to make two versions of each request type, for example:
`fn get(path)` and `fn get_named(path, name)`. Perhaps this is cleaner, as you don't have to pass in `None` when you're not naming requests (and I expect normally people will _not_ be naming requests).

examples/drupal_loadtest.rs
```diff
@@ -81,19 +81,27 @@ fn main() {
...
                     if url[1].contains("/misc") || url[1].contains("/themes") {
-                        let _response = client.set_request_name("static asset").get(&url[1]);
+                        urls.push(url[1].to_string());
                     }
                 }
+                for index in 0..urls.len() {
+                    futures.push(client.get(&urls[index], Some("static asset")));
+                }
+                // Asynchronously load all static assets together.
+                futures::future::join_all(futures).await;
```
jeremyandrews, line 104: @LionsAd Here's an example of an optimization now possible with an immutable `GooseClient`. We are no longer limited to requesting static assets sequentially.

(In early testing, this means where 10 clients were enough to saturate 100% of our CPU due to blocking, we can now run 10 clients with 85-95% of the CPU) -- this also considerably changes how many logged-in versus logged-out pages are loaded, as we're no longer throttled by blocking.

examples/drupal_loadtest.rs
```diff
@@ -81,19 +81,27 @@ fn main() {
...
+                let mut urls = Vec::new();
+                let mut futures = Vec::new();
                 for url in re.captures_iter(&t) {
                     if url[1].contains("/misc") || url[1].contains("/themes") {
-                        let _response = client.set_request_name("static asset").get(&url[1]);
+                        urls.push(url[1].to_string());
                     }
                 }
+                for index in 0..urls.len() {
+                    futures.push(client.get(&urls[index], Some("static asset")));
```
jeremyandrews, line 101: @LionsAd here's an example of naming a request -- we used to use `set_request_name()`, but this is no longer possible as `client` is immutable. Instead we have to pass in the name with the request.

I'm leaning toward adding `get()` and `get_named()`, `post()` and `post_named`, etc, as I don't like passing in `None` or `Some("foo")`. Instead we'd have `client.get("/path/no/name");` and `client.get_named("/named/path", "foo");`.

examples/drupal_loadtest.rs
```diff
@@ -108,20 +116,20 @@ async fn drupal_loadtest_front_page(client: &mut GooseClient) {
 }
 
 /// View a node from 1 to 10,000, created by preptest.sh.
-async fn drupal_loadtest_node_page(client: &mut GooseClient) {
+async fn drupal_loadtest_node_page(client: &GooseClient) {
```
jeremyandrews, line 119: @LionsAd `mut` is gone -- I still need to explore whether or not we can also get rid of `box::pin` as well when storing these futures.

src/lib.rs
```diff
@@ -332,6 +333,51 @@ const RUNNING_STATS_EVERY: usize = 15;
 /// Constant defining Goose's default port when running a Gaggle.
 const DEFAULT_PORT: &str = "5115";
 
+static APP_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
+
+// Share CLIENT object globally.
```
jeremyandrews, line 338: @LionsAd This is the largest change in this PR -- moving mutable elements from `GooseClient` into a new global mutable per-thread `GooseClientState`

examples/drupal_loadtest.rs
```diff
@@ -81,19 +81,27 @@ fn main() {
...
     match response.response {
         Ok(r) => match r.text().await {
             Ok(t) => {
                 let re = Regex::new(r#"src="(.*?)""#).unwrap();
+                // Collect copy of URLs to run them async
+                let mut urls = Vec::new();
+                let mut futures = Vec::new();
                 for url in re.captures_iter(&t) {
                     if url[1].contains("/misc") || url[1].contains("/themes") {
-                        let _response = client.set_request_name("static asset").get(&url[1]);
```
LionsAd, line 94: Was this a bug with missing await?

jeremyandrews: It seems so? I've not yet dug into why we were still making static requests -- at first glance it seems we shouldn't have been making any. 

I think the bug was masked because of how `.set_request_name()` worked -- the request name would have just been consumed by the _next_ request. This would invalidate some of our earlier numbers as well, as it would suggest that prior to this fix we were not actually loading static elements in this example.

jeremyandrews: I've patched `drupal_loadtest` and am re-running the pre-patch tests:

```rust
diff --git a/examples/drupal_loadtest.rs b/examples/drupal_loadtest.rs
index a34a0c4..fb4e45f 100644
@@ -89,11 +89,16 @@ async fn drupal_loadtest_front_page(client: &mut GooseClient) {
         Ok(r) => match r.text().await {
             Ok(t) => {
                 let re = Regex::new(r#"src="(.*?)""#).unwrap();
+                // Collect copy of URLs to run them async
+                let mut urls = Vec::new();
                 for url in re.captures_iter(&t) {
                     if url[1].contains("/misc") || url[1].contains("/themes") {
-                        let _response = client.set_request_name("static asset").get(&url[1]);
+                        urls.push(url[1].to_string());
                     }
                 }
+                for index in 0..urls.len() {
+                    client.set_request_name("static asset").get(&urls[index]).await;
+                }
             }
             Err(e) => {
```

examples/drupal_loadtest.rs
jeremyandrews: I split `.get()` and `.get_named()` as I find it has better ergonomics, and `.get()` is already just a convenience wrapper. (And the same for `.post()`, `.head()`, and `.delete()`.) This reduces the size of this PR, as it requires fewer changes when writing most load tests.

examples/drupal_loadtest.rs
jeremyandrews: It turns out this "boost" was actually a bug in the test from when we converted it to `async` -- we were never calling `await` for the static assets: but the way we were naming requests this was getting hidden as the wrong requests were getting the `static asset` name. In this PR this is cleaned up, and the name can only apply to the correct request.

This `join_all` actually seems to be a performance regression _except_ in the cases where the client is CPU-starved. We're not optimizing for a CPU-starved load test server, so I removed this change and now we are back to awaiting each static element in sequence.

examples/drupal_loadtest.rs
jeremyandrews: In the current version of this PR, we call `get_named()` to name the request -- as these are helpers, it feels more ergonomic to not have to pass in `None` or `Some("foo")` when not naming or naming a request.

examples/drupal_loadtest.rs
jeremyandrews: I was unable to find a way to get rid of the `box::pin` -- the problem is that `GooseTask`'s are stored in `GooseTaskSets` which are stored in `GooseAttacks` and the only method I've found for storing immutable objects without a `box::pin` has a ripple effect that I've been unable to get working. I think it should be left as a future optimization.

examples/drupal_loadtest.rs
jeremyandrews: Confirmed that this was a bug in the drupal_loadtest example that has been there since we converted to using async.

src/client.rs
```diff
@@ -46,35 +47,46 @@ pub async fn client_main(
                     thread_client.task_request_name = Some(thread_task_name.to_string());
                 }
                 // Invoke the task function.
-                function(&mut thread_client).await;
+                function(&thread_client).await;
```
LionsAd, line 50: Unrelated whitespace change?

examples/drupal_loadtest.rs
```diff
@@ -146,8 +151,8 @@ async fn drupal_loadtest_login(client: &mut GooseClient) {
                         ("form_id", "user_login"),
                         ("op", "Log+in"),
                     ];
-                    let request_builder = client.goose_post("/user");
-                    let _response = client.goose_send(request_builder.form(&params)).await;
+                    let request_builder = client.goose_post("/user").await;
+                    let _response = client.goose_send(request_builder.form(&params), None).await;
```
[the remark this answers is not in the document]
jeremyandrews: Yes, `.get()` and `.get_named` are a helper wrappers, whereas `goose_send()` is not - the wrappers handle the request_name for us in most cases, and you only have to manage a `None` or `Some()` manually if you invoke the `goose_` functions directly.

src/client.rs
jeremyandrews: This is removing the `mut`. When `&` precedes a variable there's no space. (For the most part, white space is managed by `cargo fmt`)

PR #46, opened 2020-06-03
[EXPERIMENTAL] removing locks with unsafe

jeremyandrews:
Rust makes us wrap the global client object in locks to ensure we're using it safely, but in reality our code enforces that each client only modifies its own GooseClientState, so we don't _need_ the locks. This is an experimental branch to see how much overhead the locks actually have.

 - the first commit removed the RWLock
 - the second commit removed the Mutex around the Reqwest Client
 - the third commit removed the Atomics
 - performance tests were run after each commit: there was no measurable performance improvement (if anything, there was a small performance penalty: perhaps the Rust compiler is less-able to optimize when we use unsafe?)
 - raw statistics are on the right side of this tab in this document: https://docs.google.com/spreadsheets/d/18qW_ZTWeY5apSSIeGhT298DzVA0IIs2kHxq9wAbkgGE/edit#gid=2009779167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.7.3-dev
+ - move client out of GooseClient into global GooseClientState
+
 ## 0.7.2 June 1, 2020
  - don't shuffle order of weighted task sets when launching clients
  - remove GooseClientMode as it serves no useful purpose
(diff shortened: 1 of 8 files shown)


jeremyandrews:
Using unsafe to avoid locks was less performant. Closing.

PR #47, opened 2020-06-03
fix #45 run optional test_start_task and test_stop_task

### LINKED ISSUE #45: add support for startup/teardown on_test_start/on_test_stop
In current versions of Locust this is referred to as `on_test_start` and `on_test_stop`:
https://docs.locust.io/en/latest/writing-a-locustfile.html#test-start-and-test-stop-events

In earlier versions of locust this was referred to as `startup` and `teardown`:
https://docs.locust.io/en/0.14.6/writing-a-locustfile.html#setups-teardowns-on-start-and-on-stop

In all cases, this allows for the creation of a task that runs only once, regardless of how many clients or how many workers, at the very start of the test and the very end of the test (versus `on_start` and `on-stop` which runs once-per-client).

We're using `startup` on a client Locust test, and will need this functionality to properly convert the test to Goose.

jeremyandrews:
Required for a client load test, adds ability to define an optional `test_start_task` and/or an optional `test_stop_task`. These tasks are only run one time. If running in a Gaggle, the tasks are run by the Manager process.

### Example
```rust
use goose::prelude::*;

fn main() {
    GooseAttack::initialize()
        .test_start(task!(setup))
        .register_taskset(taskset!("FooTasks")
            .register_task(task!(foo_task))
        )
        .test_stop(task!(teardown))
        .execute();
}

async fn setup(client: &GooseClient) {
    // do stuff to set up the load test ...
}

async fn teardown(client: &GooseClient) {
    // do stuff to tear down the load test ...
}

async fn foo_task(client: &GooseClient) {
    // this is a normal load test task ...
    client.get("/foo");
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -396,6 +396,10 @@ pub struct Socket {}
 /// Internal global state for load test.
 #[derive(Clone)]
 pub struct GooseAttack {
+    /// An optional task to run one time before starting clients and running task sets.
+    test_start_task: Option<GooseTask>,
+    /// An optional task to run one time after clients have finished running task sets.
+    test_stop_task: Option<GooseTask>,
     /// A vector containing one copy of each GooseTaskSet that will run during this load test.
     task_sets: Vec<GooseTaskSet>,
     /// A checksum of the task_sets vector to be sure all workers are running the same load test.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -930,8 +984,23 @@ impl GooseAttack {
...
+
         // Initilize per-client states.
         if !self.configuration.worker {
+            // First run global test_start_task, if defined.
+            match &self.test_start_task {
+                Some(t) => {
+                    info!("running test_start_task");
+                    // Create a one-time-use Client to run the test_start_task.
+                    let client =
+                        GooseClient::new(0, self.host.clone(), None, 0, 0, &self.configuration, 0);
```
LionsAd, line 996: Good that client is immutable 😉

PR #48, opened 2020-06-03
fix #44 dont panic if there are no normal tasks

### LINKED ISSUE #44: goose currently assumes load tests have normal tasks
If you write a load test that only has `on_start` and/or `on_stop` tasks, Goose clients will keep displaying the following panic:
```
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/libcore/slice/mod.rs:2842:10
```

This is because the code currently assumes the TaskSet will have 1 or more normal tasks.

jeremyandrews:
- allow a load-test to be just `on_start` and/or `on_exit` tasks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.7.3-dev
  - move client out of GooseClient into global GooseClientState
  - introduce `test_start_task` and `test_stop_task` allowing global setup and teardown
+ - don't panic if a load test doesn't define any normal tasks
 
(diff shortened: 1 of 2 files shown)


jeremyandrews:
Completely agreed -- in this case, there's a bigger problem in that I've not spent the time to figure out how to actually test the load test itself -- so far tests only just cover some specific logic, but they don't start an actual load test. I need to create an issue and be sure this happens sooner than later.

PR #51, opened 2020-06-04
mock http server and automate test of actual load test

jeremyandrews:
I started with https://docs.rs/wiremock/ but didn't find a way to control which port it listens on: I like the idea of auto-selecting a port, but I didn't have a way to then pass this into the load test functions.

I then switched to https://docs.rs/httptest/ which is simple and has worked great.

Tests are working locally, debugging why they're failing here.

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -2179,7 +2179,6 @@ mod tests {
         assert_eq!(response.request.status_code, Some(http::StatusCode::OK));
     }
 
-    //#[tokio::test]
     #[test]
     #[with_mock_server]
     fn load_test_one_taskset() {
@@ -2196,7 +2195,7 @@ mod tests {
             clients: Some(1),
             hatch_rate: 1,
             run_time: "1".to_string(),
-            no_stats: false,
+            no_stats: true,
             status_codes: false,
             only_summary: false,
             reset_stats: false,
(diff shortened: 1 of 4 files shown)


src/goose.rs
```diff
@@ -1200,10 +1200,7 @@ impl GooseClient {
             match &response {
                 Ok(r) => {
                     let status_code = r.status();
-                    // Only increment status_code_counts if we're displaying the results
-                    if self.config.status_codes {
-                        raw_request.set_status_code(Some(status_code));
-                    }
+                    raw_request.set_status_code(Some(status_code));
```
LionsAd, line 1203: STRICT mode on - the tests PR should only have the changes in tests.

Any other changes should be made outside in an extra PR.

src/goose.rs
```diff
@@ -1562,6 +1561,26 @@ impl Hash for GooseTask {
 mod tests {
```
LionsAd, line 1561: It would probably make sense to put the `mod tests` to a tests.rs?

LionsAd: can be follow-up

src/goose.rs
```diff
@@ -1224,20 +1221,22 @@ impl GooseClient {
...
+                    Ok(_) => (),
+                    Err(e) => {
+                        info!("unable to communicate with parent thread, exiting: {}", e);
+                        std::process::exit(1);
+                    }
+                }
             }
+            // No parent thread is defined when running test_start_task,
+            // test_stop_task, and during testing.
+            None => (),
```
LionsAd, line 1239: nit - non test related changes should be separate PR (and justified also without the test)

src/goose.rs
jeremyandrews: This was required for the response to correctly include the status code. It was a lingering bug from when we used to process statistics in real-time, but now we must always store the status code.

src/goose.rs
jeremyandrews: Currently each file has tests appropriate to that file. If we create a `tests.rs` do we then move ALL tests from all files, or just those from `goose.rs`? My concern is, I assume we will move them all, but then I wonder if this is less-organized, not more?

src/goose.rs
jeremyandrews: This change was required for tests to work (granted it _also_ meant that `test_start_task` and `test_stop_task` wasn't working).

jeremyandrews:
closed in favor if #52 and #53

PR #53, opened 2020-06-04
automate load test testing

jeremyandrews:
- introduce httpmock
 - test normal load test
 - test load test with no normal tasks (only `on_start` and `on_stop` tasks)


src/goose.rs
```diff
@@ -2054,7 +2075,7 @@ mod tests {
         assert_eq!(built_request.method(), &Method::POST);
         assert_eq!(
             built_request.url().as_str(),
-            "http://example.com/path/to/post"
+            "http://127.0.0.1:5000/path/to/post"
```
LionsAd, line 2078: Are you sure you want to hardcode that vs using a constant?

jeremyandrews: When I first implemented I thought this was a limitation of the testing framework, but it looks like I can set the port manually so I’ll expose this:
https://docs.rs/httpmock/0.3.5/httpmock/struct.HttpMockConfig.html

jeremyandrews: This does seem a limitation of the testing framework after all -- I don't see a clean way to expose this at compile time. Is it ugly enough I should explore a different testing framework?

LionsAd: I just meant to be DRY - don’t repeat yourself

Just use a constant for wherever you use this special string.

If you ever need to change it, you can do it in one place

PR #54, opened 2020-06-05
propery initialize temp client for setup/teardown

jeremyandrews:
- fix panic when calling `test_start` or `test_stop` to setup/teardown test

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -356,6 +356,10 @@ impl GooseClientState {
     async fn initialize(clients: usize) {
         // Grab a write lock to initialize state for each client.
         let mut goose_client_state = CLIENT.write().await;
+        // State can be reinitialized for `test_start` or `test_stop`.
+        if !goose_client_state.is_empty() {
+            goose_client_state.clear();
+        }
         for _ in 0..clients {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -994,9 +998,12 @@ impl GooseAttack {
             match &self.test_start_task {
                 Some(t) => {
                     info!("running test_start_task");
+                    // Setup temporary state for our single client.
+                    GooseClientState::initialize(1).await;
                     // Create a one-time-use Client to run the test_start_task.
-                    let client =
+                    let mut client =
                         GooseClient::new(0, self.host.clone(), None, 0, 0, &self.configuration, 0);
+                    client.weighted_clients_index = 0;
```
LionsAd, line 1006: Why can’t the constructor do that? Making it nit for just this feels wrong - also if it’s needs to be mutable probably there should be a setter function for it?

jeremyandrews: This is a special case, we’re not initializing normally. Instead of launching a client thread, we’re running the task in the parent. 

That said, we also special case this for tests, so I suppose there could be a second initializer if it’s worth it.

PR #55, opened 2020-06-05
tests for test_start and test_stop

jeremyandrews:
- requires #54 to pass
 - test 1: run `test_start` to perform load-test setup
 - test 2: run `test_stop` to perform load-test cleanup
 - test 3: run `test_start` and `test_stop`, launch multiple clients, confirm both run correctly one time only

### DIFF
--- a/tests/setup_teardown.rs
+++ b/tests/setup_teardown.rs
@@ -1,0 +1,120 @@
+use httpmock::Method::{GET, POST};
+use httpmock::{mock, with_mock_server};
+
+mod common;
+
+use goose::prelude::*;
+
+const INDEX_PATH: &str = "/";
+const SETUP_PATH: &str = "/setup";
+const TEARDOWN_PATH: &str = "/teardown";
+
+pub async fn setup(client: &GooseClient) {
+    let _response = client.post(SETUP_PATH, "setting up load test").await;
+}
+
+pub async fn teardown(client: &GooseClient) {
+    let _response = client.post(TEARDOWN_PATH, "cleaning up after load test").await;
+}
+
+pub async fn get_index(client: &GooseClient) {
+    let _response = client.get(INDEX_PATH).await;
+}
+
+/// Test test_start alone.
+#[test]
+#[with_mock_server]
+fn test_start() {
+    let mock_setup = mock(POST, SETUP_PATH).return_status(201).create();
+    let mock_teardown = mock(POST, TEARDOWN_PATH).return_status(205).create();
+    let mock_index = mock(GET, INDEX_PATH).return_status(200).create();
+
+    crate::GooseAttack::initialize_with_config(common::build_configuration())
+        .setup()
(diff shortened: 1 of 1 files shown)


tests/setup_teardown.rs
```diff
@@ -0,0 +1,113 @@
...
+        .test_start(task!(setup))
+        .register_taskset(taskset!("LoadTest").register_task(task!(get_index).set_weight(9)))
+        .execute();
+
+    let called_setup = mock_setup.times_called();
+    let called_index = mock_index.times_called();
+    let called_teardown = mock_teardown.times_called();
+
+    // Confirm the load test ran.
+    assert_ne!(called_index, 0);
```
LionsAd, line 45: Comment does not match the assertion.

tests/setup_teardown.rs
```diff
@@ -0,0 +1,113 @@
...
+
+/// Test test_stop alone.
+#[test]
+#[with_mock_server]
+fn test_stop() {
+    let mock_setup = mock(POST, SETUP_PATH).return_status(201).create();
+    let mock_teardown = mock(POST, TEARDOWN_PATH).return_status(205).create();
+    let mock_index = mock(GET, INDEX_PATH).return_status(200).create();
+
+    crate::GooseAttack::initialize_with_config(common::build_configuration())
```
LionsAd, line 62: White space looks weird (but might be the mobile app that displays this incorrectly)

tests/setup_teardown.rs
```diff
@@ -0,0 +1,113 @@
...
+        .register_taskset(taskset!("LoadTest").register_task(task!(get_index).set_weight(9)))
+        .test_stop(task!(teardown))
+        .execute();
+
+    let called_setup = mock_setup.times_called();
+    let called_index = mock_index.times_called();
+    let called_teardown = mock_teardown.times_called();
+
+    // Confirm the load test ran.
+    assert_ne!(called_index, 0);
```
LionsAd, line 106: Comment does not match assertion

LionsAd: Also this is probably a failed test in reality as get_index was never called so load test did not run - but I think it should have been?

tests/setup_teardown.rs
jeremyandrews: If the test doesn’t run, this will be zero. If the test runs it will not be zero. We are confirming the test runs by confirming the mock endpoint served more than 0 requests. 

Comment does match assertion.

tests/setup_teardown.rs
jeremyandrews: I’m using ‘cargo fmt’ to enforce code formatting standards.

tests/setup_teardown.rs
jeremyandrews: I do not understand your question. We’re asserting it’s non-zero which is only true if the test runs. The tests are working.

PR #71, opened 2020-06-11
fix #66 track if we redirected

### LINKED ISSUE #66: handle redirects
Finalize how Goose handles redirects and tracks related information. Decide whether any configuration is Goose-specific, or should happen at the client level (ie, currently w/in Reqwest).

As a followup to #64 `GooseRawRequest` should store the final URL, as well as how many redirects happened to get there.

jeremyandrews:
This sets a boolean to `true` or `false` depending on if there was a redirect, and also stores the final `URL`.

I explored ways to track how many times the request was redirected, and the redirect header(s) used. A custom redirect policy can be defined when building the Reqwest client (https://docs.rs/reqwest/0.10.6/reqwest/redirect/index.html) but I didn't find any way to preserve the redirect information. I explored adding a redirections object within our global CLIENT lazy_static, but we don't have enough information when registering the callback, and the callback can't be async (which we require to unlock the CLIENT object).

Another idea to pursue is to completely disable Reqwests manual handling of redirects, and instead re-implement them in Goose. This may be worth the effort, as otherwise Goose has no way to collect statistics about which redirect headers it ran into, and how many times a given URL was redirected.

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -254,8 +254,7 @@
 
 use http::method::Method;
 use http::StatusCode;
-use reqwest::{ClientBuilder, Error};
-use reqwest::{RequestBuilder, Response};
+use reqwest::{ClientBuilder, Error, RequestBuilder, Response};
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashMap};
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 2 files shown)


jeremyandrews:
Ideally this should have tests. Statistics are only output to stdout, I'm not sure how to verify the contents of stdout from a test.

jeremyandrews:
With further discussion, for now it's sufficient that we detect whether or not a redirect happened, not the status code used or the number of times it happened. Adding a basic test to confirm we respect redirect headers.

PR #73, opened 2020-06-11
fix #72 properly handle changed success or fail

### LINKED ISSUE #72: statistics update from set_failure or set_success are inflating statistics
When calling `set_failure` or `set_success` we flag that we're sending an update, however the parent process is not checking this flag. This means each time `set_failure` or `set_success` is checked, we're incrementing failure or success, but not decrementing the previous increment resulting in inflated statistics.

jeremyandrews:
I've confirmed this works through manual testing, but it really should have a test.

Currently Goose only displays statistics to `stdout`, the test would need to process/validate the output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.7.6-dev
+ - properly subtract previous statistic when handling `set_failure()` and `set_success()`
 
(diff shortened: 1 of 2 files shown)


jeremyandrews:
https://users.rust-lang.org/t/write-to-stdout-stderr-or-file/29739/4

PR #75, opened 2020-06-14
fix #74 optionally make redirects of base_url sticky

### LINKED ISSUE #74: redirect of base_url should be sticky
Some websites use multiple domains to serve traffic, redirecting depending on the user's roll. For this reason, Goose needs to respect a redirect of the base_url and subsequent paths should be built from the redirect domain.

For example, if the base_url (ie `--host`) is set to `foo.example.com` and the load test requests `/login`, thereby loading `http://foo.example.com/login` and this request gets redirected to `http://foo-secure.example.com/`, subsequent requests made by this client need to be against the new `foo-secure.example.com` domain. (Further, if the base_url is again redirected, such as when loading `http://foo-secure.example.com/logout`, the client should again follow for subsequent requests,  perhaps in this case back to `foo.example.com`.)

Load tests can also requests absolute URLs, and if these URLs are redirected it should _not_ affect the base_url of the load test. For example, if `foo.example.com` is the base url, and the load test requests `http://bar.example.com` and gets redirected to `http://other.example.com` subsequent relative requests would still be made against `foo.example.com`.

If the load test requests

jeremyandrews:
This functionality is required for a client load test ported from Locust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.7.6-dev
  - properly subtract previous statistic when handling `set_failure()` and `set_success()`
  - detect and track redirects in `GooseRawRequest`
- - redirect of test base_url are sticky and affect subsequent requests
+ - redirect of GooseClient base_url is sticky and affects subsequent requests
 
 ## 0.7.5 June 10, 2020
  - store actual URL requested in GooseRawRequest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.7.6-dev
  - properly subtract previous statistic when handling `set_failure()` and `set_success()`
  - detect and track redirects in `GooseRawRequest`
+ - redirect of test base_url are sticky and affect subsequent requests
 
 ## 0.7.5 June 10, 2020
  - store actual URL requested in GooseRawRequest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.7.6-dev
  - properly subtract previous statistic when handling `set_failure()` and `set_success()`
  - detect and track redirects in `GooseRawRequest`
- - redirect of GooseClient base_url is sticky and affects subsequent requests
+ - redirect of GooseClient base_url is sticky and affects subsequent requests (disable with `--no-sticky-follow`)
 
 ## 0.7.5 June 10, 2020
  - store actual URL requested in GooseRawRequest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.7.6-dev
  - properly subtract previous statistic when handling `set_failure()` and `set_success()`
  - detect and track redirects in `GooseRawRequest`
- - redirect of GooseClient base_url is sticky and affects subsequent requests (disable with `--no-sticky-follow`)
+ - `--sticky-follow` makes redirect of GooseClient base_url sticky, affecting subsequent requests
 
 ## 0.7.5 June 10, 2020
  - store actual URL requested in GooseRawRequest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,3 +37,4 @@ gaggle = ["nng"]
 
 [dev-dependencies]
 httpmock = "0.3"
+mockito = "0.25"
--- a/README.md
+++ b/README.md
@@ -166,25 +166,25 @@ load tests. For example, pass the `-h` flag to the `simple` example,
 
 ```
 client 0.7.5
-CLI options available when launching a Goose loadtest
+CLI options available when launching a Goose load test
 
 USAGE:
     simple [FLAGS] [OPTIONS]
 
 FLAGS:
-    -h, --help                Prints help information
-    -l, --list                Shows list of all possible Goose tasks and exits
-    -g, --log-level           Log level (-g, -gg, -ggg, etc.)
-        --manager             Enables manager mode
-        --no-hash-check       Ignore worker load test checksum
-        --no-stats            Don't print stats in the console
-        --no-sticky-follow    Prevent load test client from following redirect of base_url with subsequent requests
-        --only-summary        Only prints summary stats
-        --reset-stats         Resets statistics once hatching has been completed
-        --status-codes        Includes status code counts in console stats
-    -V, --version             Prints version information
-    -v, --verbose             Debug level (-v, -vv, -vvv, etc.)
-        --worker              Enables worker mode
+    -h, --help             Prints help information
+    -l, --list             Shows list of all possible Goose tasks and exits
+    -g, --log-level        Log level (-g, -gg, -ggg, etc.)
+        --manager          Enables manager mode
+        --no-hash-check    Ignore worker load test checksum
+        --no-stats         Don't print stats in the console
+        --only-summary     Only prints summary stats
+        --reset-stats      Resets statistics once hatching has been completed
+        --status-codes     Includes status code counts in console stats
+        --sticky-follow    Client follows redirect of base_url with subsequent requests
+    -V, --version          Prints version information
+    -v, --verbose          Debug level (-v, -vv, -vvv, etc.)
+        --worker           Enables worker mode
 
 OPTIONS:
     -c, --clients <clients>                        Number of concurrent Goose users (defaults to available CPUs)
--- a/README.md
+++ b/README.md
@@ -172,18 +172,19 @@ USAGE:
     simple [FLAGS] [OPTIONS]
 
 FLAGS:
-    -h, --help             Prints help information
-    -l, --list             Shows list of all possible Goose tasks and exits
-    -g, --log-level        Log level (-g, -gg, -ggg, etc.)
-        --manager          Enables manager mode
-        --no-hash-check    Ignore worker load test checksum
-        --no-stats         Don't print stats in the console
-        --only-summary     Only prints summary stats
-        --reset-stats      Resets statistics once hatching has been completed
-        --status-codes     Includes status code counts in console stats
-    -V, --version          Prints version information
-    -v, --verbose          Debug level (-v, -vv, -vvv, etc.)
-        --worker           Enables worker mode
+    -h, --help                Prints help information
+    -l, --list                Shows list of all possible Goose tasks and exits
+    -g, --log-level           Log level (-g, -gg, -ggg, etc.)
+        --manager             Enables manager mode
+        --no-hash-check       Ignore worker load test checksum
+        --no-stats            Don't print stats in the console
+        --no-sticky-follow    Prevent load test client from following redirect of base_url with subsequent requests
+        --only-summary        Only prints summary stats
+        --reset-stats         Resets statistics once hatching has been completed
+        --status-codes        Includes status code counts in console stats
+    -V, --version             Prints version information
+    -v, --verbose             Debug level (-v, -vv, -vvv, etc.)
+        --worker              Enables worker mode
 
 OPTIONS:
     -c, --clients <clients>                        Number of concurrent Goose users (defaults to available CPUs)
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -1524,7 +1524,7 @@ impl GooseClient {
     /// Determine if the load test has been redirected, and if so return the redirected
     /// host against which all relative URLs should be loaded.
     pub async fn get_redirected_host(&self) -> Option<String> {
-        let redirected_host = self.redirected_host.write().await;
+        let redirected_host = self.redirected_host.read().await;
         if redirected_host.is_empty() {
             None
         } else {
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -1206,7 +1206,7 @@ impl GooseClient {
                     raw_request.set_final_url(r.url().as_str());
 
                     // Load test client was redirected.
-                    if raw_request.url != raw_request.final_url {
+                    if !self.config.no_sticky_follow && raw_request.url != raw_request.final_url {
                         let base_url = self.base_url.read().await.to_string();
                         // Check if the URL redirected started with the load test base_url.
                         if !raw_request.final_url.starts_with(&base_url) {
@@ -1220,7 +1220,9 @@ impl GooseClient {
                             };
                             info!(
                                 "base_url for client {} redirected from {} to {}",
-                                self.weighted_clients_index + 1, &base_url, &redirected_base_url
+                                self.weighted_clients_index + 1,
+                                &base_url,
+                                &redirected_base_url
                             );
                             self.set_base_url(&redirected_base_url).await;
                         }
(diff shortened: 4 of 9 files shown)


src/goose.rs
```diff
@@ -1425,6 +1468,69 @@ impl GooseClient {
...
+    ///     }
+    /// ```
+    pub async fn set_redirected_host(&self, host: String) {
+        *self.redirected_host.write().await = host;
+    }
+
+    /// Determine if the load test has been redirected, and if so return the redirected
+    /// host against which all relative URLs should be loaded.
+    pub async fn get_redirected_host(&self) -> Option<String> {
+        let redirected_host = self.redirected_host.write().await;
```
LionsAd, line 1527: Write lock for the getter?

jeremyandrews: Heh, whoops! Thanks.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.7.6-dev
  - detect and track redirects in `GooseRawRequest`
+ - redirect of test base_url are sticky and affect subsequent requests
```
LionsAd, line 5: Redirects are sticky OR redirect is sticky

src/goose.rs
```diff
@@ -713,43 +716,61 @@ impl GooseClient {
             weighted_on_start_tasks: Vec::new(),
             weighted_tasks: Vec::new(),
             weighted_on_stop_tasks: Vec::new(),
+            redirected_host: Arc::new(RwLock::new("".to_string())),
```
LionsAd, line 719: I would not store the redirected_host, but the client base_url.

(Which can change over time and is initialized originally)

src/goose.rs
```diff
@@ -713,43 +716,61 @@ impl GooseClient {
...
+    /// paths.
+    ///
+    /// The first of these defined will be returned as the prepended host:
+    ///  1. self.redirected_host
+    ///  2. `--host` (host specified on the command line when running load test)
+    ///  3. `GooseTaskSet.host` (default host defined for the current task set)
+    ///  4. `GooseAttack.host` (default host defined for the current load test)
+    async fn get_base_url(&self) -> Url {
+        // If the base url has been redirected, build the URL with it.
+        if let Some(host) = self.get_redirected_host().await {
```
LionsAd, line 736: You need to call that anyway, so best to store the base_url directly in there.

src/goose.rs
```diff
@@ -713,43 +716,61 @@ impl GooseClient {
...
-            Url::parse(&self.config.host).unwrap()
-        } else {
-            match &self.task_set_host {
-                // Otherwise, if `GooseTaskSet.host` is defined, usee this
-                Some(host) => Url::parse(host).unwrap(),
-                // Otherwise, use global `GooseAttack.host`. `unwrap` okay as host validation was done at startup.
-                None => Url::parse(&self.default_host.clone().unwrap()).unwrap(),
-            }
-        };
+        let base_url = self.get_base_url().await;
```
LionsAd, line 773: 👍

src/goose.rs
```diff
@@ -1222,6 +1243,28 @@ impl GooseClient {
...
                     raw_request.set_status_code(Some(status_code));
                     raw_request.set_final_url(r.url().as_str());
+
+                    // Load test client was redirected.
+                    if raw_request.url != raw_request.final_url {
+                        // Extract the base_url of the request made.
+                        if let Ok(original_url) = Url::parse(&raw_request.url) {
+                            if let Some(original_host) = original_url.host() {
+                                if let Some(test_base_url) = self.get_base_url().await.host() {
+                                    // Check if the base_url of the request made matches the base_url load test.
```
LionsAd, line 1253: I would just check and store if it was a relative request.

src/goose.rs
```diff
@@ -1222,6 +1243,28 @@ impl GooseClient {
...
+                                    if test_base_url == original_host {
+                                        if let Ok(redirected_url) =
+                                            Url::parse(&raw_request.final_url)
+                                        {
+                                            if let Some(host) = redirected_url.host() {
```
LionsAd, line 1258: I feel other rust code is not using Option<> with nested ifs style wise. There feels to be a better way.

jeremyandrews:
> I would change things around a little bit and store the „state“ of base_url directly in the struct and just initialize it once.

I didn't do this initially as it was a much bigger change, but it's a nice cleanup so now it's done.

CHANGELOG.md
jeremyandrews: thanks, grammar now fixed

src/goose.rs
jeremyandrews: This had much larger repercussions throughout the code, but the end effect is a net cleanup; it certainly makes sense to figure out the base_url one time per client instead of each time a URL is requested. Fixed.

src/goose.rs
jeremyandrews: Done

src/goose.rs
jeremyandrews: Indeed; the way I was nesting it was completely ignoring possible errors, and therefor was wrong. Storing the base_url simplifies things, and I re-implemented this section to deal with all possible errors and without such deep nesting.

src/goose.rs
```diff
@@ -1425,6 +1430,85 @@ impl GooseClient {
...
+    ///
+    ///     async fn task_foo(client: &GooseClient) {
+    ///       let _response = client.get("/");
+    ///     }
+    ///
+    ///     async fn task_bar(client: &GooseClient) {
+    ///       // Before this task runs, all requests are being made against
+    ///       // http://foo.exampe.com, after this task runs all subsequent
+    ///       // requests are made against http://bar.example.com/.
+    ///       client.set_base_url(Url::parse("http://bar.example.com/").unwrap());
```
LionsAd, line 1479: 👍 - this is a really nice side effect of this change that this is now possible.

src/goose.rs
```diff
@@ -1222,6 +1204,29 @@ impl GooseClient {
                     }
                     raw_request.set_status_code(Some(status_code));
                     raw_request.set_final_url(r.url().as_str());
+
+                    // Load test client was redirected.
+                    if raw_request.url != raw_request.final_url {
+                        // Extract the base_url of the request made.
+                        let request_base_url = match Url::parse(&raw_request.url) {
+                            Ok(base_url) => base_url.to_string(),
```
LionsAd, line 1212: Why here the .to_sting() and ...

src/goose.rs
```diff
@@ -1222,6 +1204,29 @@ impl GooseClient {
...
+                            Err(e) => {
+                                error!("goose_send failed to parse requested URL: {}", e);
+                                std::process::exit(1);
+                            }
+                        };
+                        // If a request for the base_url was redirected, change base_url.
+                        if request_base_url == self.base_url.read().await.to_string() {
+                            let redirected_base_url = match Url::parse(&raw_request.final_url) {
+                                Ok(base_url) => base_url,
```
LionsAd, line 1221: ... here not .to_string()?

src/goose.rs
```diff
@@ -1222,6 +1204,29 @@ impl GooseClient {
...
+                        // Extract the base_url of the request made.
+                        let request_base_url = match Url::parse(&raw_request.url) {
+                            Ok(base_url) => base_url.to_string(),
+                            Err(e) => {
+                                error!("goose_send failed to parse requested URL: {}", e);
+                                std::process::exit(1);
+                            }
+                        };
+                        // If a request for the base_url was redirected, change base_url.
+                        if request_base_url == self.base_url.read().await.to_string() {
```
LionsAd, line 1219: It might be needed to disable that. i.e. if a test accidentally would redirect to IDMS auth we definitely don't want to load test IDMS. Or let's assume a load test accidentally forwards to google.com:

You would not want to accidentally loadtest google.com and get blamed for DDOS ...

Maybe this should only be done for `a subdomain of the original host`?

Again it is really important to understand how locust does it.

LionsAd:
I like the base_url cleanup :)

src/goose.rs
jeremyandrews: we're using the string to make comparisons (however I'm rewriting this section of the code as it's not properly comparing the base_url)

src/goose.rs
jeremyandrews: we store the base_url as a Url, not as a String.

src/goose.rs
jeremyandrews: Any suggestions on what to call this option? I was going to use `--no-redirect` but it wouldn't actually disable redirects, it would just disable the client following the redirect on subsequent requests. Perhaps `--no-follow` is therefor a better option name?

I consider it the load test writer's responsibility to keep their load test from load testing the wrong domains/paths/whatever.

src/goose.rs
```diff
@@ -1222,6 +1204,27 @@ impl GooseClient {
...
                     raw_request.set_final_url(r.url().as_str());
+
+                    // Load test client was redirected.
+                    if raw_request.url != raw_request.final_url {
+                        let base_url = self.base_url.read().await.to_string();
+                        // Check if the URL redirected started with the load test base_url.
+                        if !raw_request.final_url.starts_with(&base_url) {
+                            let redirected_base_url = match Url::parse(&raw_request.final_url) {
+                                // Use URL to grab base_url, which is everything up to the path.
+                                Ok(base_url) => base_url[..url::Position::BeforePath].to_string(),
```
jeremyandrews, line 1215: @LionsAd I'm looking for the most "correct" way to extract the `base_url` from an arbitrary URL, with the least possible assumptions. The [`url`](https://docs.rs/url) module seems to have an internal concept of a base_url, but it's not exposed directly as far as I can see. My solution (which works in my testing so far) is to convert the path to a `url::Url`, convert that to a Rust slice, and then copy everything from the start of the slice to `url::Position::BeforePath` to a string.

See https://docs.rs/url/*/url/enum.Position.html for the relevant docs.

README.md
```diff
@@ -172,18 +172,19 @@ USAGE:
...
-    -V, --version          Prints version information
-    -v, --verbose          Debug level (-v, -vv, -vvv, etc.)
-        --worker           Enables worker mode
+    -h, --help                Prints help information
+    -l, --list                Shows list of all possible Goose tasks and exits
+    -g, --log-level           Log level (-g, -gg, -ggg, etc.)
+        --manager             Enables manager mode
+        --no-hash-check       Ignore worker load test checksum
+        --no-stats            Don't print stats in the console
+        --no-sticky-follow    Prevent load test client from following redirect of base_url with subsequent requests
```
jeremyandrews, line 181: @nnewton @LionsAd Does this help-text make sense? It's hard to explain in so few words (though ideally the help text would be even shorter)

src/goose.rs
jeremyandrews: I ended up with `--no-sticky-follow` as it seems to be properly descriptive -- feedback and alternate suggestions welcome

README.md
nnewton: I like the concept of "sticky follow" Makes sense to me.

LionsAd: I like it, too

src/goose.rs
jeremyandrews: We don't actually know here if the URL is relative or not, as in a previous step we prepend the load test host. We could add an extra field to track if a URL was relative or absolute, but Rust's built in string handling makes it pretty easy to just check the first part.

README.md
```diff
@@ -166,25 +166,25 @@ load tests. For example, pass the `-h` flag to the `simple` example,
...
+    -h, --help             Prints help information
+    -l, --list             Shows list of all possible Goose tasks and exits
+    -g, --log-level        Log level (-g, -gg, -ggg, etc.)
+        --manager          Enables manager mode
+        --no-hash-check    Ignore worker load test checksum
+        --no-stats         Don't print stats in the console
+        --only-summary     Only prints summary stats
+        --reset-stats      Resets statistics once hatching has been completed
+        --status-codes     Includes status code counts in console stats
+        --sticky-follow    Client follows redirect of base_url with subsequent requests
```
jeremyandrews, line 184: @LionsAd @nnewton it turns out this feature was a misunderstanding on my part, and Locust does NOT do this by default. 

In any case 1) there's good cleanup in this PR, and 2) I believe there's still a use-case -- as such, I'm making this functionality available behind a flag:  only if a load test is started with `--sticky-follow` will the base_url follow a redirect for subsequent requests. As it's not the default it shouldn't trip anyone up during normal testing, and I believe this addresses everyone's concerns.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.7.6-dev
  - detect and track redirects in `GooseRawRequest`
+ - redirect of GooseClient base_url is sticky and affects subsequent requests (disable with `--no-sticky-follow`)
```
LionsAd, line 5: Changelog is wrong - it's a feature now

src/goose.rs
LionsAd: I think that sounds good to me - unless there is a function to extract a base url, I am not sure it makes sense.

Cargo.toml
```diff
@@ -37,3 +37,4 @@ gaggle = ["nng"]
 
 [dev-dependencies]
 httpmock = "0.3"
+mockito = "0.25"
```
jeremyandrews, line 40: @LionsAd This adds more compilation time when building the tests, but currently seems to be the only way to test a redirect -- each mock server runs on a different port. The author of httpmock said he'd explore adding the ability to run multiple servers on different ports, at which time I'll remove mockito again.

src/goose.rs
```diff
@@ -1209,28 +1191,51 @@ impl GooseClient {
             .await;
         let elapsed = started.elapsed();
 
-        // Create a raw request object if we're tracking statistics.
```
jeremyandrews, line 1212: @LionsAd This is the only downside to this patch -- in order to detect a redirect I have to parse the response -- previously we only did this if stats were enabled. But it doesn't make sense for me for `--sticky-follow` to depend on stats, so now we do it always, whether or not stats are enabled.

tests/redirect.rs
```diff
@@ -92,3 +107,114 @@ fn test_redirect() {
     assert!(called_redirect == called_redirect3);
     assert!(called_redirect == called_about);
 }
+
+#[test]
+#[with_mock_server]
+/// Simulate a load test which includes a page with a redirect to another domain
+/// (which in this case is a second mock server running on a different path).
+/// all redirects are correctly followed.
+fn test_domain_redirect() {
```
jeremyandrews, line 116: @LionsAd This test confirms that by default redirects are NOT sticky

tests/redirect.rs
```diff
@@ -92,3 +107,114 @@ fn test_redirect() {
...
+    // server on a different port).
+    mock_index_alt.assert();
+
+    // Confirm the we never loaded the about page on the second domain.
+    mock_about_alt.assert();
+}
+
+#[test]
+#[with_mock_server]
+fn test_sticky_domain_redirect() {
```
jeremyandrews, line 169: @LionsAd This test confirms that with `--sticky-follow` enabled redirects ARE sticky

PR #77, opened 2020-06-23
docs: Fix loadtest example in README

playpauseandstop:
README contains invalid code, which result in `mismatched types` error. Fix this, by providing proper example.


jeremyandrews:
Thanks for the patch!

PR #78, opened 2020-06-26
Rename GooseClient to GooseUser

jeremyandrews:
Prior to 0.8.0 we used `client` to refer to both the HTTP client, and individual user threads. With this PR, `GooseClient` is renamed to `GooseUser`, which results in a lot of name changes in nearly every file in the project.

Going forward:
 - a User is an individual thread running tasks from a given task set
 - a Client is the HTTP client object user's use to make requests

Anyone upgrading from Goose 0.7 to 0.8 will need to update their load tests for this change. For an example on how to upgrade to the new API, review the changes made to `examples/simple.rs` and `examples/drupal_loadtest.rs`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Changelog
 
-## 0.7.6-dev
+## 0.8.0-dev
  - properly subtract previous statistic when handling `set_failure()` and `set_success()`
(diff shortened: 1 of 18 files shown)


jeremyandrews:
re-opening after a rebase misfire

PR #79, opened 2020-06-26
move client and counters out of global into GooseClient

jeremyandrews:
When making GooseClient immutable, I originally moved the mutable elements into a global. This commit reverts that, and instead keeps the GooseClient immutable, using locks to make individual fields mutable where needed.

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -254,16 +254,19 @@
 
 use http::method::Method;
 use http::StatusCode;
-use reqwest::{ClientBuilder, Error, RequestBuilder, Response};
+use reqwest::{Client, ClientBuilder, Error, RequestBuilder, Response};
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashMap};
 use std::hash::{Hash, Hasher};
+use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
 use std::{future::Future, pin::Pin, time::Instant};
 use tokio::sync::{mpsc, Mutex, RwLock};
 use url::Url;
 
-use crate::{GooseConfiguration, CLIENT};
+use crate::GooseConfiguration;
+
+static APP_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
 
 /// task!(foo) expands to GooseTask::new(foo), but also does some boxing to work around a limitation in the compiler.
 #[macro_export]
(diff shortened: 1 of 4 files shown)


src/goose.rs
```diff
@@ -698,24 +707,46 @@ impl GooseClient {
...
-            load_test_hash,
+        match Client::builder()
+            .user_agent(APP_USER_AGENT)
+            .cookie_store(true)
+            .build()
+        {
+            Ok(c) => {
+                GooseClient {
+                    task_sets_index,
+                    client: Arc::new(Mutex::new(c)),
```
LionsAd, line 718: Unrelated to that change - I recently learned that boxing things can make structures smaller and that can under certain circumstances avoid memcpy calls (if the whole structure < 128 bytes). Not sure this applies here though, but it came to mind as reqwest is pretty large - on the other hand there is one more indirection age when accessing the inner structures.

In how rust works it’s even feasible to just create one boxed inner struct and then refer to it like client.0.original_property.

Really just food for thought - the changes here are great.

src/goose.rs
```diff
@@ -981,11 +1012,7 @@ impl GooseClient {
     /// ```
     pub async fn goose_get(&self, path: &str) -> RequestBuilder {
         let url = self.build_url(path).await;
-        CLIENT.read().await[self.weighted_clients_index]
-            .client
-            .lock()
-            .await
-            .get(&url)
+        self.client.lock().await.get(&url)
```
LionsAd, line 1015: Full on win and looks almost the same - proving the optimization to be correct.

src/goose.rs
```diff
@@ -1421,14 +1422,13 @@ impl GooseClient {
     /// }
     /// ```
     pub async fn set_client_builder(&self, builder: ClientBuilder) {
-        let client = match builder.build() {
-            Ok(c) => Mutex::new(c),
+        match builder.build() {
+            Ok(c) => *self.client.lock().await = c,
```
LionsAd, line 1426: Can we use a mutable helper variable here? - this syntax while correct looks a little bit weird.

src/lib.rs
```diff
@@ -727,13 +677,14 @@ impl GooseAttack {
         let config = self.configuration.clone();
         loop {
             for task_sets_index in &weighted_task_sets {
+                let base_url = goose::get_base_url(
+                    Some(config.host.to_string()),
+                    self.task_sets[*task_sets_index].host.clone(),
```
LionsAd, line 682: Why do we use & and then dereference here? More curious as I am trying to imagine how it maps to the generated iterator code here.

jeremyandrews: We use `&` so we don't move `weighted_task_sets` when iterating on it.

If we don't use `&` and then dereference, it won't compile because of:

```
error: use of moved value: `weighted_task_sets`
label: value moved here, in previous iteration of loop
```

src/goose.rs
jeremyandrews: You mean something like this? I personally prefer the existing code.

```
diff --git a/src/goose.rs b/src/goose.rs
index aa497a1..9693a2d 100644
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -1423,7 +1423,10 @@ impl GooseClient {
     /// ```
     pub async fn set_client_builder(&self, builder: ClientBuilder) {
         match builder.build() {
-            Ok(c) => *self.client.lock().await = c,
+            Ok(c) => {
+                let mut client = self.client.lock().await;
+                *client = c;
+            },
             Err(e) => {
                 error!("failed to build client: {}", e);
                 std::process::exit(1);
```

jeremyandrews:
I confirmed there's no performance impact of this change -- any difference is within statistical norms.
https://docs.google.com/spreadsheets/d/17I2e4c9JQhykjhKBtMRXYF89WG2mOAHnfVlY_cOrQ50/edit#gid=0

PR #82, opened 2020-06-29
fix #42 optionally log all requests to json lines formatted log

### LINKED ISSUE #42: write/send statistics
Currently Goose processes all statistics internally, and only shares them as optional running-stats during the test and a final summary at the end. With #37, client threads push raw statistics to the parent, so it's now trivial for the parent to also optionally write these statistics to a log file, or send them to a remote service.

We need to define the format of the logs written/sent. 

Locally, to start, it's enough to write to a log file in a consistent CSV format. Something like:

```
timestamp, method, url, name, response_time, status_code, success, update
```

Remotely, to start, it's enough to send logs in a consistent JSON format.

Currently threads only send `method` and `name`, we'll have to add `url` to `GooseRawRequest` so it can be included in logs.

jeremyandrews:
- write request logs from parent thread when statistics are received
- convert internal representation of status code from `Option<StatusCode>` to simple `u16`
- document log format
- asynchronously log to file
- include time elapsed since start of load test, and client thread id in logged data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.8.1-dev
  - sort stats by method:name to ease comparisons
+ - optionally log all requests in JSON Lines format to file specified with `--log-stats-file=`
 
 ## 0.8.0 June 26, 2020
  - properly subtract previous statistic when handling `set_failure()` and `set_success()`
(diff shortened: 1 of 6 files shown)


jeremyandrews:
> RTBM - Is it simple to also log to stdout if one wanted to do that?

I explored this a little, I think it would require some special handling. For example, I could special case the strings `stdout` and `stderr` mapping them to println!() and eprintln!() instead of logging to file. It feels a little ugly, but I was unable to find a more standard way to do this.

If this is important, likely I shouldn't be using File functions directly, but should instead use a logging library to handle all this.

LionsAd:
You can just open stdout as a standard file stream. As long as you interface against Trait IO that should work.

PR #85, opened 2020-07-01
Make Reqwest use Rustls instead on Nativetls

JordiPolo:
### Background
I have added Goose to [a CLI tool](https://github.com/JordiPolo/minos/releases). I create static binary releases for it. To do so in Linux I use the Musl target and it happens that it is a major PITA to work with openssl and Musl. 

It also happens that Reqwest allows to choose between [Rustls](https://github.com/ctz/rustls) or Nativetls (which in Linux depends on openssl). Rustls is pretty nice, based on Ring which recently passed a security audit with flying colors. 

### Changes
Change default from using nativetls to rustls. 

I've not changed the release version as you may add more changes before the next release.
@jeremyandrews

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ num_cpus = "1.0"
 num-format = "0.4"
 rand = "0.7"
 regex = "1"
-reqwest = { version = "0.10",  default-features = false, features = ["cookies", "json", "rustls-tls"] }
+reqwest = { version = "0.10",  default-features = false, features = ["cookies", "json"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_cbor = "0.11"
 serde_json = "1.0"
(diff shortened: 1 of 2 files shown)


jeremyandrews:
Adding a link to a blog discussing the Ring audit:
http://jbp.io/2020/06/14/rustls-audit.html

And the issue that ultimately prompted the Ring audit:
https://github.com/ctz/rustls/issues/189

I've seen general concerns about Ring in the past, though it also seems to be a rather de-facto building block of Rust-based crypto.

This needs more review -- a few questions:

1. Will this affect Reqwest functionality in any way?
2. Will this affect Reqwest performance in any way?
3. Does the audit address all concerns around Ring?
4. Can we instead expose this in a configurable way, rather than either/or?
5. What are other projects doing? (For example, Tokio has a discussion here: https://github.com/tokio-rs/tokio/issues/1268)

JordiPolo:
@jeremyandrews To make is non-problematic (I believe!) I've added it as an optional feature. The default stays the same so there should be no changes to current code/users of Goose.

jeremyandrews:
Thanks! Merged in #92

ElijahLynn:
If the goal is to statically link the binary, there is also another way by using the "vendor" feature of openssl. 

https://github.com/naftulikay/rust-openssl-static-example/issues/5

PR #86, opened 2020-07-01
add optional debug log

jeremyandrews:
## Load Test Debug Logging

Goose can optionally log details about requests and responses for debug purposes. First, when writing a load test you must invoke `client.log_debug(tag, Option<headers>, Option<body>, Option<request>)`) where `tag` is an arbitrary string to identify where in the load test and/or why debug is being written, `headers` are the HTTP headers returned by the server, `body` is the web page body returned by the server, and `request` is a `GooseRawRequest` object.

For an example on how to correctly use `client.log_debug()`, including how to obtain the response headers and body, see `examples/drupal_loadtest`.

If the load test is run with the `--debug-log-file=foo` command line option, where `foo` is either a relative or an absolute path, Goose will log all debug generated by calls to `client.log_debug()` to this file. Debug is logged in JSON Lines format. For example:

```json
{"body":"<!DOCTYPE html>\n<html>\n  <head>\n    <title>503 Backend fetch failed</title>\n  </head>\n  <body>\n    <h1>Error 503 Backend fetch failed</h1>\n    <p>Backend fetch failed</p>\n    <h3>Guru Meditation:</h3>\n    <p>XID: 923425</p>\n    <hr>\n    <p>Varnish cache server</p>\n  </body>\n</html>\n","header":"{\"date\": \"Wed, 01 Jul 2020 10:27:31 GMT\", \"server\": \"Varnish\", \"content-type\": \"text/html; charset=utf-8\", \"retry-after\": \"5\", \"x-varnish\": \"923424\", \"age\": \"0\", \"via\": \"1.1 varnish (Varnish/6.1)\", \"x-varnish-cache\": \"MISS\", \"x-varnish-cookie\": \"SESSd7e04cba6a8ba148c966860632ef3636=hejsW1mQnnsHlua0AicCjEpUjnCRTkOLubwL33UJXRU\", \"content-length\": \"283\", \"connection\": \"keep-alive\"}","request":{"elapsed":4192,"final_url":"http://local.dev/node/3247","method":"GET","name":"(Auth) comment form","redirected":false,"response_time":8,"status_code":503,"success":false,"update":false,"url":"http://local.dev/node/3247","user":4},"tag":"post_comment: no form_build_id found on node/3247"}
```

If `--debug-log-file=foo` is not specified at run time, nothing will be logged.


README.md
```diff
@@ -342,6 +343,27 @@ elapsed,method,name,url,final_url,redirected,response_time,status_code,success,u
 1294,GET,"/","http://local.dev/","http://local.dev/",false,4,200,true,false,1
 ```
 
+## Load Test Debug Logging
+
+Goose can optionally log details about requests and responses for debug purposes. First, when writing
```
LionsAd, line 348: The "first" is off here, because there is no 2nd.

src/lib.rs
```diff
@@ -1049,6 +1049,18 @@ impl GooseAttack {
...
+        let (all_threads_logger, logger_receiver): (
+            mpsc::UnboundedSender<Option<GooseDebug>>,
+            mpsc::UnboundedReceiver<Option<GooseDebug>>,
+        ) = mpsc::unbounded_channel();
+
+        // Launch a new user.
+        let logger_thread = tokio::spawn(logger::logger_main(
+            self.configuration.clone(),
+            logger_receiver,
+        ));
```
LionsAd, line 1062: Is there some effort to setup this process?

README.md
jeremyandrews: Thanks, fixed. This got rewritten a few times during development, somehow that got left in from an earlier version.

(I was trying to make it clear that two things are necessary: 1) call `log_debug()` from your load test, and 2) start the load test with `--debug-log-file=foo` for debugging logging to happen.)

src/logger.rs
```diff
@@ -0,0 +1,87 @@
...
+        // Wait here until a GooseUser thread sends us an error to log, or all GooseUser threads
+        // close the error log channel.
+        match log_receiver.recv().await {
+            Some(message) => {
+                match message {
+                    Some(goose_debug) => {
+                        match debug_log_file.as_mut() {
+                            Some(file) => {
+                                match file
+                                    .write(format!("{}\n", json!(goose_debug)).as_ref())
```
LionsAd, line 48: Maybe it would be good to make the log-format here configurable as well? Does it already work correctly with gaggles?

src/lib.rs
jeremyandrews: It consumes minimal system resources. But really it should be wrapped in Option<>.

src/logger.rs
jeremyandrews: With the stats log it was semi-trivial to add multiple formats. With the debug log it's more effort, as in particular the header and footer would need special escaping. With `serde_json` this is handled automagically. For any other formats we desired, I'd need to find an equally proficient library.

The other main format we support is CSV, and there are some large shortcomings with the existing `csv` package -- in particular, it's not async (so we'd block during logging), and it takes over ownership of the file (so the underlying code would be more complex to support two different ways of writing).

Each Gaggle Worker would need to create their own local debug log currently. But that's a good point, we should throw an error if someone tries to configure this on the Manager instance as nothing would happen. I'm not sure it's worth adding another socket for sending debug logs upstream, especially considering that web pages may be quite large.

src/lib.rs
jeremyandrews: fixed

src/logger.rs
jeremyandrews: We now throw an error if you try and start Goose as a `--manager` and also enable `--debug-log-file`, instructing you instead to enable it on the workers.

LionsAd: Thx - makes sense. I mainly asked if someone would want rather "GooseClient {}" rust default debug output instead of JSON.

Of course one could always print locally, too.

jeremyandrews: Well, `raw` format (what I called them in the other log) would be semi-trivial to add. And agreed, they _may_ be a little easier to read than json, so maybe there's value in it. I can certainly add this if you feel it's worth it (and it's okay to have a format option without csv).

jeremyandrews: okay, added (together with tests)

PR #95, opened 2020-07-04
fix docs, requests need to be awaited

vemoo:
something that tripped me up at first


jeremyandrews:
Yes, thanks for the fix!

PR #96, opened 2020-07-04
fix panic that happens if no host is provided from cli, even if `GooseAttack::set_host` was used.

vemoo:
I tried to write a test, but I didn't find an easy way, and it's a small fix.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -731,11 +731,10 @@ impl GooseAttack {
         info!("initializing user states...");
         let mut weighted_users = Vec::new();
         let mut user_count = 0;
-        let config = self.configuration.clone();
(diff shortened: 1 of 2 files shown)


jeremyandrews:
Thanks for the fix! I'm traveling this week, but will explore adding a test when I'm back online and then get it merged.

If you want to implement yourself, add a second test to the `one_taskset` integration test, and make two changes in the new test compared to the already existing test:

1) store a mutable `configuration` of `common::build_configuration()` and set `configuration.host` to `""`
2) set the host "manually" when calling `crate::GooseAttack()`... when registering the task set (and set it to what's defined on `common.rs`

PR #97, opened 2020-07-11
optionally throttle requests per second

### LINKED ISSUE #90: Make it possible to rate limit a loadtest to x req/s

jeremyandrews:
Initial implementation of requests/second throttle. Fix #90 

 - rebased on top of #98 and #99 (splitting out rename of outer `response` wrapper to `goose`)
 - return `Result<GooseResponse, Error)` from `goose_send()` to allow a clean and quick shutdown even when the throttle is enabled -- this is an API change, bumping the version from `0.8.y` to `0.9.x`
 - throttle implemented with a modified leaky bucket algorithm, when a `GooseClient` makes a request and the throttle is enabled, they have to put a token in the throttle channel before making a request; a dedicated throttle thread removes tokens freeing up space at the desired rate of throttling
 - support up to a throttle-rate of 1,000,000 requests per second, keeping the minimum sleep time to ~10ms by removing multiple tokens at a time
 - integration test, confirms requests are throttled, and increasing the throttle increases the requests per second accordingly

--

In a followup PR will explore returning `Result<>` from Goose task functions as well, allowing the use of `?` to streamline writing load tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ keywords = ["loadtesting", "performance", "web", "framework", "tool"]
 license = "Apache-2.0"
 
 [dependencies]
-async-std = { version = "1.6", features = ["unstable", "tokio02"] }
 ctrlc = "3.1"
 futures = "0.3"
 http = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "goose"
-version = "0.8.3-dev"
+version = "0.9.0-dev"
 authors = ["Jeremy Andrews <jeremy@tag1consulting.com>"]
 edition = "2018"
 description = "A load testing tool inspired by Locust."
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ keywords = ["loadtesting", "performance", "web", "framework", "tool"]
 license = "Apache-2.0"
 
 [dependencies]
+async-std = { version = "1.6", features = ["unstable", "tokio02"] }
 ctrlc = "3.1"
 futures = "0.3"
 http = "0.2"
--- a/examples/drupal_loadtest.rs
+++ b/examples/drupal_loadtest.rs
@@ -82,42 +82,46 @@ fn main() {
 
 /// View the front page.
 async fn drupal_loadtest_front_page(user: &GooseUser) {
-    if let Ok(mut goose) = user.get("/").await {
-        // Grab some static assets from the front page.
-        match goose.response {
-            Ok(response) => {
-                // Copy the headers so we have them for logging if there are errors.
-                let headers = &response.headers().clone();
-                match response.text().await {
-                    Ok(t) => {
-                        let re = Regex::new(r#"src="(.*?)""#).unwrap();
-                        // Collect copy of URLs to run them async
-                        let mut urls = Vec::new();
-                        for url in re.captures_iter(&t) {
-                            if url[1].contains("/misc") || url[1].contains("/themes") {
-                                urls.push(url[1].to_string());
-                            }
-                        }
-                        for asset in &urls {
-                            let _ = user.get_named(asset, "static asset").await;
+    let mut goose = match user.get("/").await {
+        // Return early if get fails, there's nothing else to do.
+        Err(_) => return,
+        // Otherwise unwrap the Result.
+        Ok(g) => g,
+    };
+
+    match goose.response {
+        Ok(response) => {
+            // Copy the headers so we have them for logging if there are errors.
+            let headers = &response.headers().clone();
+            match response.text().await {
+                Ok(t) => {
+                    let re = Regex::new(r#"src="(.*?)""#).unwrap();
+                    // Collect copy of URLs to run them async
+                    let mut urls = Vec::new();
+                    for url in re.captures_iter(&t) {
+                        if url[1].contains("/misc") || url[1].contains("/themes") {
+                            urls.push(url[1].to_string());
                         }
                     }
-                    Err(e) => {
-                        user.set_failure(&mut goose.request);
-                        let error = format!("front_page: failed to parse pag: {}", e);
-                        // We choose to both log and display errors to stdout.
-                        eprintln!("{}", &error);
-                        user.log_debug(&error, Some(goose.request), Some(&headers), None);
+                    for asset in &urls {
+                        let _ = user.get_named(asset, "static asset").await;
                     }
                 }
+                Err(e) => {
+                    user.set_failure(&mut goose.request);
+                    let error = format!("front_page: failed to parse pag: {}", e);
+                    // We choose to both log and display errors to stdout.
+                    eprintln!("{}", &error);
+                    user.log_debug(&error, Some(goose.request), Some(&headers), None);
+                }
             }
-            Err(e) => {
-                user.set_failure(&mut goose.request);
-                let error = format!("front_page: no response from server: {}", e);
-                // We choose to both log and display errors to stdout.
-                eprintln!("{}", &error);
-                user.log_debug(&error, Some(goose.request), None, None);
-            }
+        }
+        Err(e) => {
+            user.set_failure(&mut goose.request);
+            let error = format!("front_page: no response from server: {}", e);
+            // We choose to both log and display errors to stdout.
+            eprintln!("{}", &error);
+            user.log_debug(&error, Some(goose.request), None, None);
         }
     }
 }
@@ -136,64 +140,68 @@ async fn drupal_loadtest_profile_page(user: &GooseUser) {
 
 /// Log in.
 async fn drupal_loadtest_login(user: &GooseUser) {
-    if let Ok(mut goose) = user.get("/user").await {
-        match goose.response {
-            Ok(response) => {
-                // Copy the headers so we have them for logging if there are errors.
-                let headers = &response.headers().clone();
-                match response.text().await {
-                    Ok(html) => {
-                        let re = Regex::new(r#"name="form_build_id" value=['"](.*?)['"]"#).unwrap();
-                        let form_build_id = match re.captures(&html) {
-                            Some(f) => f,
-                            None => {
-                                user.set_failure(&mut goose.request);
-                                let error = "login: no form_build_id on page: /user page";
-                                // We choose to both log and display errors to stdout.
-                                eprintln!("{}", error);
-                                user.log_debug(
-                                    error,
-                                    Some(goose.request),
-                                    Some(&headers),
-                                    Some(html.clone()),
-                                );
-                                return;
-                            }
-                        };
+    let mut goose = match user.get("/user").await {
+        // Return early if get fails, there's nothing else to do.
+        Err(_) => return,
+        // Otherwise unwrap the Result.
+        Ok(g) => g,
+    };
 
-                        // Log the user in.
-                        let uid: usize = rand::thread_rng().gen_range(3, 5_002);
-                        let username = format!("user{}", uid);
-                        let params = [
-                            ("name", username.as_str()),
-                            ("pass", "12345"),
-                            ("form_build_id", &form_build_id[1]),
-                            ("form_id", "user_login"),
-                            ("op", "Log+in"),
-                        ];
-                        let request_builder = user.goose_post("/user").await;
-                        let _goose = user.goose_send(request_builder.form(&params), None).await;
-                        // @TODO: verify that we actually logged in.
-                    }
-                    Err(e) => {
-                        user.set_failure(&mut goose.request);
-                        let error =
-                            format!("login: unexpected error when loading /user page: {}", e);
-                        // We choose to both log and display errors to stdout.
-                        eprintln!("{}", &error);
-                        user.log_debug(&error, Some(goose.request), Some(&headers), None);
-                    }
+    match goose.response {
+        Ok(response) => {
+            // Copy the headers so we have them for logging if there are errors.
+            let headers = &response.headers().clone();
+            match response.text().await {
+                Ok(html) => {
+                    let re = Regex::new(r#"name="form_build_id" value=['"](.*?)['"]"#).unwrap();
+                    let form_build_id = match re.captures(&html) {
+                        Some(f) => f,
+                        None => {
+                            user.set_failure(&mut goose.request);
+                            let error = "login: no form_build_id on page: /user page";
+                            // We choose to both log and display errors to stdout.
+                            eprintln!("{}", error);
+                            user.log_debug(
+                                error,
+                                Some(goose.request),
+                                Some(&headers),
+                                Some(html.clone()),
+                            );
+                            return;
+                        }
+                    };
+
+                    // Log the user in.
+                    let uid: usize = rand::thread_rng().gen_range(3, 5_002);
+                    let username = format!("user{}", uid);
+                    let params = [
+                        ("name", username.as_str()),
+                        ("pass", "12345"),
+                        ("form_build_id", &form_build_id[1]),
+                        ("form_id", "user_login"),
+                        ("op", "Log+in"),
+                    ];
+                    let request_builder = user.goose_post("/user").await;
+                    let _goose = user.goose_send(request_builder.form(&params), None).await;
+                    // @TODO: verify that we actually logged in.
+                }
+                Err(e) => {
+                    user.set_failure(&mut goose.request);
+                    let error = format!("login: unexpected error when loading /user page: {}", e);
+                    // We choose to both log and display errors to stdout.
+                    eprintln!("{}", &error);
+                    user.log_debug(&error, Some(goose.request), Some(&headers), None);
                 }
             }
-            // Goose will catch this error.
-            Err(e) => {
-                user.log_debug(
-                    format!("login: no response from server: {}", e).as_str(),
-                    None,
-                    None,
-                    None,
-                );
-            }
+        }
+        // Goose will catch this error.
+        Err(e) => {
+            user.log_debug(
+                format!("login: no response from server: {}", e).as_str(),
+                None,
+                None,
+                None,
+            );
         }
     }
 }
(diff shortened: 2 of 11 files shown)


jeremyandrews:
Example of it working:

```
jandrews@piccolo:~/devel/rust/goose$ cargo run --example drupal_loadtest -- --host http://apache.fosciana/ -u10 -r25 -t10s -v --throttle-requests 20
   Compiling goose v0.8.3-dev (/home/jandrews/devel/rust/goose)
    Finished dev [unoptimized + debuginfo] target(s) in 7.73s
     Running `target/debug/examples/drupal_loadtest --host 'http://apache.fosciana/' -u10 -r25 -t10s -v --throttle-requests 20`
08:08:23 [ INFO] Output verbosity level: INFO
08:08:23 [ INFO] Logfile verbosity level: INFO
08:08:23 [ INFO] Writing to log file: goose.log
08:08:23 [ INFO] run_time = 10
08:08:23 [ INFO] global host configured: http://apache.fosciana/
08:08:23 [ INFO] initializing user states...
08:08:23 [ INFO] throttle allowing 1 request every 50ms
   ...
 Name                    | # reqs         | # fails        | req/s  | fail/s
 ----------------------------------------------------------------------------- 
 GET (Anon) front page   | 33             | 0 (0%)         | 2      | 0    
 GET (Anon) node page    | 11             | 0 (0%)         | 0      | 0    
 GET (Anon) user page    | 3              | 0 (0%)         | 0      | 0    
 GET (Auth) front page   | 2              | 0 (0%)         | 0      | 0    
 GET (Auth) login        | 2              | 0 (0%)         | 0      | 0    
 GET static asset        | 210            | 0 (0%)         | 16     | 0    
 POST (Auth) login       | 2              | 0 (0%)         | 0      | 0    
 ------------------------+----------------+----------------+--------+--------- 
 Aggregated              | 263            | 0 (0%)         | 20     | 0    
-------------------------------------------------------------------------------

```

src/throttle.rs
```diff
@@ -0,0 +1,19 @@
...
+
+/// Throttle thread, limits how fast requests can be made across all GooseUser threads.
+pub async fn throttle_main(configuration: GooseConfiguration, throttle_sender: Sender<bool>) {
+    let sleep_duration =
+        time::Duration::from_secs_f32(1.0 / configuration.throttle_requests.unwrap() as f32);
+    info!("throttle allowing 1 request every {:?}", sleep_duration);
+
+    // Loop until parent thread exits, adding one element
+    loop {
+        time::delay_for(sleep_duration).await;
```
LionsAd, line 14: All looks good - except this part.

A Token Bucket Filter is waiting exactly one second and then putting k tokens until the bucket is full.

Right now if no one is consuming tokens the req/s can be way higher than configured.

Maybe a bounded channel with try_send would work?

jeremyandrews: We create a bounded channel here:

```rust
             let (sender, receiver): (Sender<bool>, Receiver<bool>) = channel(throttle_requests);
```

Specifically `throttle_requests` is the value passed in with the `--throttle-requests` flag.

This means if nobody consumes tokens for some time, the bucket will just fill up to this maximum value then `throttle_sender.send(true).await` will block.

The only reason I added a comment about `try_send` is when we auto-tune we'll need to handle a full bucket (ie, increase the number of user threads).

jeremyandrews: Also, we space out the adding of tokens as opposed to filling the bucket every second to avoid bursty traffic.

LionsAd: Ah - I indeed missed the bounded channel argument.

So that part is good.

I wonder about the clock resolution of the wait.

"
Implementers of this algorithm on platforms lacking the clock resolution necessary to add a single token to the bucket every 
1
/
r
1/r seconds may want to consider an alternative formulation. Given the ability to update the token bucket every S milliseconds, the number of tokens to add every S milliseconds = 
(
r
∗
S
)
/
1000
(r*S)/1000.
„

So likely to instead add tokens every 10 ms might be a good compromise.

Cargo.toml
```diff
@@ -12,6 +12,7 @@ keywords = ["loadtesting", "performance", "web", "framework", "tool"]
 license = "Apache-2.0"
 
 [dependencies]
+async-std = { version = "1.6", features = ["unstable", "tokio02"] }
```
JordiPolo, line 15: I believe Tokio also [has bounded channels](https://github.com/tokio-rs/tokio/blob/master/tokio/src/sync/mpsc/bounded.rs). As it is already a dependency, is there any need to bring async-std ?

jeremyandrews:
Unfortunately mpsc stands for Multi-Producer-Single-Consumer, and we're looking for the opposite. See https://github.com/tag1consulting/goose/issues/90#issuecomment-653363182 where I explored some options, and ended up with async-std.

LionsAd:
Maybe we could use a leaky bucket instead then?

It works the same way, but the client threads push tokens into the bucket and block while the bucket is full.

The consumer then removes tokens from the bucket a the given rate.

jeremyandrews:
The only thing I'm not sure how to address w/ the Leaky Bucket method is exiting in an efficient and timely manner. 

In the current implementation, the throttle thread passes `true` to children until the parent exits then (not yet implemented) passes `false` telling GooseClients to cancel. With a leaky bucket, I'm not sure how we'd tell clients to cleanly exit mid-request.

Possibly a slow shutdown isn't a big enough problem to worry about. Removing a big dependency is certainly nice.

Cargo.toml
jeremyandrews: The patch has now been reworked to use an `mpsc channel` implementing a modified leaky bucket algorithm, removing the extra dependency on `async-std`.

src/throttle.rs
jeremyandrews: I've added logic to remove multiple tokens if necessary to keep the sleep >= 10ms. In theory we can support throttled rates up to 1,000,000 requests per second, but likely additionally optimizations would be necessary to truly operate at that rate.

examples/drupal_loadtest.rs
```diff
@@ -85,39 +85,41 @@ async fn drupal_loadtest_front_page(user: &GooseUser) {
...
-            let headers = &r.headers().clone();
-            match r.text().await {
-                Ok(t) => {
-                    let re = Regex::new(r#"src="(.*?)""#).unwrap();
-                    // Collect copy of URLs to run them async
-                    let mut urls = Vec::new();
-                    for url in re.captures_iter(&t) {
-                        if url[1].contains("/misc") || url[1].contains("/themes") {
-                            urls.push(url[1].to_string());
+    if let Some(response) = goose.response {
```
LionsAd, line 88: Maybe an early return (as we know that is the error case to exit on) would help to remove this much indentation changes and so many levels deep in indentation?

While I am not a fan of it necessarily, an Exception usually is thrown in other languages when the client is no longer valid (as the load test should end). This Exception is then caught by the thread runner which cleanly ends the loadtest.

Even without Exceptions an early return should be what is done here - as all cleanup if needed should be in a different task and not here.

jeremyandrews: I'm confused what we're accomplishing. We'd still have to unwrap the `goose.response` to see if it's None or `Some()`. Your goal is simply to minimize indention?

LionsAd: Okay - I've informed myself:

The thing to wrap in an `Result<T, Goose>` is the return of the .get(), etc. functions itself.

As load test ended is an Err on the Goose, not on reqwest.

Then the question mark operator would work as well:

```
    let mut goose = user.get("/").await?;

    // Grab some static assets from the front page.
    match goose.response {
```

is replaced with:

```
    // Early return on error
    let mut goose = match user.get("/").await {
           Err(e) => return Err(e),
           Ok(f) => f,
    };

    // Grab some static assets from the front page.
    match goose.response {
```

That is also way more rust-y :).

Reference:

https://static.rust-lang.org/doc/master/std/result/index.html

jeremyandrews: Here's one way to exit early, but I'm not convinced this is better ...?  The end result is the same, but this is actually a little extra code and the addition of a `return` which feels kind of like using a `goto`.

```rust
diff --git a/examples/drupal_loadtest.rs b/examples/drupal_loadtest.rs
index 1ad89b2..266e97f 100644
@@ -84,42 +84,47 @@ fn main() {
 async fn drupal_loadtest_front_page(user: &GooseUser) {
     let mut goose = user.get("/").await;
 
+    let response = match goose.response {
+        // Unwrap response.
+        Some(r) => (r),
+        // If there was no response, exit.
+        None => return,
+    };
+
     // Grab some static assets from the front page.
-    if let Some(response) = goose.response {
-        match response {
-            Ok(r) => {
-                // Copy the headers so we have them for logging if there are errors.
-                let headers = &r.headers().clone();
-                match r.text().await {
-                    Ok(t) => {
-                        let re = Regex::new(r#"src="(.*?)""#).unwrap();
-                        // Collect copy of URLs to run them async
-                        let mut urls = Vec::new();
-                        for url in re.captures_iter(&t) {
-                            if url[1].contains("/misc") || url[1].contains("/themes") {
-                                urls.push(url[1].to_string());
-                            }
-                        }
-                        for asset in &urls {
-                            user.get_named(asset, "static asset").await;
+    match response {
+        Ok(r) => {
+            // Copy the headers so we have them for logging if there are errors.
+            let headers = &r.headers().clone();
+            match r.text().await {
+                Ok(t) => {
+                    let re = Regex::new(r#"src="(.*?)""#).unwrap();
+                    // Collect copy of URLs to run them async
+                    let mut urls = Vec::new();
+                    for url in re.captures_iter(&t) {
+                        if url[1].contains("/misc") || url[1].contains("/themes") {
+                            urls.push(url[1].to_string());
                         }
                     }
-                    Err(e) => {
-                        user.set_failure(&mut goose.request);
-                        let error = format!("front_page: failed to parse pag: {}", e);
-                        // We choose to both log and display errors to stdout.
-                        eprintln!("{}", &error);
-                        user.log_debug(&error, Some(goose.request), Some(&headers), None);
+                    for asset in &urls {
+                        user.get_named(asset, "static asset").await;
                     }
                 }
+                Err(e) => {
+                    user.set_failure(&mut goose.request);
+                    let error = format!("front_page: failed to parse pag: {}", e);
+                    // We choose to both log and display errors to stdout.
+                    eprintln!("{}", &error);
+                    user.log_debug(&error, Some(goose.request), Some(&headers), None);
+                }
             }
-            Err(e) => {
-                user.set_failure(&mut goose.request);
-                let error = format!("front_page: no response from server: {}", e);
-                // We choose to both log and display errors to stdout.
-                eprintln!("{}", &error);
-                user.log_debug(&error, Some(goose.request), None, None);
-            }
+        }
+        Err(e) => {
+            user.set_failure(&mut goose.request);
+            let error = format!("front_page: no response from server: {}", e);
+            // We choose to both log and display errors to stdout.
+            eprintln!("{}", &error);
+            user.log_debug(&error, Some(goose.request), None, None);
         }
     }
 }
```

LionsAd: And yes - we need to change our function signature again for the load test functions, but it's cleaner to return a result anyway.

Even though we need to add a Ok(()) at the end, that's still cleaner than these if chains.

Also while I am unclear how that works in practice, it should be possible to also return an error from Reqwest early that way - though we probably need to provide an `into trait`, which converts one type in another.

jeremyandrews: The other option is to change the task function signature, requiring them to return a [result](https://doc.rust-lang.org/std/result/).

I'm onboard with the general idea, but it's a much larger change and would need to be done in a different PR. It will need further thought/discussion/architecture to decide how Goose handles errors. (ie, do we increment a statistic somewhere? is it a different statistic depending on the error? Can we auto-magically implement `set_failure()` and `set_success()`?) I'd prefer to explore this as a possible followup.

jeremyandrews: Ok, reworked to return a `Result` instead of making `goose.response` optional.

This begs a followup to let task functions return a Result as well, but that will happen in a followup.

examples/drupal_loadtest.rs
```diff
@@ -82,42 +82,42 @@ fn main() {
...
-            let headers = &r.headers().clone();
-            match r.text().await {
-                Ok(t) => {
-                    let re = Regex::new(r#"src="(.*?)""#).unwrap();
-                    // Collect copy of URLs to run them async
-                    let mut urls = Vec::new();
-                    for url in re.captures_iter(&t) {
-                        if url[1].contains("/misc") || url[1].contains("/themes") {
-                            urls.push(url[1].to_string());
+    if let Ok(mut goose) = user.get("/").await {
```
LionsAd, line 85: Let's use match.+ early return and not if let ... - please.

That makes the ? transition more natural.

src/throttle.rs
```diff
@@ -0,0 +1,65 @@
...
+/// tokens are removed thereby throttling how fast requests can be made. It is a variation
+/// on the leaky bucket algorithm: instead of leaking the overflow we asynchronously block.
+pub async fn throttle_main(
+    configuration: GooseConfiguration,
+    mut throttle_receiver: Receiver<bool>,
+    mut parent_receiver: Receiver<bool>,
+) {
+    // Use microseconds to allow configurations up to 1,000,000 requests per second.
+    let mut sleep_duration =
+        time::Duration::from_micros(1_000_000 / configuration.throttle_requests.unwrap() as u64);
```
JordiPolo, line 18: What happens if I `--throttle_requests 0` ?

src/lib.rs
```diff
@@ -1096,6 +1109,40 @@ impl GooseAttack {
             all_threads_logger = None;
         }
 
+        let throttle_receiver;
```
JordiPolo, line 1112: would it be possible for this new code to be in a different method that returns `all_threads_throttle` , `parent_to_throttle_tx` ? This one method here is getting large

src/throttle.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use crate::GooseConfiguration;
+
+/// This throttle thread limits the maximum number of requests that can be made across
+/// all GooseUser threads. When enabled, GooseUser threads must add a token to the
+/// bounded channel before making a request, and this thread limits how frequently
+/// tokens are removed thereby throttling how fast requests can be made. It is a variation
+/// on the leaky bucket algorithm: instead of leaking the overflow we asynchronously block.
+pub async fn throttle_main(
+    configuration: GooseConfiguration,
```
JordiPolo, line 12: As the only thing used from the configuration struct is `throttle_requests` and the callee more clearly is ok to `unwrap()` its value , I'd say this parameter should be instead be an usize and be called as `throttle_main(configuration.throttle_requests.unwrap(), ....` in lib.rs

src/throttle.rs
jeremyandrews: thanks for catching that! panic fixed, we do bounds checking on both sides now

src/throttle.rs
jeremyandrews: Nice little cleanup. The parent has already unwrapped the value, and there's no need to clone the entire configuration struct -- fixed.

examples/drupal_loadtest.rs
LionsAd: Some more justification:

Two reasons for early return:

a) More than two levels on indentation are by today’s standards usually considered a code smell

b) It’s good to have the same structure as used by ? operator later anyway. (It early returns on error).

Else you indent all, then outdent all for no reason and that makes the git blame way harder to read.

src/lib.rs
jeremyandrews: Great suggestion, I've introduced `setup_throttle()` which simplified the logic as well.  (And in #100 I did a similar cleanup for `setup_logger()`).

examples/drupal_loadtest.rs
jeremyandrews: Further reworked to exit early, setting us up for using `?` in the load test function itself when supported.

examples/drupal_loadtest.rs
jeremyandrews: Thanks for patiently explaining. I like it, implemented.

src/goose.rs
```diff
@@ -854,9 +863,9 @@ impl GooseUser {
     ///       let _goose = user.get("/path/to/foo/").await;
     ///     }
     /// ```
-    pub async fn get(&self, path: &str) -> GooseResponse {
+    pub async fn get(&self, path: &str) -> Result<GooseResponse, mpsc::error::SendError<bool>> {
```
LionsAd, line 866: Are you sure that's the only error that can happen?

Can be follow-up though as it's internal.

src/goose.rs
```diff
@@ -908,9 +921,13 @@ impl GooseUser {
...
     /// ```
-    pub async fn post(&self, path: &str, body: &str) -> GooseResponse {
+    pub async fn post(
+        &self,
+        path: &str,
+        body: &str,
+    ) -> Result<GooseResponse, mpsc::error::SendError<bool>> {
         let request_builder = self.goose_post(path).await.body(body.to_string());
-        self.goose_send(request_builder, None).await
+        Ok(self.goose_send(request_builder, None).await?)
```
LionsAd, line 930: Interesting pattern, but I guess it's completely identical to:

```
let result = self.goose_send(request_builder, None).await?;
Ok(result)
```

which is:

```
let result = match self.goose_send(request_builder, None).await {
  Err(e) => return Err(e),
  Ok(g) => g,
}

Ok(result)
```

So yeah - makes sense now :) - was just new for me.

src/goose.rs
```diff
@@ -1434,44 +1477,45 @@ impl GooseUser {
...
-    ///                         },
-    ///                         Err(e) => {
-    ///                             // No body was returned, log everything else.
-    ///                             user.log_debug(
-    ///                                 "error loading /",
-    ///                                 Some(goose.request),
-    ///                                 Some(headers),
-    ///                                 None,
-    ///                             );
+    ///         if let Ok(mut goose) = user.get_named("/", "index").await {
```
LionsAd, line 1480: Maybe - adjust those patterns as well to the new pattern and not the HUGE indentation change ;)

src/goose.rs
```diff
@@ -1344,15 +1388,13 @@ impl GooseUser {
     ///
     ///     /// A simple task that makes a GET request.
     ///     async fn get_function(user: &GooseUser) {
-    ///         let mut goose = user.get("/404").await;
-    ///         match &goose.response {
-    ///             Ok(r) => {
+    ///         if let Ok(mut goose) = user.get("/404").await {
```
LionsAd, line 1391: Wrong pattern

src/goose.rs
```diff
@@ -1379,28 +1421,29 @@ impl GooseUser {
...
-    ///                 // We only need to check pages that returned a success status code.
-    ///                 if r.status().is_success() {
-    ///                     match r.text().await {
-    ///                         Ok(text) => {
-    ///                             // If the expected string doesn't exist, this page load
-    ///                             // was a failure.
-    ///                             if !text.contains("this string must exist") {
-    ///                                 // As this is a named request, pass in the name not the URL
-    ///                                 user.set_failure(&mut goose.request);
+    ///         if let Ok(mut goose) = user.get_named("/", "index").await {
```
LionsAd, line 1424: Wrong pattern

src/goose.rs
```diff
@@ -2426,30 +2470,32 @@ mod tests {
 
         // Make a GET request to the mock http server and confirm we get a 200 response.
         assert_eq!(mock_index.times_called(), 0);
-        let goose = user.get("/").await;
-        let status = goose.response.unwrap().status();
-        assert_eq!(status, 200);
+        if let Ok(goose) = user.get("/").await {
```
LionsAd, line 2473: I would not do that as a panic here is correct if that fails.

src/goose.rs
```diff
@@ -2461,17 +2507,18 @@ mod tests {
 
         // Make a POST request to the mock http server and confirm we get a 200 OK response.
         assert_eq!(mock_comment.times_called(), 0);
-        let goose = user.post(COMMENT_PATH, "foo").await;
-        let unwrapped_response = goose.response.unwrap();
-        let status = unwrapped_response.status();
-        assert_eq!(status, 200);
-        let body = unwrapped_response.text().await.unwrap();
-        assert_eq!(body, "foo");
+        if let Ok(goose) = user.post(COMMENT_PATH, "foo").await {
```
LionsAd, line 2510: Just double unwrap() instead IMHO

src/goose.rs
LionsAd: Because if that fails, this part of the test is not run and that's definitely wrong ;)

src/lib.rs
```diff
@@ -1069,6 +1089,50 @@ impl GooseAttack {
...
+            throttle_receiver,
+            throttle_rx,
+        )));
+
+        let mut sender = all_threads_throttle.clone();
+        // Fill all but one slot in the channel to avoid a burst of traffic during
+        // startup.
+        for _ in 1..throttle_requests {
+            let _ = sender.send(true).await;
+        }
```
LionsAd, line 1131: That's really clever

LionsAd: The "all but one channel" is the 1..throttle_requests where else it would be 0..?

src/lib.rs
```diff
@@ -1069,6 +1089,50 @@ impl GooseAttack {
...
+        // Launch a new thread for throttling, no need to rejoin it.
+        let _ = Some(tokio::spawn(throttle::throttle_main(
+            throttle_requests,
+            throttle_receiver,
+            throttle_rx,
+        )));
+
+        let mut sender = all_threads_throttle.clone();
+        // Fill all but one slot in the channel to avoid a burst of traffic during
+        // startup.
```
LionsAd, line 1128: It might be good to add somewhere as a comment in the helper that this implements leaky-bucket :)

src/lib.rs
```diff
@@ -1140,13 +1207,19 @@ impl GooseAttack {
             ) = mpsc::unbounded_channel();
             user_channels.push(parent_sender);
 
+            // Copy the GooseUser-to-logger sender channel, used by all threads.
             if !self.configuration.debug_log_file.is_empty() {
-                // Copy the GooseUser-to-logger sender channel, used by all threads.
```
LionsAd, line 1144: nit nit nit - unrelated change :p

tests/logs.rs
```diff
@@ -16,18 +16,19 @@ pub async fn get_index(user: &GooseUser) {
...
-        match r.text().await {
-            Ok(_) => {}
-            Err(_) => {
-                user.log_debug(
-                    "there was an error",
-                    Some(goose.request),
-                    Some(headers),
-                    None,
-                );
+    if let Ok(goose) = user.get(ERROR_PATH).await {
```
LionsAd, line 19: wrong pattern

tests/redirect.rs
```diff
@@ -23,20 +23,21 @@ pub async fn get_about(user: &GooseUser) {
...
 // Task function, load REDRECT_PATH and follow redirects to ABOUT_PATH.
 pub async fn get_redirect(user: &GooseUser) {
-    let mut goose = user.get(REDIRECT_PATH).await;
-    if let Ok(r) = goose.response {
-        match r.text().await {
-            Ok(html) => {
-                // Confirm that we followed redirects and loaded the about page.
-                if !html.contains("about page") {
-                    eprintln!("about page body wrong");
+    if let Ok(mut goose) = user.get(REDIRECT_PATH).await {
```
LionsAd, line 26: wrong pattern

src/goose.rs
jeremyandrews: At this time it is the only error that can happen, so there was no need to make it generic or add a custom wrapper. We can tackle this when it's necessary/useful.

src/goose.rs
jeremyandrews: Is it overly compact? I'm not a fan of the shortest way to write something if it makes it harder to read, and would not have any problem splitting it out so we return Ok(result) instead.

src/goose.rs
jeremyandrews: It seems my grepping skills betrayed me horribly, thanks

src/goose.rs
jeremyandrews: Yeah, silly of me. It actually would still catch the failure as below this {} block it verifies it actually invoked the mock endpoint, but agreed it's better to just panic here.

src/goose.rs
jeremyandrews: We can't double unwrap the response here as we need both the `request` and the `response`, but fixing this to use `.expect()` instead so we helpfully panic if there's an unexpected error during the test.

src/lib.rs
jeremyandrews: Rewording the comment to make this explicitly clear.

src/lib.rs
jeremyandrews: Unless I'm misunderstanding, it's not _really_ a leaky bucket. A leaky bucket would be if we're not blocking when the bucket fills up, instead leaking the overflow. Hence why I've been referring to it as a "modified leaky-bucket" as we the only reason this works is we block when the bucket fills, and _don't_ leak.

jeremyandrews: Reading a little further, I see this is an implementation of the leaky bucket as a queue -- updating comment.

jeremyandrews:
```rust
        // We start from 1 instead of 0 to intentionally fill all but one slot in the
        // channel to avoid a burst of traffic during startup. The channel then provides
        // an implementation of the leaky bucket algorithm as a queue. Requests have to
        // add a token to the bucket before making a request, and are blocked until this
        // throttle thread "leaks out" a token thereby creating space. More information
        // can be found at: https://en.wikipedia.org/wiki/Leaky_bucket
```

src/lib.rs
jeremyandrews: Rebase artifact!

src/goose.rs
jeremyandrews: fixed

src/goose.rs
jeremyandrews: fixed

src/goose.rs
jeremyandrews: fixed

src/goose.rs
jeremyandrews: fixed

src/goose.rs
jeremyandrews: fixed

src/lib.rs
jeremyandrews: fixed

src/lib.rs
jeremyandrews: fixed

src/lib.rs
jeremyandrews: fixed

tests/logs.rs
jeremyandrews: fixed

tests/redirect.rs
jeremyandrews: fixed

src/lib.rs
LionsAd: 👍

src/goose.rs
LionsAd: For me:

```
let result = self.goose_send(request_builder, None).await?;
Ok(result)
```

is easier to read, but that's a question for a rust person really ...

jeremyandrews: I will leave as-is for now, we can address in a followup if we decide to split it.

src/goose.rs
```diff
@@ -2426,30 +2488,33 @@ mod tests {
 
         // Make a GET request to the mock http server and confirm we get a 200 response.
         assert_eq!(mock_index.times_called(), 0);
-        let goose = user.get("/").await;
+        let goose = user.get("/").await.expect("get returned unexpected error");
```
LionsAd, line 2491: Very nice 👍 - that's perfect

src/goose.rs
```diff
@@ -2426,30 +2488,33 @@ mod tests {
...
         let status = goose.response.unwrap().status();
         assert_eq!(status, 200);
-        assert_eq!(mock_index.times_called(), 1);
         assert_eq!(goose.request.method, GooseMethod::GET);
         assert_eq!(goose.request.name, "/");
         assert_eq!(goose.request.success, true);
         assert_eq!(goose.request.update, false);
         assert_eq!(goose.request.status_code, 200);
+        assert_eq!(mock_index.times_called(), 1);
```
LionsAd, line 2499: This is an intentional change?

jeremyandrews: The original reason for this change was so it was outside of the {} which I temporarily added through this PR. I left the change as it seemed tidy for it to consistently come last in all blocks.

src/goose.rs
```diff
@@ -1379,13 +1431,18 @@ impl GooseUser {
...
+    ///         };
+    ///
     ///         match goose.response {
-    ///             Ok(r) => {
+    ///             Ok(response) => {
     ///                 // We only need to check pages that returned a success status code.
-    ///                 if r.status().is_success() {
-    ///                     match r.text().await {
+    ///                 if response.status().is_success() {
+    ///                     match response.text().await {
```
LionsAd, line 1445: There is enough other changes that warrants this rename to be made here, but strictly speaking this is unrelated, too. ;)

PR #108, opened 2020-07-18
return a Result from Task functions

jeremyandrews:
Tasks must be converted to return `GooseTaskResult`.

This allows the use of `?` as it supports returning a GooseTaskError:

```rust
pub type GooseTaskResult = Result<(), GooseTaskError>;
```
Currently we define 3 errors:

```rust
/// Definition of all errors Goose Tasks can return.
#[derive(Debug)]
pub enum GooseTaskError {
    /// Contains any possible Reqwest library error.
    Reqwest(reqwest::Error),
    /// The request failed.
    RequestFailed,
    /// The request was canceled (this happens when the throttle is enabled and
    /// the load test finished).
    RequestCanceled,
}
```

### DIFF
--- a/examples/drupal_loadtest.rs
+++ b/examples/drupal_loadtest.rs
@@ -156,7 +156,7 @@ async fn drupal_loadtest_login(user: &GooseUser) -> GooseTaskResult {
                                 "login: no form_build_id on page: /user page",
                                 &mut goose.request,
                                 Some(&headers),
-                                Some(html.clone()),
+                                Some(&html),
                             );
                         }
                     };
@@ -221,7 +221,7 @@ async fn drupal_loadtest_post_comment(user: &GooseUser) -> GooseTaskResult {
                                 &format!("post_comment: no form_build_id found on {}", &node_path),
                                 &mut goose.request,
                                 Some(&headers),
-                                Some(html.clone()),
+                                Some(&html),
                             );
                         }
                     };
@@ -234,7 +234,7 @@ async fn drupal_loadtest_post_comment(user: &GooseUser) -> GooseTaskResult {
                                 &format!("post_comment: no form_token found on {}", &node_path),
                                 &mut goose.request,
                                 Some(&headers),
-                                Some(html.clone()),
+                                Some(&html),
                             );
                         }
                     };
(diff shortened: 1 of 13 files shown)


tests/redirect.rs
```diff
@@ -12,44 +12,51 @@ const REDIRECT3_PATH: &str = "/redirect3";
...
             }
         }
     }
+    Ok(())
 }
 
 // Task function, load REDRECT_PATH and follow redirect to new domain.
-pub async fn get_domain_redirect(user: &GooseUser) {
+pub async fn get_domain_redirect(user: &GooseUser) -> GooseTaskResult {
     let _goose = user.get(REDIRECT_PATH).await;
```
LionsAd, line 58: Missing ? At the end

jeremyandrews: Good catch, thanks!

examples/drupal_loadtest.rs
```diff
@@ -157,17 +152,12 @@ async fn drupal_loadtest_login(user: &GooseUser) {
                     let form_build_id = match re.captures(&html) {
                         Some(f) => f,
                         None => {
-                            user.set_failure(&mut goose.request);
-                            let error = "login: no form_build_id on page: /user page";
-                            // We choose to both log and display errors to stdout.
```
LionsAd, line 162: It might be a good idea to say that user.set_failure will do that automatically instead of removing?

src/goose.rs
```diff
@@ -1424,63 +1490,83 @@ impl GooseUser {
...
+        body: Option<&str>,
+    ) -> GooseTaskResult {
         // Only send update if this was previously a success.
         if request.success {
             request.success = false;
             request.update = true;
             self.send_to_parent(&request);
         }
+        // Log failure, converting `&mut request` to `&request` as needed by `log_debug()`.
+        self.log_debug(&tag, Some(&*request), headers, body);
```
LionsAd, line 1550: Maybe make it possible to log to stdout as well here - as the code did before?

examples/drupal_loadtest.rs
jeremyandrews: yes, good idea -- fixed

src/goose.rs
jeremyandrews: I meant to do that but forgot, thanks for catching it! we now use `info!()` so the debug output shows up when `-v` is enabled.

jeremyandrews:
> One thing:
> 
> I’d leave one example where we explictly call the debug_log — we optimized it away from all examples unless I miss something.

There was a commented out debug `println!()` in `drupal_loadtest`, which I converted to a commented out call to `log_debug`, along with a comment explaining why it's useful (if you're editing the load test).

PR #110, opened 2020-07-20
remove `exit(1)`, instead return error on failure

### LINKED ISSUE #109: refactor GooseAttack to return a Result
Currently GooseAttack doesn't return anything. This issue is to refactor the code to return useful errors, and to make collected statistics available for better integration into other applications (and tests).

jeremyandrews:
- Return a `Result` from all `GooseAttack` methods that can fail.
- Remove all unnecessary `exit(1)` calls from code, to simplify integrating Goose with other applications.
  - This included adding several new `GooseTaskError` types
- Update documentation briefly explaining the use of `Result<>`, `?` and `Ok(())`.

Note: This PR required a considerable number of document tests be rewritten for tests to pass. In order to simplify this process and avoid accidental errors I standardized coding style in all affected comments. Comments for functions that were not affected by this PR will be standardized in a follow up PR.

Fixes #109

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
  - properly support setting host when registering task set 
  - rename `response` wrapper to `goose`, so we end up with `goose.request` and `goose.response`
  - add `--throttle-requests` to optionally limit the maximum requests per second (api change)
+ - introduce `GooseError` and `GooseTaskError`
  - change task function signature, tasks must return a `GooseTaskResult`
- - introduce `GooseTaskError`
- - change `GooseAttack.execute()` to return `Result<GooseAttack, GooseError>`
- - introduce `GooseAttack.display_stats()` which consumes the execute Result and displays statistics
+ - change `GooseAttack` method signatures where an error is possible
+ - where possible, passs error up the stack instead of calling `exit(1)`
+ - introduce `GooseAttack.display()` which consumes the load test state and displays statistics
 
 ## 0.8.2 July 2, 2020
  - `client.log_debug()` will write debug logs to file when specified with `--debug-log-file=`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
  - add `--throttle-requests` to optionally limit the maximum requests per second (api change)
  - change task function signature, tasks must return a `GooseTaskResult`
  - introduce `GooseTaskError`
+ - change `GooseAttack.execute()` to return `Result<GooseAttack, GooseError>`
+ - introduce `GooseAttack.display_stats()` which consumes the execute Result and displays statistics
 
 ## 0.8.2 July 2, 2020
  - `client.log_debug()` will write debug logs to file when specified with `--debug-log-file=`
--- a/README.md
+++ b/README.md
@@ -102,8 +102,8 @@ will return a `GooseError` on failure. This is helpful as several of `GooseAttac
 methods can fail, returning an error. In our example, `initialize()` and `execute()`
 each may fail. The `?` that follows the method's name tells our program to exit and
 return an error on failure, otherwise continue on. The `display()` method consumes
-the everything returned by `GooseAttack` and prints a summary if statistics are
-enabled. The final line, `Ok(())` returns the empty result expected on success.
+everything returned by `GooseAttack` and prints a summary if statistics are enabled.
+The final line, `Ok(())` returns the empty result expected on success.
 
 And that's it, you've created your first load test! Let's run it and see what
 happens.
--- a/README.md
+++ b/README.md
@@ -80,19 +80,30 @@ async fn loadtest_index(user: &GooseUser) -> GooseTaskResult {
 }
 ```
 
-Finally, edit the `main()` function, setting a return type so we can use `?` to unwrap
-results, and replacing the hello world text as follows:
+Finally, edit the `main()` function, setting a return type and replacing the hello
+world text as follows:
 
 ```rust
 fn main() -> Result<(), GooseError> {
-    GooseAttack::initialize()
+    GooseAttack::initialize()?
         .register_taskset(taskset!("LoadtestTasks")
             .register_task(task!(loadtest_index))
         )
         .execute()?
-        .display_stats();
+        .display();
+    
+    Ok(())
 }
 ```
+
+If you're new to Rust, `main()`'s return type of `Result<(), GooseError>` may look
+strange. It essentially says that `main` will return nothing (`()`) on success, and
+will return a `GooseError` on failure. This is helpful as several of `GooseAttack`'s
+methods can fail, returning an error. In our example, `initialize()` and `execute()`
+each may fail. The `?` that follows the method's name tells our program to exit and
+return an error on failure, otherwise continue on. The `display()` method consumes
+the everything returned by `GooseAttack` and prints a summary if statistics are
+enabled. The final line, `Ok(())` returns the empty result expected on success.
 
 And that's it, you've created your first load test! Let's run it and see what
 happens.
--- a/README.md
+++ b/README.md
@@ -89,7 +89,8 @@ fn main() {
         .register_taskset(taskset!("LoadtestTasks")
             .register_task(task!(loadtest_index))
         )
-        .execute();
+        .execute()?
+        .display_stats();
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -80,11 +80,11 @@ async fn loadtest_index(user: &GooseUser) -> GooseTaskResult {
 }
 ```
 
-Finally, edit the `main()` function, removing the hello world text and replacing
-it as follows:
+Finally, edit the `main()` function, setting a return type so we can use `?` to unwrap
+results, and replacing the hello world text as follows:
 
 ```rust
-fn main() {
+fn main() -> Result<(), GooseError> {
     GooseAttack::initialize()
         .register_taskset(taskset!("LoadtestTasks")
             .register_task(task!(loadtest_index))
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -19,7 +19,7 @@
 
 use goose::prelude::*;
 
-fn main() {
+fn main() -> Result<(), GooseError> {
     GooseAttack::initialize()
         // In this example, we only create a single taskset, named "WebsiteUser".
         .register_taskset(
@@ -32,7 +32,10 @@ fn main() {
                 .register_task(task!(website_index))
                 .register_task(task!(website_about)),
         )
-        .execute();
+        .execute()?
+        .display_stats();
+
+    Ok(())
 }
 
 /// Demonstrates how to log in when a user starts. We flag this task as an
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -20,7 +20,7 @@
 use goose::prelude::*;
 
 fn main() -> Result<(), GooseError> {
-    GooseAttack::initialize()
+    GooseAttack::initialize()?
         // In this example, we only create a single taskset, named "WebsiteUser".
         .register_taskset(
             taskset!("WebsiteUser")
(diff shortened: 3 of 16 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
  - properly support setting host when registering task set 
  - rename `response` wrapper to `goose`, so we end up with `goose.request` and `goose.response`
  - add `--throttle-requests` to optionally limit the maximum requests per second (api change)
+ - change task function signature, tasks must return a GooseTaskResult
+ - introduce `GooseTaskError`
```
LionsAd, line 9: Ah - this belongs to the previous PR?

I would probably make a quick commit / PR outside of this one as it's confusing ;)

examples/simple.rs
```diff
@@ -32,7 +32,10 @@ fn main() {
                 .register_task(task!(website_index))
                 .register_task(task!(website_about)),
         )
-        .execute();
+        .execute()?
+        .display_stats();
```
LionsAd, line 36: Is there also a way to return the stats and not just display them?

jeremyandrews: Currently you can either return the stats, or display the stats. It would be trivial to add a wrapper to make a third option of returning and displaying the stats both, I just don't have a use case. 

For now I chose to simply consume the stats (and errors) when displaying them, as otherwise you end up needing to do something like this:
```
let _ = GooseAttack::...
```
I may rename `display_stats()` to just `display()` as this may also be where we display errors in the normal CLI use-case. I will explore what's required for that shortly.

CHANGELOG.md
jeremyandrews: Ok, there's also a tiny README change that was also missed in the previous PR, I'll merge both outside of this new PR

examples/simple.rs
LionsAd: Just .display() feels nice — it would be cool if that optionally could take the stats so that you can do:

let goose_atk = ...;
let stats = goose_atk.get_stats();
goose_atk.display(Some(stats));

Or so ...

jeremyandrews: Errors are all handled through the logger already, so it turns out the only thing we have to display are stats. But sure, `.display()` still works.

I'm not sure what you're trying to accomplish with your example? Currently you could do that, but you'd have to pass in `goose_atk` in your example:

```rust
let goose_atk = ...
// do other stuff
display_stats(goose_atk);
```

LionsAd: What I mean is:

I thought it might be possible to return the stats, modify them (custom cleanup) and then display them.

LionsAd: I dislike APIs that encapsulate data needlessly - I like that it's easy to display stats though.

jeremyandrews: The issue here is that this is all internal state, and the stats are not likely to be of use without some of this internal state (ie, information like when the test started, how long it ran...)

I suggest we create an issue and tackle exposing the stats in a more useful way through an API (and that may be the same as your example: `let stats = get_stats(internal_goose_state)`). That said, if we remove everything but the stats themselves, we'd need a different `.display()` function that prints stats with less information at its disposal.

README.md
```diff
@@ -80,19 +80,31 @@ async fn loadtest_index(user: &GooseUser) -> GooseTaskResult {
...
 }
 ```
 
+If you're new to Rust, `main()`'s return type of `Result<(), GooseError>` may look
+strange. It essentially says that `main` will return nothing (`()`) on success, and
+will return a `GooseError` on failure. This is helpful as several of `GooseAttack`'s
+methods can fail, returning an error. In our example, `initialize()` and `execute()`
+each may fail. The `?` that follows the method's name tells our program to exit and
+return an error on failure, otherwise continue on. The `display()` method consumes
+the everything returned by `GooseAttack` and prints a summary if statistics are
```
LionsAd, line 105: nit - 'the' is too much here

src/goose.rs
```diff
@@ -427,19 +478,23 @@ impl GooseTaskSet {
...
+    ///     Ok(())
+    /// }
     /// ```
-    pub fn set_weight(mut self, weight: usize) -> Self {
+    pub fn set_weight(mut self, weight: usize) -> Result<Self, GooseError> {
         trace!("{} set_weight: {}", self.name, weight);
         if weight < 1 {
             error!("{} weight of {} not allowed", self.name, weight);
-            std::process::exit(1);
+            return Err(GooseError::InvalidWeight);
```
LionsAd, line 493: First I was like: Why would set_weight fail, but now it makes sense :)

src/goose.rs
```diff
@@ -1695,29 +1753,26 @@ impl GooseUser {
...
-    pub async fn set_client_builder(&self, builder: ClientBuilder) {
-        match builder.build() {
-            Ok(c) => *self.client.lock().await = c,
-            Err(e) => {
-                error!("failed to build web client: {}", e);
-                std::process::exit(1);
-            }
-        };
+    pub async fn set_client_builder(&self, builder: ClientBuilder) -> Result<(), GooseTaskError> {
+        *self.client.lock().await = builder.build()?;
```
LionsAd, line 1773: Out of interest: Is *self == self.unwrap()?

src/goose.rs
```diff
@@ -1748,35 +1803,39 @@ impl GooseUser {
...
-    ///         .set_wait_time(0, 3)
-    ///         .register_task(task!(task_foo).set_weight(10))
-    ///         .register_task(task!(task_bar))
-    ///     )
-    ///     .execute();
+    /// fn main() -> Result<(), GooseError> {
+    ///     GooseAttack::initialize()?
+    ///         .register_taskset(taskset!("LoadtestTasks").set_host("http//foo.example.com/")
+    ///             .set_wait_time(0, 3)?
+    ///             .register_task(task!(task_foo).set_weight(10)?)
```
LionsAd, line 1810: The question mark operator really looks nice here in all this encapsualation :)

src/lib.rs
```diff
@@ -357,6 +365,53 @@ pub fn get_worker_id() -> usize {
...
+        match *self {
+            GooseError::Io(ref err) => err.fmt(f),
+            GooseError::Reqwest(ref err) => err.fmt(f),
+            GooseError::FeatureNotEnabled => write!(f, "Compile time feature not enabled."),
+            GooseError::InvalidHost => write!(f, "Unable to parse host."),
+            GooseError::InvalidOption => write!(f, "Invalid option specified."),
+            GooseError::InvalidWaitTime => write!(f, "Invalid wait time specified."),
+            GooseError::InvalidWeight => write!(f, "Invalid weight specified."),
+            GooseError::NoTaskSets => write!(f, "No task sets defined."),
+        }
```
LionsAd, line 395: This looks really nice now :)

src/goose.rs
jeremyandrews: We pass in a reference (&self). The `*` dereferences this memory pointer it so we can grab the actual lock and set the builder contained.

src/lib.rs
```diff
@@ -571,14 +630,14 @@ impl GooseAttack {
...
-                        std::process::exit(1);
+                        return Err(GooseError::InvalidOption);
                     } else {
                         0
                     }
                 } else {
                     if self.configuration.worker {
                         error!("The --users option is only available to the manager.");
-                        std::process::exit(1);
+                        return Err(GooseError::InvalidOption);
```
LionsAd, line 640: Would it make sense to include the string in the GooseError::InvalidOption in case someone does not want to display it via error!()? [can be follow-up as it's unrelated]

jeremyandrews: I have not explored how to put more data into the Error, and decided to explore this as a followup as the PR is already so large. But I agree, it would be nice to include more information, and to not force an `error!()` on them.

src/lib.rs
```diff
@@ -1285,16 +1402,12 @@ impl GooseAttack {
...
-                        self.configuration.stats_log_file, e
-                    );
-                    std::process::exit(1);
-                }
-            }
+            info!(
+                "opening file to log statistics: {}",
+                self.configuration.stats_log_file
+            );
+            let file = File::create(&self.configuration.stats_log_file).await?;
```
LionsAd, line 1409: :D - looks great!

src/lib.rs
```diff
@@ -1824,27 +1945,40 @@ fn weight_tasks(
...
 }
+
 #[cfg(test)]
 mod test {
     use super::*;
 
     #[test]
     fn valid_host() {
-        // We can only test valid domains, as we exit on failure.
-        // @TODO: rework so we don't exit on failure
```
LionsAd, line 1843: Nice! :)

src/worker.rs
```diff
@@ -133,14 +133,20 @@ pub async fn worker_main(goose_attack: &GooseAttack) -> GooseAttack {
...
                 initializer.min_wait,
                 initializer.max_wait,
                 &initializer.config,
                 goose_attack.task_sets_hash,
-            );
+            ) {
+                Ok(u) => u,
+                Err(e) => {
+                    error!("[{}] failed to create GooseUser: {}", get_worker_id(), e);
+                    std::process::exit(1);
```
LionsAd, line 147: Could use .expect() so it "only" panics?

src/worker.rs
```diff
@@ -232,9 +238,16 @@ pub async fn worker_main(goose_attack: &GooseAttack) -> GooseAttack {
...
-    worker_goose_attack
-        .launch_users(started, sleep_duration, Some(manager))
+    match worker_goose_attack
+        .launch_users(sleep_duration, Some(manager))
         .await
+    {
+        Ok(w) => w,
+        Err(e) => {
+            error!("[{}] failed to launch GooseAttack: {}", get_worker_id(), e);
+            std::process::exit(1);
```
LionsAd, line 248: .expect instead?

tests/one_taskset.rs
```diff
@@ -27,14 +27,16 @@ fn test_single_taskset() {
         .return_body("<HTML><BODY>about page</BODY></HTML>")
         .create();
 
-    crate::GooseAttack::initialize_with_config(common::build_configuration())
+    let _stats = crate::GooseAttack::initialize_with_config(common::build_configuration())
```
LionsAd, line 30: Technically it's a goose attack and not stats ...

tests/no_normal_tasks.rs
```diff
@@ -26,14 +26,16 @@ fn test_no_normal_tasks() {
     let mock_login = mock(POST, LOGIN_PATH).return_status(200).create();
     let mock_logout = mock(GET, LOGOUT_PATH).return_status(200).create();
 
-    crate::GooseAttack::initialize_with_config(common::build_configuration())
+    let _stats = crate::GooseAttack::initialize_with_config(common::build_configuration())
```
LionsAd, line 29: Technically it's a goose attack and not stats ...

tests/gaggle.rs
```diff
@@ -36,11 +36,13 @@ fn test_gaggle() {
         master_configuration.manager = true;
         master_configuration.expect_workers = 1;
         master_configuration.run_time = "3".to_string();
-        crate::GooseAttack::initialize_with_config(master_configuration)
+        let _stats = crate::GooseAttack::initialize_with_config(master_configuration)
```
LionsAd, line 39: Technically it's a goose attack and not stats ...

src/worker.rs
jeremyandrews: Yes, agreed a Panic would be better. Will do in a followup.

README.md
jeremyandrews: whoops, rewrote too many times, thanks

tests/one_taskset.rs
jeremyandrews: fixed

tests/gaggle.rs
jeremyandrews: fixed

tests/no_normal_tasks.rs
jeremyandrews: fixed

PR #113, opened 2020-07-22
add more detail to errors

jeremyandrews:
The goal of this PR is to make errors more rich, allowing application developers the option of what they want to display. In our examples, `main()` returns the errors so they are simply displayed.

 - enriched `GooseError::InvalidHost`, for example:
   ```
   $ cargo run --example drupal_loadtest -- --host=foo
       Finished dev [unoptimized + debuginfo] target(s) in 0.08s
        Running `target/debug/examples/drupal_loadtest --host=foo`
   Error: InvalidHost { host: "foo", detail: Some("Failure parsing host specified with --host"), parse_error: RelativeUrlWithoutBase }
   ```
 - enriched `GooseError::FeatureNotEnabled`, for example:
   ```
   $ cargo run --example drupal_loadtest -- --worker
       Finished dev [unoptimized + debuginfo] target(s) in 0.08s
        Running `target/debug/examples/drupal_loadtest --worker`
   Error: FeatureNotEnabled { feature: "gaggle", detail: Some("goose must be recompiled with `--features gaggle` to start in worker mode") }
   ```

 - enriched `GooseError::InvalidWaitTime`, for example:
   ```
        Running `target/debug/examples/drupal_loadtest`
   Error: InvalidWaitTime { min_wait: 5, max_wait: 1, detail: Some("min_wait can not be larger than max_wait") }

   ```

 - enriched `GooseError::InvalidWeight`, for example:
```
        Running `target/debug/examples/drupal_loadtest`
   Error: InvalidWeight { weight: 0, detail: Some("weight of 0 not allowed") }
```

 - enriched `GooseError::NoTaskSets`
   ```
        Running `target/debug/examples/drupal_loadtest`
   Error: NoTaskSets { detail: Some("no task sets defined") }
   ```

- enriched `GooseError::InvalidOption`
   ```
   $ cargo run --example drupal_loadtest -- --host https://127.0.0.1/ --expect-workers=10
       Finished dev [unoptimized + debuginfo] target(s) in 0.08s
        Running `target/debug/examples/drupal_loadtest --host 'https://127.0.0.1/' --expect-workers=10`
   Error: InvalidOption { option: "--expect-workers", value: "10", detail: Some("--expect-workers is only available when running in manager mode") }
   ```

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -537,11 +537,11 @@ impl GooseTaskSet {
             max_wait
         );
         if min_wait > max_wait {
-            error!(
-                "min_wait({}) can't be larger than max_wait({})",
-                min_wait, max_wait
-            );
-            return Err(GooseError::InvalidWaitTime);
+            return Err(GooseError::InvalidWaitTime {
+                min_wait,
+                max_wait,
+                detail: Some("min_wait can not be larger than max_wait".to_string()),
+            });
         }
         self.min_wait = min_wait;
         self.max_wait = max_wait;
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -488,9 +488,11 @@ impl GooseTaskSet {
     /// ```
     pub fn set_weight(mut self, weight: usize) -> Result<Self, GooseError> {
         trace!("{} set_weight: {}", self.name, weight);
-        if weight < 1 {
-            error!("{} weight of {} not allowed", self.name, weight);
-            return Err(GooseError::InvalidWeight);
+        if weight == 0 {
+            return Err(GooseError::InvalidWeight {
+                weight,
+                detail: Some("weight of 0 not allowed".to_string()),
+            });
         } else {
             self.weight = weight;
         }
@@ -2031,9 +2033,11 @@ impl GooseTask {
             self.tasks_index,
             weight
         );
-        if weight < 1 {
-            error!("{} weight of {} not allowed", self.name, weight);
-            return Err(GooseError::InvalidWeight);
+        if weight == 0 {
+            return Err(GooseError::InvalidWeight {
+                weight,
+                detail: Some("weight of 0 not allowed".to_string()),
+            });
         } else {
             self.weight = weight;
         }
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -1850,16 +1850,37 @@ pub fn get_base_url(
     config_host: Option<String>,
     task_set_host: Option<String>,
     default_host: Option<String>,
-) -> Url {
+) -> Result<Url, GooseError> {
     // If the `--host` CLI option is set, build the URL with it.
     match config_host {
-        Some(host) => Url::parse(&host).unwrap(),
+        Some(host) => Ok(
+            Url::parse(&host).map_err(|parse_error| GooseError::InvalidHost {
+                host,
+                detail: Some("Failure parsing host specified with --host".to_string()),
+                parse_error,
+            })?,
+        ),
         None => {
             match task_set_host {
                 // Otherwise, if `GooseTaskSet.host` is defined, usee this
-                Some(host) => Url::parse(&host).unwrap(),
+                Some(host) => {
+                    Ok(
+                        Url::parse(&host).map_err(|parse_error| GooseError::InvalidHost {
+                            host,
+                            detail: Some(
+                                "Failure parsing host specified with GooseTaskSet.set_host()"
+                                    .to_string(),
+                            ),
+                            parse_error,
+                        })?,
+                    )
+                }
                 // Otherwise, use global `GooseAttack.host`. `unwrap` okay as host validation was done at startup.
-                None => Url::parse(&default_host.unwrap()).unwrap(),
+                None => {
+                    // Host is required, if we get here it's safe to unwrap this variable.
+                    let default_host = default_host.unwrap();
+                    Ok(Url::parse(&default_host).map_err(|parse_error| GooseError::InvalidHost { host: default_host.to_string(), detail: Some("Failure parsing host specified globally with GooseAttack.set_host()".to_string()), parse_error })?)
+                }
             }
         }
     }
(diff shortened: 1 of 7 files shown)


src/goose.rs
```diff
@@ -1850,16 +1852,37 @@ pub fn get_base_url(
...
-                Some(host) => Url::parse(&host).unwrap(),
+                Some(host) => {
+                    Ok(
+                        Url::parse(&host).map_err(|parse_error| GooseError::InvalidHost {
+                            host,
+                            detail: Some(
+                                "Failure parsing host specified with GooseTaskSet.set_host()"
+                                    .to_string(),
+                            ),
+                            parse_error,
```
LionsAd, line 1876: This should be called:

`source: parse_error` to work with how standard Errors work.

src/goose.rs
```diff
@@ -488,9 +488,11 @@ impl GooseTaskSet {
...
         trace!("{} set_weight: {}", self.name, weight);
-        if weight < 1 {
-            error!("{} weight of {} not allowed", self.name, weight);
-            return Err(GooseError::InvalidWeight);
+        if weight == 0 {
+            return Err(GooseError::InvalidWeight {
+                weight,
+                detail: Some("weight of 0 not allowed".to_string()),
+            });
         } else {
```
LionsAd, line 496: Unrelated: else is strictly speaking not needed as you early return (but preference for sure)

src/goose.rs
```diff
@@ -488,9 +488,11 @@ impl GooseTaskSet {
     /// ```
     pub fn set_weight(mut self, weight: usize) -> Result<Self, GooseError> {
         trace!("{} set_weight: {}", self.name, weight);
-        if weight < 1 {
-            error!("{} weight of {} not allowed", self.name, weight);
-            return Err(GooseError::InvalidWeight);
+        if weight == 0 {
```
LionsAd, line 491: Oh - usize ensures it won't ever be unsigned -- good catch.

src/goose.rs
```diff
@@ -1850,16 +1852,37 @@ pub fn get_base_url(
...
+                            parse_error,
+                        })?,
+                    )
+                }
                 // Otherwise, use global `GooseAttack.host`. `unwrap` okay as host validation was done at startup.
-                None => Url::parse(&default_host.unwrap()).unwrap(),
+                None => {
+                    // Host is required, if we get here it's safe to unwrap this variable.
+                    let default_host = default_host.unwrap();
+                    Ok(Url::parse(&default_host).map_err(|parse_error| GooseError::InvalidHost { host: default_host.to_string(), detail: Some("Failure parsing host specified globally with GooseAttack.set_host()".to_string()), parse_error })?)
```
LionsAd, line 1884: nit - Maybe split this line?

src/goose.rs
jeremyandrews: Clippy has caught this for me in the past, I don't know why it didn't complain about this one.

src/goose.rs
jeremyandrews: we now define the source for errors, if any -- I kept the name `parse_error` instead of `source` though as I feel it provides more context when displaying the error through debug

src/lib.rs
```diff
@@ -368,36 +367,81 @@ pub struct Socket {}
...
+            GooseError::Io(ref source) => Some(source),
+            GooseError::Reqwest(ref source) => Some(source),
+            GooseError::FeatureNotEnabled { .. } => None,
+            GooseError::InvalidHost {
+                ref parse_error, ..
+            } => Some(parse_error),
+            GooseError::InvalidOption { .. } => None,
+            GooseError::InvalidWaitTime { .. } => None,
+            GooseError::InvalidWeight { .. } => None,
+            GooseError::NoTaskSets { .. } => None,
```
LionsAd, line 440: I think you can also match `_ => None` if you don't want to enumerate all ...

src/lib.rs
```diff
@@ -368,36 +367,81 @@ pub struct Socket {}
...
+    /// `.detail`.
+    NoTaskSets {
+        detail: Option<String>,
+    },
 }
 
+// Define how to display errors.
 impl fmt::Display for GooseError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::Display::fmt(&self, f)
```
LionsAd, line 423: I still don't get what this does. Maybe that is even the default implementation?

src/goose.rs
jeremyandrews: Fixed. I let rustfmt handle this for me, but this line was ever-so-slightly different than the lines above so it chose to not reformat it. Doing it manually I agree it reads better.

src/lib.rs
jeremyandrews: Excellent point, easier to maintain as we add more errors.

src/goose.rs
jeremyandrews: fixed

src/lib.rs
jeremyandrews: It is required so our errors can be displayed via debug, and seems to be the minimum required.

src/goose.rs
```diff
@@ -317,49 +316,66 @@ macro_rules! taskset {
...
 }
 
-/// Goose tasks return a result, which is empty on success, or contains a GooseTaskError
-/// on error.
-pub type GooseTaskResult = Result<(), GooseTaskError>;
-
+// Define how to display errors.
 impl fmt::Display for GooseTaskError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::Display::fmt(&self, f)
```
LionsAd, line 361: @jeremyandrews This will create an infinite loop as per https://github.com/rust-lang/rust/issues/74892.

LionsAd: It only works right now as we use the Debug and not the Display output for errors when returning from main.

jeremyandrews: see #128

PR #116, opened 2020-07-25
expose statistics from GooseAttack via GooseStats

### LINKED ISSUE #115: Options for running statistics and summary statistics
Currently stats are enabled or disabled "globally". Either the running statistics every 15seconds and the final summary both appear or if disabled neither appear.
In my use-case I love the final statistics but pretty much want to get rid of the running statistics but it is not possible. All the structs are also private so can't do much from there either.

jeremyandrews:
- return `GooseStats` from `GooseAttack.execute()`, a public sub-object contained within `GooseAttack`
 - move `GooseStats` to `mod.stats` and rework functions as methods
 - consume and display statistics with `goose_stats.print()`
 - other methods: `.print_running()`, `.print_requests()`, `.print_response_times()`, `.print_percentiles()` and `.print_status_codes()`
 - add test coverage to validate `GooseStats` fields are public
 - add test coverage to validate that Goose statistics match server statistics

Fixes #115

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -90,7 +90,7 @@
 //! use goose::prelude::*;
 //!
 //! fn main() -> Result<(), GooseError> {
-//!     GooseAttack::initialize()?
+//!     let _goose_stats = GooseAttack::initialize()?
 //!         .register_taskset(taskset!("LoadtestTasks")
 //!             .set_wait_time(0, 3)?
 //!             // Register the foo task, assigning it a weight of 10.
@@ -339,6 +339,7 @@ use url::Url;
 use crate::goose::{
     GooseDebug, GooseRawRequest, GooseRequest, GooseTask, GooseTaskSet, GooseUser, GooseUserCommand,
 };
+use crate::stats::GooseStats;
 
 /// Constant defining how often statistics should be displayed while load test is running.
 const RUNNING_STATS_EVERY: usize = 15;
@@ -363,9 +364,6 @@ pub fn get_worker_id() -> usize {
 #[derive(Debug)]
 /// Socket used for coordinating a Gaggle, a distributed load test.
 pub struct Socket {}
-
-/// Goose optionally tracks statistics about requests made during a load test.
-pub type GooseRequestStats = HashMap<String, GooseRequest>;
 
 /// Definition of all errors a GooseAttack can return.
 #[derive(Debug)]
@@ -450,45 +448,6 @@ impl From<io::Error> for GooseError {
     }
 }
 
-/// Statistics collected during a Goose load test.
-///
-/// # Example
-/// ```rust,no_run
-///     use goose::prelude::*;
-///
-/// fn main() -> Result<(), GooseError> {
-///     let stats = GooseAttack::initialize()?
-///         .register_taskset(taskset!("ExampleUsers")
-///             .register_task(task!(example_task))
-///         )
-///         .execute()?
-///         .get();
-///
-///     // Do something with stats ...
-///     println!("{:#?}", stats);
-///
-///     Ok(())
-/// }
-///
-/// async fn example_task(user: &GooseUser) -> GooseTaskResult {
-///     let _goose = user.get("/").await?;
-///
-///     Ok(())
-/// }
-/// ```
-#[derive(Clone, Debug)]
-pub struct GooseStats {
-    /// A hash of the load test, useful to verify if different statistics are from
-    /// the same load test.
-    pub hash: u64,
-    /// How many seconds the load test ran.
-    pub duration: usize,
-    /// Total number of users simulated during this load test.
-    pub users: usize,
-    /// Goose request statistics.
-    pub requests: GooseRequestStats,
-}
-
 /// Internal global state for load test.
 #[derive(Clone)]
 pub struct GooseAttack {
@@ -498,8 +457,6 @@ pub struct GooseAttack {
     test_stop_task: Option<GooseTask>,
     /// A vector containing one copy of each GooseTaskSet that will run during this load test.
     task_sets: Vec<GooseTaskSet>,
-    /// A checksum of the task_sets vector to be sure all workers are running the same load test.
-    task_sets_hash: u64,
     /// A weighted vector containing a GooseUser object for each user that will run during this load test.
     weighted_users: Vec<GooseUser>,
     /// An optional default host to run this load test against.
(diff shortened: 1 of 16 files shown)


src/lib.rs
```diff
@@ -454,31 +454,31 @@ impl From<io::Error> for GooseError {
...
+    pub users: usize,
     /// Track how many users are already loaded.
-    active_users: usize,
+    pub active_users: usize,
     /// All requests statistics merged together.
-    merged_requests: GooseRequestStats,
+    pub statistics: GooseRequestStats,
     /// When the load test started.
-    started: Option<time::Instant>,
+    pub started: Option<time::Instant>,
```
LionsAd, line 481: I am a fan of public properties, but in PHP world I learned that that says to users that they can rely on a stable structure of the struct.

It is like an API / ABI then.

It’s also the sole reason for getters / setters to hide the internal structure.

Just some food for thought. I am not sure how Rust handles that in general.

LionsAd:
https://users.rust-lang.org/t/public-getter-method-vs-pub-field/20147

LionsAd:
„Getters and setters make sense to me over direct access when direct access could allow users to violate some invariants of your code.“

LionsAd:
Overall maybe a substruct of GooseAttack that is a pure value object would make sense to have all properties public?

jeremyandrews:
@LionsAd what about `execute()` returning a `GooseStats` object instead of returning the entire `GooseAttack` object, which would be everything needed to display the final stats:

```rust
pub struct GooseStats {
    /// A hash of the load test, useful to compare if two tests are the same.
    pub hash: u64,
    /// The instant the load test was started.
    pub started: time::Instant,
    /// How many seconds the load test ran.
    pub run_time: usize,
    /// Total number of users simulated during this load test.
    pub users: usize,
    /// Goose request statistics.
    pub statistics: GooseRequestStats,
}
```

(The internal `GooseAttack` object would be unchanged, but each field would not be made `pub` as is currently done in this PR)

src/lib.rs
jeremyandrews: This has been reworked so you have to call `get_stats()` on the `GooseAttack` internal state, which then exposes the relevant statistics. 

At this time `GooseAttack` doesn't use `GooseStats` for anything, it merely exposes the information to anyone needing it (and is currently only used by one integration test to validate that statistics are accurate).

LionsAd:
GooseState would be preferable - also display() makes more sense to be implemented for the GooseState trait.

That was roughly where I wanted to go originally.

So glad we end up there again. Turns out also your intuition on _goose_stats was right ;)

LionsAd:
So I would have execute() return the value object — as that is the final step that is a common pattern to no longer chain then but return a value object.

And then move display to GooseStats which means someone can do what I wanted to get_stats, clean and display.

src/lib.rs
```diff
@@ -450,6 +450,45 @@ impl From<io::Error> for GooseError {
...
+///     Ok(())
+/// }
+/// ```
+#[derive(Clone, Debug)]
+pub struct GooseStats {
+    /// A hash of the load test, useful to verify if different statistics are from
+    /// the same load test.
+    pub hash: u64,
+    /// How many seconds the load test ran.
+    pub run_time: usize,
```
JordiPolo, line 485: The GooseAttack struct has `run_time` also,  confusing to have both attributes with the same name or useful?

Also, in my case at least I want to know the time the test will run before running it so the `run_time` in GooseAttack, not sure if it makes sense to make fields on that struct pub also.

jeremyandrews:
> GooseState would be preferable - also display() makes more sense to be implemented for the GooseState trait.

What is now `GooseAttack` was originally `GooseState`, That said, what do you prefer about `GooseState` as compared to `GooseStats` or `GooseStatistics`? My general thinking is, "these are the relevant statistics generated from a goose attack", whereas to me a goose state implies it's a paused test that could potentially be restarted.

jeremyandrews:
> So I would have execute() return the value object — as that is the final step that is a common pattern to no longer chain then but return a value object.

`.execute()` has to return the `GooseAttack` object otherwise displaying and/or getting can't be chained. We'd end up having to do something like this when writing a goose load test:

```rust
@@ -28,7 +28,7 @@ use rand::Rng;
 use regex::Regex;
 
 fn main() -> Result<(), GooseError> {
-    GooseAttack::initialize()?
+    let goose_attack = GooseAttack::initialize()?
         .register_taskset(
             taskset!("AnonBrowsingUser")
                 .set_weight(4)?
@@ -77,8 +77,8 @@ fn main() -> Result<(), GooseError> {
                         .set_name("(Auth) comment form"),
                 ),
         )
-        .execute()?
-        .display();
+        .execute()?;
+    GooseAttack::display(goose_attack);
 
     Ok(())
 }
```

I have instead added three functions, any of which can be chained to the `GooseAttack`: `.display()`, `display_and_get()`, and `get()`. What's not yet stalled is your preference to be able to modify the stats before displaying them, it's possible but it would require a fourth function which can't be chained.

src/lib.rs
jeremyandrews: My original intent was that these are the same name because they are the same value. But if you ctrl-c a test before it finishes, they would not be the same value. So in `GooseStats` I've now renamed it to `duration`.

You would have to set `run_time` in your code, as otherwise Goose runs forever. So it seems you already have access to this information?

jeremyandrews:
> What's not yet implemented is your preference to be able to modify the stats before displaying them, it's possible but it would require a fourth function which can't be chained.

Actually, this is already possible, though not especially intuitive. You can call `stats::print_final_stats(stats);` where `stats` is the `GooseStats` which you can first clean if desired.

jeremyandrews:
Reworking a bit so `GooseStats` lives in `mod.stats` and `display` is a method of it, and so `GooseAttack` contains a `GooseStats` object rather than converts to one.

src/lib.rs
JordiPolo: Yes, but I think then the parse_time function in utils would need to be public? I want time as usize, already parsed from command-line.

jeremyandrews: I believe you'd do it the same as the integration tests? For example: https://github.com/tag1consulting/goose/blob/master/tests/common.rs#L12 <-- this sets it to 1 second, but that could be set to any desired number of seconds. Inside `GooseStats` the `duration` is returned as a `usize` measured in seconds.

jeremyandrews:
> So I would have execute() return the value object — as that is the final step that is a common pattern to no longer chain then but return a value object.
> 
> And then move display to GooseStats which means someone can do what I wanted to get_stats, clean and display.

This is now how it is implemented. We chain it in tests and examples, but you could also take the `goose_stats` and modify/clean them first if desired.

src/lib.rs
JordiPolo: I want to be able to pass "3m" or "4h" though the command line as Goose does but work later with an usize.
The reason I want the usize is because I am [displaying a progress bar](https://github.com/JordiPolo/minos/blob/master/src/command_performance.rs#L38-L57) while the user is waiting for the final stats. Maybe I should implement this inside Goose directly, are you interested?

jeremyandrews: A progress bar in goose would be very welcome.

jeremyandrews:
Closing in favor of #118

jeremyandrews:
Whoops, this will get committed and then the next will be followup

PR #118, opened 2020-07-28
Implement fmt::display for GooseStats

### LINKED ISSUE #115: Options for running statistics and summary statistics

jeremyandrews:
This is a followup on #116.

 - implement `fmt::Display` allowing us to simply `print!("{}", goose_stats);`
 - rename methods: `.fmt_requests()`, `.fmt_response_times()`, `.fmt_percentiles()` and `.fmt_status_codes()`

### DIFF
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -1,8 +1,8 @@
 use itertools::Itertools;
 use num_format::{Locale, ToFormattedString};
 use std::collections::{BTreeMap, HashMap};
+use std::io::{Result, Write};
 use std::{f32, fmt, str};
-use std::io::{Write, Result};
 
 use crate::goose::GooseRequest;
 use crate::util;
@@ -94,8 +94,8 @@ impl GooseStats {
         println!("{}", self);
     }
 
-    /// Display a table of requests and fails.
-    pub fn print_requests(&self, buffer: &mut Vec<u8>) -> Result<&GooseStats> {
+    /// Prepares a table of requests and fails.
+    pub fn fmt_requests(&self, buffer: &mut Vec<u8>) -> Result<&GooseStats> {
         // If there's nothing to display, exit immediately.
         if self.requests.is_empty() {
             return Ok(self);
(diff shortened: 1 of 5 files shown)


src/stats.rs
```diff
@@ -47,6 +46,10 @@ pub struct GooseStats {
     pub users: usize,
     /// Goose request statistics.
     pub requests: GooseRequestStats,
+    /// Flag indicating whether or not to display percentile.
+    pub display_percentile: bool,
+    /// Flag indicating whether or not to display status_codes.
+    pub display_status_codes: bool,
```
LionsAd, line 52: Is it ensured that both are false at the start?

src/stats.rs
```diff
@@ -262,32 +281,45 @@ impl GooseStats {
...
+        )?;
+        writeln!(
+            f,
+            " Slowest page load within specified percentile of requests (in ms):"
+        )?;
+        writeln!(
+            f,
+            " ------------------------------------------------------------------------------"
+        )?;
+        writeln!(
```
LionsAd, line 314: This is preference, but I feel:

```
writeln!(f,
```

is almost as short as:

```
println!(
```

so I would not split it over several lines in the whole PR as it makes things much more difficult to read.

Also the diff is shorter ;)

src/stats.rs
jeremyandrews: Yes, because of https://doc.rust-lang.org/std/primitive.bool.html#impl-Default

I should update the comment to make this explicitly clear.

src/stats.rs
jeremyandrews: I do not format white space manually, but instead use `cargo fmt`. We would need to set up custom rustfmt rules if we want to change how it formats things like this. (In this case I believe we'd need to adjust the line-length before a wrap). 

I prefer to stick with the defaults.

src/stats.rs
jeremyandrews: Updated:

```rust
    /// Flag indicating whether or not to display percentile. Because we're deriving Default,
    /// this defaults to false.
    pub display_percentile: bool,
    /// Flag indicating whether or not to display status_codes. Because we're deriving Default,
    /// this defaults to false.
    pub display_status_codes: bool,
```

src/stats.rs
LionsAd: I agree

Issue #76, opened 2020-06-20
is it possible to pass paths programatically?


JordiPolo:
Probably this is not really an issue but really my inability with Rust.

I have[ a project ](https://github.com/JordiPolo/minos)that reads from an openapi file to generate tests. I thought I could integrate performance tests also via Goose.

First I create the main structs needed
```rust
    let mut configuration = GooseConfiguration::default();
    configuration.hatch_rate = 1; 
    configuration.host = config.base_url.clone();
    let goose_attack = GooseAttack::initialize_with_config(configuration);
    let mut task_set = taskset!("ExampleTasks");
```

This works. BY THE WAY, although there is a default of 1 for hatch_rate in the GooseConfiguration struct, somehow I needed to set it manually to 1 to make it work.

Then in my codebase I have a loop where I execute against every path specified in the openapi spec, I want to pass the path, it would look like this:
```
for path in paths {
        let task = GooseTask::new(move |s| std::boxed::Box::pin(website_with_path(s, &path))));
        task_set = task_set.register_task(task);
}

goose_attack
    .register_taskset(
        task_set)
    .execute();
```

And it almost works. If instead of passing path to the task, I pass `""`, it works. But with path there is a complain that the closure does not map the function signature.

As I said, probably is my inability with Rust, I'm clueless on how to pursue. 
Still I thought it would be useful to write an example on how to do this as I would imagine programmatically setting paths should be a common case.
I am willing to write a minimal example once I can make it work.

jeremyandrews:
What about loading the paths in the task itself, instead of trying to pass it into the task?

ie, instead of hard coding the path "/" here:
https://github.com/tag1consulting/goose/blob/master/examples/simple.rs#L50

Instead load one or more paths from your file?

LionsAd:
I think you need another closure that has a function signature of `func(c: GooseClient)`.

I would first create the closure (and move the path inside it).

Then you pass this closure to task! and it should work unless I miss something.

JordiPolo:
Thanks for the quick responses!

@jeremyandrews there's a lot of computation between reading the original files and getting the callable urls with proper variables filled in. The only way would be have them somewhere accessible by the task. 

@LionsAd  Not 100% sure of what you meant but if you meant this:
```rust
        let my_c =  |client: &GooseClient |  async {  client.get(&path); };
        let task = task!(my_c);
```
I tried and got:
```
[...]

PR #120, opened 2020-07-28
Make it possible to pass closure to GooseTask::new

LionsAd:
This works perfectly now - as proven by the CI runs.

The tests error I first ran into is described in #122 and led to a stack overflow error, but was unrelated to things.

Closes: #76

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -1919,7 +1919,7 @@ pub fn get_base_url(
 }
 
 /// The function type of a goose task function.
-type GooseTaskFunction = Arc<
+pub type GooseTaskFunction = Arc<
     dyn for<'r> Fn(&'r GooseUser) -> Pin<Box<dyn Future<Output = GooseTaskResult> + Send + 'r>>
         + Send
         + Sync,
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -304,7 +304,9 @@ static APP_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_P
 #[macro_export]
 macro_rules! task {
     ($task_func:ident) => {
-        GooseTask::new(move |s| std::boxed::Box::pin($task_func(s)))
+        GooseTask::new(std::sync::Arc::new(move |s| {
+            std::boxed::Box::pin($task_func(s))
+        }))
     };
 }
 
@@ -1916,6 +1918,13 @@ pub fn get_base_url(
     }
 }
 
+/// The function type of a goose task function.
+type GooseTaskFunction = Arc<
+    dyn for<'r> Fn(&'r GooseUser) -> Pin<Box<dyn Future<Output = GooseTaskResult> + Send + 'r>>
+        + Send
+        + Sync,
+>;
+
 /// An individual task within a `GooseTaskSet`.
 #[derive(Clone)]
 pub struct GooseTask {
@@ -1932,15 +1941,10 @@ pub struct GooseTask {
     /// A flag indicating that this task runs when the user stops.
     pub on_stop: bool,
     /// A required function that is executed each time this task runs.
-    pub function:
-        for<'r> fn(&'r GooseUser) -> Pin<Box<dyn Future<Output = GooseTaskResult> + Send + 'r>>,
+    pub function: GooseTaskFunction,
 }
 impl GooseTask {
-    pub fn new(
-        function: for<'r> fn(
-            &'r GooseUser,
-        ) -> Pin<Box<dyn Future<Output = GooseTaskResult> + Send + 'r>>,
-    ) -> Self {
+    pub fn new(function: GooseTaskFunction) -> Self {
         trace!("new task");
         GooseTask {
             tasks_index: usize::max_value(),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1421,7 +1421,7 @@ impl GooseAttack {
                         self.host.clone(),
                     )?;
                     let user = GooseUser::single(base_url, &self.configuration)?;
-                    let function = t.function;
+                    let function = &t.function;
                     let _ = function(&user).await;
                 }
                 // No test_start_task defined, nothing to do.
@@ -1764,7 +1764,7 @@ impl GooseAttack {
                     )?;
                     // Create a one-time-use user to run the test_stop_task.
                     let user = GooseUser::single(base_url, &self.configuration)?;
-                    let function = t.function;
+                    let function = &t.function;
                     let _ = function(&user).await;
                 }
                 // No test_stop_task defined, nothing to do.
--- a/src/user.rs
+++ b/src/user.rs
@@ -38,7 +38,7 @@ pub async fn user_main(
             for task_index in &sequence {
                 // Determine which task we're going to run next.
                 let thread_task_name = &thread_task_set.tasks[*task_index].name;
-                let function = &thread_task_set.tasks[*task_index].function;
+                let function = &thread_task_set.tasks[*task_index].function.clone();
                 debug!(
                     "launching on_start {} task from {}",
                     thread_task_name, thread_task_set.name
@@ -88,7 +88,7 @@ pub async fn user_main(
         let thread_weighted_task =
             thread_user.weighted_tasks[weighted_bucket][weighted_bucket_position];
         let thread_task_name = &thread_task_set.tasks[thread_weighted_task].name;
-        let function = &thread_task_set.tasks[thread_weighted_task].function;
+        let function = &thread_task_set.tasks[thread_weighted_task].function.clone();
         debug!(
             "launching {} task from {}",
             thread_task_name, thread_task_set.name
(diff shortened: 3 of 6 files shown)


LionsAd:
This is so weird -- test fail locally for me, but pass on CI ...

src/lib.rs
```diff
@@ -1421,7 +1421,7 @@ impl GooseAttack {
                         self.host.clone(),
                     )?;
                     let user = GooseUser::single(base_url, &self.configuration)?;
-                    let function = t.function;
+                    let function = &t.function;
```
LionsAd, line 1424: We don’t need to clone that because rust can statically determine that this can never be accessed in parallel and automatically derefs it **pretty slick**.

src/user.rs
```diff
@@ -88,7 +88,7 @@ pub async fn user_main(
         let thread_weighted_task =
             thread_user.weighted_tasks[weighted_bucket][weighted_bucket_position];
         let thread_task_name = &thread_task_set.tasks[thread_weighted_task].name;
-        let function = &thread_task_set.tasks[thread_weighted_task].function;
+        let function = &thread_task_set.tasks[thread_weighted_task].function.clone();
```
LionsAd, line 91: This is the only real change needed for Arc - Rust could determine that due to the await something else can run during this.

tests/one_taskset.rs
```diff
@@ -41,14 +42,31 @@ fn test_single_taskset() {
...
+                let _goose = user.get(path).await?;
+                Ok(())
+            })
+        });
+
+        let mut task = GooseTask::new(closure).set_weight(weight).unwrap();
+        // @todo Cannot use taskset.register_task as that won't work in a loop.
+        task.tasks_index = taskset.tasks.len();
+        taskset.tasks.push(task);
+    }
```
LionsAd, line 64: And here is a first programmatic test 🎉

src/goose.rs
```diff
@@ -304,7 +304,9 @@ static APP_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_P
 #[macro_export]
 macro_rules! task {
     ($task_func:ident) => {
-        GooseTask::new(move |s| std::boxed::Box::pin($task_func(s)))
+        GooseTask::new(std::sync::Arc::new(move |s| {
```
jeremyandrews, line 307: we already `use std::sync::Arc;` in this file, so you can just `Arc::new()` for consistency

src/goose.rs
```diff
@@ -1916,6 +1918,13 @@ pub fn get_base_url(
     }
 }
 
+/// The function type of a goose task function.
+pub type GooseTaskFunction = Arc<
+    dyn for<'r> Fn(&'r GooseUser) -> Pin<Box<dyn Future<Output = GooseTaskResult> + Send + 'r>>
+        + Send
+        + Sync,
+>;
+
```
jeremyandrews, line 1927: :+1:  definitely makes sense to declare it as a type.

src/lib.rs
jeremyandrews: rust++ :smile:

src/user.rs
jeremyandrews: It will be interesting to see if there's any overhead from this change. I'll aim to do load testing today to compare.

tests/one_taskset.rs
jeremyandrews: Can we split this out into its own test function, perhaps `test_single_taskset_closure` or `test_closure`, instead of adding this to an existing test? As all tests within a single file run in parallel, it helps to see what's actually failing when tests fail. Plus, integration tests can be useful demonstrations of how to do something: for someone needing to use closures we can direct them to the closure-specific test. (And perhaps add a second test that demonstrates/tests closures together with passing in functions?)

tests/one_taskset.rs
```diff
@@ -41,14 +42,33 @@ fn test_single_taskset() {
...
+    let mut taskset = GooseTaskSet::new("LoadTest");
+
+    while let Some(item) = paths_and_weights.pop() {
+        let path = item.0;
+        let weight = item.1;
+
+        let closure: GooseTaskFunction = Arc::new(move |user| {
+            Box::pin(async move {
+                let _goose = user.get(path).await?;
+                Ok(())
```
jeremyandrews, line 56: nit, please add a blank line before the `Ok(())` to match the rest of our codebase (theoretically).

tests/one_taskset.rs
```diff
@@ -41,14 +42,33 @@ fn test_single_taskset() {
...
+
     let goose_stats = crate::GooseAttack::initialize_with_config(config.clone())
         .setup()
         .unwrap()
-        .register_taskset(
-            taskset!("LoadTest")
-                .register_task(task!(get_index).set_weight(9).unwrap())
-                .register_task(task!(get_about).set_weight(3).unwrap()),
-        )
+        .register_taskset(taskset)
```
jeremyandrews, line 71: As noted above, this removes an existing test and replaces it with a new one. Instead it should leave the old existing test, and add a new one in the same file with a different name, so we're testing multiple code paths.

src/goose.rs
LionsAd: We never want to use a short form in a macro ;) - I tried that and it did not end well ...

tests/one_taskset.rs
LionsAd: Yep - sounds good.

I will copy the test.

tests/one_taskset.rs
LionsAd: That should have been a verbatim copy from above - will check.

LionsAd:
```
pub async fn get_index(user: &GooseUser) -> GooseTaskResult {
    let _goose = user.get(INDEX_PATH).await?;
    Ok(())
}
pub async fn get_about(user: &GooseUser) -> GooseTaskResult {
    let _goose = user.get(ABOUT_PATH).await?;
    Ok(())
}
```

was the inspiration.

tests/one_taskset.rs
LionsAd: done

tests/one_taskset.rs
LionsAd: done

tests/one_taskset.rs
LionsAd: Fixed in the new test

tests/one_taskset.rs
```diff
@@ -153,3 +155,96 @@ fn test_single_taskset_empty_config_host() {
...
+            })
+        });
+
+        let task = GooseTask::new(closure).set_weight(weight).unwrap();
+        // We need to do the variable dance as taskset.register_task returns self and hence moves
+        // self out of `taskset`. By storing it in a new local variable and then moving it over
+        // we can avoid that error.
+        let new_taskset = taskset.register_task(task);
+        taskset = new_taskset;
+    }
```
LionsAd, line 204: And here is a first programmatic test 🎉

src/user.rs
LionsAd: Yes - indeed. In theory this is only: atomic_inc() + atomic_dec() which compared to a .read() lock has neglect-able performance impact, but we'll see.

tests/one_taskset.rs
```diff
@@ -153,3 +155,96 @@ fn test_single_taskset_empty_config_host() {
...
+        .create_on(&server);
+
+    let mut config = common::build_configuration(&server);
+    config.no_stats = false;
+    // Start users in .5 seconds.
+    config.users = Some(2);
+    config.hatch_rate = 4;
+    config.status_codes = true;
+
+    let mut paths_and_weights = vec![(INDEX_PATH, 9), (ABOUT_PATH, 3)];
```
LionsAd, line 183: Btw - that can be follow up, but we seem to never check these numbers -- only that the statistics match the actual test.

src/goose.rs
jeremyandrews: Ah, that makes sense!

tests/one_taskset.rs
jeremyandrews: We should open an issue: there's a todo item in the back of my mind to validate each and every configuration option is working as intended. (I am most likely to tackle this when exposing setters/getters)

tests/one_taskset.rs
jeremyandrews: fantastic! :+1:

tests/one_taskset.rs
LionsAd: Oh - I see now where we do that just with the 1/3 pattern ...

jeremyandrews:
Performance comparisons between the `master` branch and your PR:
- https://docs.google.com/document/d/1xZPCBKRVs74kqg74sn6O4LD2pX_B8Fb_6oaoIl2hKrI/edit#
- https://docs.google.com/spreadsheets/d/1yPaYO37wx7MmGfuvoMLNDI8V2w6TFP7jE2ObV-5OKK0/edit#gid=0

src/user.rs
jeremyandrews: no significant change in performance :+1:

PR #126, opened 2020-07-29
Add a closure example

LionsAd:


examples/simple_closure.rs
```diff
@@ -0,0 +1,57 @@
...
+
+fn main() -> Result<(), GooseError> {
+    let mut taskset = taskset!("WebsiteUser")
+        // After each task runs, sleep randomly from 5 to 15 seconds.
+        .set_wait_time(5, 15)?;
+
+    let paths = vec!["/", "/about", "/our-team"];
+    for request_path in paths {
+        // Clone the string so that it can be moved into the closure.
+        let path = request_path;
```
jeremyandrews, line 32: Are we now cloning the reference, not the string?

LionsAd: We actually don't need to clone as I used paths and not &paths now ...

PR #128, opened 2020-07-29
properly display errors

### LINKED ISSUE #127: [BUG] [CRITICAL] Infinite loop for Display in GooseError and GooseTaskError
See comment here:

https://github.com/tag1consulting/goose/pull/113/files#r462282759

jeremyandrews:
Display an actual/useful error.

Fix #127

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -375,7 +375,11 @@ impl GooseTaskError {
 // Define how to display errors.
 impl fmt::Display for GooseTaskError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "GooseTaskError: {}", self.describe())
+        match *self {
+            GooseTaskError::Reqwest(ref source) => write!(f, "GooseTaskError: {} ({})", self.describe(), source),
+            GooseTaskError::Url(ref source) => write!(f, "GooseTaskError: {} ({})", self.describe(), source),
+            _ => write!(f, "GooseTaskError: {}", self.describe())
+        }
     }
 }
 
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -376,9 +376,22 @@ impl GooseTaskError {
 impl fmt::Display for GooseTaskError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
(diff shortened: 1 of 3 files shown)


src/goose.rs
```diff
@@ -356,11 +356,26 @@ pub enum GooseTaskError {
     /// is available in `.method`.
     InvalidMethod { method: Method },
 }
+impl GooseTaskError {
+    fn describe(&self) -> &str {
+        match *self {
+            GooseTaskError::Reqwest(_) => "reqwest::Error",
+            GooseTaskError::Url(_) => "url::ParseError",
```
LionsAd, line 363: This is okay, but I would rather that we pass through information from child errors still, e.g. with `:`.

src/goose.rs
```diff
@@ -356,11 +356,26 @@ pub enum GooseTaskError {
...
+            GooseTaskError::InvalidMethod { .. } => "Unrecognized HTTP request method",
+        }
+    }
+}
 
 // Define how to display errors.
 impl fmt::Display for GooseTaskError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Display::fmt(&self, f)
+        write!(f, "GooseTaskError: {}", self.describe())
```
LionsAd, line 378: What about we do:

```
if Some(err) = self.source() {
  err.fmt(f);
}
```

src/goose.rs
jeremyandrews: the proposed implementation didn't work, I tried an alternative

src/lib.rs
```diff
@@ -412,11 +412,34 @@ pub enum GooseError {
...
-        fmt::Display::fmt(&self, f)
+        match *self {
+            GooseError::Io(ref source) => write!(f, "GooseError: {} ({})", self.describe(), source),
+            GooseError::Reqwest(ref source) => {
+                write!(f, "GooseError: {} ({})", self.describe(), source)
+            }
+            GooseError::InvalidHost {
+                ref parse_error, ..
+            } => write!(f, "GooseError: {} ({})", self.describe(), parse_error),
+            _ => write!(f, "GooseError: {}", self.describe()),
```
LionsAd, line 441: It's a bummer all this double-work is needed, but it's good enough for now. We can always simplify later.

jeremyandrews: :+1:

PR #138, opened 2020-08-02
subtract overhead to avoid time drift

### LINKED ISSUE #102: subtract overhead from delay between starting users and throttling requests
When starting users, and when throttling request, the code is currently naive about how long it delays, always sleeping for a set amount of time regardless of how long is spent on the required logic. The delay logic should be enhanced to subtract overhead.

In the current implementation, there's a very slow drift visible if you enable the statistics logs and review the timestamps. When fixed, requests should be more consistently grouped into a steady cadence that's not drifting as users are launched and/or a throttled test runs.

jeremyandrews:
In smaller load tests the drift is barely perceptible. But in larger tests the drift becomes more pronounced. This PR subtracts the drift from delays in two loops: launching users, and throttling requests.

Fixes #102

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1475,6 +1475,11 @@ impl GooseAttack {
             mpsc::UnboundedSender<GooseRawRequest>,
             mpsc::UnboundedReceiver<GooseRawRequest>,
         ) = mpsc::unbounded_channel();
+
+        // Update user_thread_started after each delay. This is used to prevent a time
+        // drift due to the time it takes to launch a new thread.
+        let mut user_thread_started = time::Instant::now();
+
         // Spawn users, each with their own weighted task_set.
         for mut thread_user in self.weighted_users.clone() {
             // Stop launching threads if the run_timer has expired, unwrap is safe as we only get here if we started.
@@ -1539,7 +1544,10 @@ impl GooseAttack {
             users.push(user);
             self.stats.users += 1;
             debug!("sleeping {:?} milliseconds...", sleep_duration);
-            tokio::time::delay_for(sleep_duration).await;
+            tokio::time::delay_for(sleep_duration - user_thread_started.elapsed()).await;
+
+            // Update user_thread_started after each delay.
+            user_thread_started = time::Instant::now();
         }
         if self.configuration.worker {
             info!(
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1476,9 +1476,10 @@ impl GooseAttack {
             mpsc::UnboundedReceiver<GooseRawRequest>,
         ) = mpsc::unbounded_channel();
 
-        // Update user_thread_started after each delay. This is used to prevent a time
-        // drift due to the time it takes to launch a new thread.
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1539,7 +1544,10 @@ impl GooseAttack {
             users.push(user);
             self.stats.users += 1;
             debug!("sleeping {:?} milliseconds...", sleep_duration);
-            tokio::time::delay_for(sleep_duration).await;
+            tokio::time::delay_for(sleep_duration - user_thread_started.elapsed()).await;
```
LionsAd, line 1547: I think this belongs in a small async helper function - because it should also not sleep a negative time.

Also: Could consider printing the drift corrected sleep_duration.

src/lib.rs
```diff
@@ -1539,7 +1544,10 @@ impl GooseAttack {
...
+
+            // Update user_thread_started after each delay.
+            user_thread_started = time::Instant::now();
```
LionsAd, line 1550: The helper can encapsulate this update, too.

src/lib.rs
jeremyandrews: I could do it at `debug!()` -- when writing this I started with printing it and because of the granularity of `.elapsed()` there's always a very small drift (microseconds), so it's a lot of "noise" in the end. Moving it into a helper makes sense.

jeremyandrews: BTW: My understanding is the duration between two instances is already guaranteed to never be negative: https://doc.rust-lang.org/std/time/struct.Instant.html

jeremyandrews: Ah, I see what you mean: if drift gets really big. Okay, handled in `util::sleep_minus_drift()`

src/lib.rs
jeremyandrews: implemented

src/util.rs
```diff
@@ -46,6 +46,20 @@ pub fn parse_timespan(time_str: &str) -> usize {
     }
 }
 
+/// Sleep for a specified duration, minus an time spent doing other things.
+pub async fn sleep_minus_drift(
```
LionsAd, line 50: This is probably a great candidate for:

[#inline(always)]

as it makes sense to inline this to reduce also the effects of the function call on the drift itself.

jeremyandrews: Reading https://doc.rust-lang.org/reference/attributes/codegen.html and https://internals.rust-lang.org/t/when-should-i-use-inline/598/3 it seems the general consensus is that the compiler does a good job of this automatically if/when it makes sense? Granted, in all cases it's just a suggestion, which the compiler may still ignore.

I'm happy to add if you're confident it's the right thing to do; it's outside my experience.

jeremyandrews: Does the async nature of this function defeat the purpose of inlining it? We will nearly always sleep and await.

LionsAd: That’s a very good point - inlining on async fns does actually not work, yet per https://github.com/rust-lang/rust/issues/63647, so let’s leave this as is.

RTBM

PR #140, opened 2020-08-11
introduce per-task stats

### LINKED ISSUE #107: also optionally provide per-task-function statistics
Currently statistics are per-request. With #103 tasks return a Result() giving us insight into how often the entire Task succeeds and/or fails. This issue is to optionally provide statistics about this at the end of a load test.

One specific use-case of this feature would be to confirm that all task sets have been assigned to a `GooseUser`, and that all contained tasks have actually run. These statistics would provide insight into how often each Task is running, and whether any were skipped (ie, due to weighting and too few users).

jeremyandrews:
Optionally tracks per-task statistics (in addition to the already implemented per-request statistics). Fixes #107 

Adds the following table:
```
=== PER TASK STATISTICS ===
 Name                    | # times run    | # fails        | task/s | fail/s
 ----------------------------------------------------------------------------- 
 1: AnonBrowsingUser     |
   1: (Anon) front page  | 1,730          | 234 (13.5%)    | 57     | 7    
   2: (Anon) node page   | 1,162          | 0 (0%)         | 38     | 0    
   3: (Anon) user page   | 347            | 0 (0%)         | 11     | 0    
 2: AuthBrowsingUser     |
   1: (Auth) login       | 2              | 0 (0%)         | 0      | 0    
   2: (Auth) front page  | 338            | 56 (16.6%)     | 11     | 1    
   3: (Auth) node page   | 226            | 0 (0%)         | 7      | 0    
   4: (Auth) user page   | 68             | 0 (0%)         | 2      | 0    
   5: (Auth) comment f.. | 67             | 12 (17.9%)     | 2      | 0    
 ------------------------+----------------+----------------+--------+--------- 
 Aggregated              | 3,940          | 302 (7.7%)     | 131    | 10   
-------------------------------------------------------------------------------
 Name                    | Avg (ms)   | Min        | Max        | Median    
 ----------------------------------------------------------------------------- 
 1: AnonBrowsingUser     |
   1: (Anon) front page  | 110        | 3          | 1163       | 110       
   2: (Anon) node page   | 38         | 3          | 1478       | 33        
   3: (Anon) user page   | 27         | 3          | 1057       | 21        
 2: AuthBrowsingUser     |
   1: (Auth) login       | 89         | 80         | 99         | 80        
   2: (Auth) front page  | 127        | 4          | 1060       | 130       
   3: (Auth) node page   | 34         | 4          | 250        | 31        
   4: (Auth) user page   | 24         | 6          | 111        | 19        
   5: (Auth) comment f.. | 129        | 6          | 1056       | 130       
 ------------------------+------------+------------+------------+------------- 
 Aggregated              | 77         | 3          | 1478       | 57        
```

The "simple" example doesn't name its tasks, so it just shows numbers. Here's the complete statistics returned by Goose by default. Note that in the final statistics, task 1 shows as having run 0 times, this is because it's an `on_start` task so it only runs one time, and this happened before the statistics were reset.

```
$ cargo run --example simple -- --host http://local.dev -t30 --only-summary 
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/examples/simple --host 'http://local.dev' -t30 --only-summary`

=== PER TASK STATISTICS ===
 Name                    | # times run    | # fails        | task/s | fail/s
 ----------------------------------------------------------------------------- 
 1: WebsiteUser          |
   1:                    | 8              | 0 (0%)         | 1      | 0    
   2:                    | 8              | 0 (0%)         | 1      | 0    
   3:                    | 1              | 0 (0%)         | 0      | 0    
 ------------------------+----------------+----------------+--------+--------- 
 Aggregated              | 17             | 0 (0%)         | 2      | 0    
-------------------------------------------------------------------------------
 Name                    | Avg (ms)   | Min        | Max        | Median    
 ----------------------------------------------------------------------------- 
 1: WebsiteUser          |
   1:                    | 17         | 16         | 19         | 17        
   2:                    | 20         | 2          | 139        | 3         
   3:                    | 9          | 9          | 9          | 9         
 ------------------------+------------+------------+------------+------------- 
 Aggregated              | 18         | 2          | 139        | 16        

=== PER REQUEST STATISTICS ===
 Name                    | # reqs         | # fails        | req/s  | fail/s
 ----------------------------------------------------------------------------- 
 GET /                   | 8              | 0 (0%)         | 1      | 0    
 GET /about/             | 1              | 1 (100%)       | 0      | 0    
 POST /login             | 8              | 8 (100%)       | 1      | 1    
 ------------------------+----------------+----------------+--------+--------- 
 Aggregated              | 17             | 9 (52.9%)      | 2      | 1    
-------------------------------------------------------------------------------
 Name                    | Avg (ms)   | Min        | Max        | Median    
 ----------------------------------------------------------------------------- 
 GET /                   | 19         | 2          | 138        | 3         
 GET /about/             | 9          | 9          | 9          | 9         
 POST /login             | 17         | 15         | 19         | 16        
 ------------------------+------------+------------+------------+------------- 
 Aggregated              | 17         | 2          | 138        | 15        

All 8 users hatched, resetting statistics (disable with --no-reset-stats).


=== PER TASK STATISTICS ===
 Name                    | # times run    | # fails        | task/s | fail/s
 ----------------------------------------------------------------------------- 
 1: WebsiteUser          |
   1:                    | 0              | 0 (0%)         | 0      | 0    
   2:                    | 8              | 0 (0%)         | 0      | 0    
   3:                    | 14             | 0 (0%)         | 0      | 0    
 ------------------------+----------------+----------------+--------+--------- 
 Aggregated              | 22             | 0 (0%)         | 0      | 0    
-------------------------------------------------------------------------------
 Name                    | Avg (ms)   | Min        | Max        | Median    
 ----------------------------------------------------------------------------- 
 1: WebsiteUser          |
   1:                    | 0          | 0          | 0          | 0         
   2:                    | 16         | 8          | 69         | 9         
   3:                    | 8          | 6          | 11         | 8         
 ------------------------+------------+------------+------------+------------- 
 Aggregated              | 11         | 6          | 69         | 9         

=== PER REQUEST STATISTICS ===
 Name                    | # reqs         | # fails        | req/s  | fail/s
 ----------------------------------------------------------------------------- 
 GET /                   | 8              | 0 (0%)         | 0      | 0    
 GET /about/             | 14             | 14 (100%)      | 0      | 0    
 ------------------------+----------------+----------------+--------+--------- 
 Aggregated              | 22             | 14 (63.6%)     | 0      | 0    
-------------------------------------------------------------------------------
 Name                    | Avg (ms)   | Min        | Max        | Median    
 ----------------------------------------------------------------------------- 
 GET /                   | 16         | 8          | 69         | 8         
 GET /about/             | 8          | 5          | 10         | 8         
 ------------------------+------------+------------+------------+------------- 
 Aggregated              | 11         | 5          | 69         | 8         
-------------------------------------------------------------------------------
 Slowest page load within specified percentile of requests (in ms):
 Name                    | 50%    | 75%    | 98%    | 99%    | 99.9%  | 99.99%
 ----------------------------------------------------------------------------- 
 GET /                   | 8      | 9      | 69     | 69     | 69     |     69
 GET /about/             | 8      | 9      | 10     | 10     | 10     |     10
 ------------------------+--------+--------+--------+--------+--------+------- 
 Aggregated              | 8      | 9      | 69     | 69     | 69     |     69
```


LionsAd:
I would add some in between headings, e.g.

```

=== PER TASK STATISTICS ===

[...]

=== PER REQUEST STATISTICS ===

```

src/lib.rs
```diff
@@ -1778,18 +1825,32 @@ impl GooseAttack {
...
-                    // As worker, push request statistics up to manager.
+                    // As worker, push all statistics up to manager.
                     if self.configuration.worker {
-                        // Push request statistics to manager process.
                         worker::push_stats_to_manager(
                             &socket.clone().unwrap(),
-                            &self.stats.requests.clone(),
+                            Some(self.stats.requests.clone()),
+                            Some(self.stats.tasks.clone()),
                             true,
```
LionsAd, line 1854: This I like 👍

src/manager.rs
```diff
@@ -329,23 +373,27 @@ pub async fn manager_main(mut goose_attack: GooseAttack) -> GooseAttack {
                     // We need another worker, accept the connection.
                     else {
                         // Validate worker load test hash.
-                        match requests.get("load_test_hash") {
-                            Some(r) => {
-                                if r.load_test_hash != goose_attack.stats.hash {
+                        if let Some(requests) = gaggle_stats.requests {
```
LionsAd, line 376: This case is actually invalid and a case for panic() as well, so I would either use `let requests = gaggle_stats.requests.expect('...')` OR `let requests = gaggle_stats.requests.unwrap()?;` - if we can pass the error up.

Adding more indentation does actually hide an error case.

jeremyandrews:
> I would add some in between headings, e.g.

Added, see updated pull-request description which now includes the suggested headers.

src/goose.rs
```diff
@@ -1529,16 +1533,16 @@ impl GooseUser {
 
         // Send raw request object to parent if we're tracking statistics.
         if !self.config.no_stats {
-            self.send_to_parent(&raw_request)?;
+            self.send_request_stats_to_parent(&raw_request)?;
```
LionsAd, line 1536: As spoken: `self.send_to_parent` is better as there is no other function, so there is no ambiguity,

src/goose.rs
```diff
@@ -911,7 +912,9 @@ pub struct GooseUser {
     /// Normal tasks are optionally throttled, test_start and test_stop tasks are not.
     pub is_throttled: bool,
     /// Channel to parent.
-    pub parent: Option<mpsc::UnboundedSender<GooseRawRequest>>,
+    pub parent_request_stats: Option<mpsc::UnboundedSender<GooseRawRequest>>,
+    /// Channel to parent.
+    pub parent_task_stats: Option<mpsc::UnboundedSender<GooseRawTask>>,
```
LionsAd, line 917: I thought about it and we definitely want the enum solution:

enum GooseMetric {
  Request(GooseRawRequest),
  Task(GooseRawTask),
  TaskSet(GooseRawTaskSet)
}
```

etc.

The storage is pretty similar, so an enum is the perfect way to always can just do:

```
parent.send(&GooseMetric::Request(req));
```

etc.

I feel it is worth it to simplify it and make it even simpler to add other statistics in the future.

Maybe even Boxed user defined ones.

File: src/manager.rs (hunk 2 above)
Comment (LionsAd) [reply src/manager.rs:376]: To explain:

```
if let Some(requests) = gaggle_stats.requests
```

is plain wrong, let's assume gaggle_stats.requests is None, then the error is silently discarded.

You want:

```
match gaggle_stats.requests.expect('Request must not be empty').get("load_test_hash") {
```

instead.

File: src/user.rs [hunk 5]
```diff
@@ -187,3 +195,33 @@ pub async fn user_main(
...
+    );
+    let success = function(&thread_user).await.is_ok();
+    raw_task.set_time(started.elapsed().as_millis(), success);
+
+    // Exit if all statistics or task statistics are disabled.
+    if thread_user.config.no_stats || thread_user.config.no_task_stats {
+        return;
+    }
+
+    // Otherwise send statistics to parent (which doesn't exist during testing).
```
LionsAd, line 222: Should there be a mock parent for when those options are enabled?

src/worker.rs
```diff
@@ -222,20 +230,21 @@ pub async fn worker_main(goose_attack: &GooseAttack) -> GooseAttack {
...
+    debug!("[{}] pushing stats to manager", get_worker_id(),);
     let mut message = Message::new().unwrap();
-    serde_cbor::to_writer(&mut message, requests)
+
+    let gaggle_stats = GaggleStats { requests, tasks };
+
+    serde_cbor::to_writer(&mut message, &gaggle_stats)
         .map_err(|error| eprintln!("{:?} worker_id({})", error, get_worker_id()))
-        .expect("failed to serialize empty Vec<GooseRequest>");
+        .expect("failed to serialize GooseRequest");
```
LionsAd, line 247: Should this say `GaggleStats` instead?

src/worker.rs
```diff
@@ -222,20 +230,21 @@ pub async fn worker_main(goose_attack: &GooseAttack) -> GooseAttack {
         .expect("failed to launch GooseAttack")
 }
 
+// Push request or task statistics to manager.
 pub fn push_stats_to_manager(
     manager: &Socket,
-    requests: &HashMap<String, GooseRequest>,
+    requests: Option<HashMap<String, GooseRequest>>,
+    tasks: Option<GooseTaskStats>,
```
LionsAd, line 237: While I like the option to make arguments optional, is there really a use-case where you want to send requests and not even an empty GooseTaskStats?

src/worker.rs
```diff
@@ -1,14 +1,22 @@
...
 use crate::manager::GooseUserInitializer;
+use crate::stats::GooseTaskStats;
 use crate::util;
 use crate::{get_worker_id, GooseAttack, GooseConfiguration, WORKER_ID};
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct GaggleStats {
+    pub requests: Option<HashMap<String, GooseRequest>>,
+    pub tasks: Option<GooseTaskStats>,
+}
```
LionsAd, line 18: While I like the option to make arguments optional, is there really a use-case where you want to send requests and not even an empty GooseTaskStats?

The option being implemented as enum means that you store space for both GooseTaskStats and the HashMap anyway - even if tasks was None.

src/user.rs
```diff
@@ -187,3 +195,33 @@ pub async fn user_main(
         );
     }
 }
+
+// Invoke the task function, collecting task statistics.
+async fn invoke_task_function(
+    function: &GooseTaskFunction,
+    thread_user: &GooseUser,
+    task_index: usize,
+    thread_task_name: &str,
```
LionsAd, line 204: 👍

src/manager.rs
jeremyandrews: The way the logic is written, requests won't be empty at this time, as we start by sending "fake packets". In any case, the Option<> goes away when we switch to an enum, so this will go away.

src/user.rs
jeremyandrews: I actually think the comment is wrong; I'll test and fix.

src/worker.rs
jeremyandrews: The enum will remove this Option<>, as we'd always be sending something. The Option<> only existed because we could send either alone, or both together.

src/worker.rs
jeremyandrews: Option<> was only necessary as we could send either requests, or tasks, or both. Reworked as an enum we won't need the Option anymore.

src/manager.rs
LionsAd: Then just use .unwrap() - let Some(x) is for cases where the None option is a valid case and mit one that can never happen.

src/lib.rs
jeremyandrews: Now that we're using an Enum, I've reworked instead to use a Vec() -- this will make it easier to add additional types of metrics in the future.

src/manager.rs
jeremyandrews: We don't want the Manager to panic if the wrong number of Workers are started, instead we have the extra Worker's panic -- the Manager and correct Workers can safely continue their load test.

This area of the codebase has been reworked to support Enums, and I think it's cleaner now.

src/goose.rs
jeremyandrews: restored

src/goose.rs
jeremyandrews: We now use Enums both for User threads pushing metrics to the Parent thread, as well as for Worker processes pushing metrics to the Manager process.

src/worker.rs
jeremyandrews: This is the new function signature:
```rust
pub fn push_metrics_to_manager(
    manager: &Socket,
    metrics: Vec<GooseMetrics>,
    get_response: bool,
) -> bool {
```

src/worker.rs
jeremyandrews: This now uses an Enum, defined as follow:
```rust
/// Workers send GooseMetrics to the Manager process to be aggregated together.
#[derive(Debug, Clone, Serialize, Deserialize)]
pub enum GooseMetrics {
    /// Load test hash, used to ensure all Workers are running the same load test.
    WorkerInit(u64),
    /// Goose request statistics.
    Requests(GooseRequestStats),
    /// Goose task statistics.
    Tasks(GooseTaskStats),
}
```

src/goose.rs
```diff
@@ -343,7 +344,7 @@ pub enum GooseTaskError {
     /// The `GooseRawRequest` that was not recorded can be extracted from the error
     /// chain, available inside `.source`.
     StatsFailed {
-        source: mpsc::error::SendError<GooseRawRequest>,
+        source: mpsc::error::SendError<GooseMetric>,
```
LionsAd, line 347: Love it 👍

src/goose.rs
```diff
@@ -1527,19 +1528,20 @@ impl GooseUser {
             }
         };
 
-        // Send raw request object to parent if we're tracking statistics.
+        // Send a copy of the raw request object to the parent process if
+        // we're tracking statistics.
         if !self.config.no_stats {
-            self.send_to_parent(&raw_request)?;
+            self.send_to_parent(GooseMetric::Request(raw_request.clone()))?;
```
LionsAd, line 1534: nit - for potential future follow-up:

The need for a different variable name as `request` would shadow the other `request` shows that the processing here likely belongs into a helper function.

jeremyandrews: Which are you suggesting to rename, `GooseMetric::Request`? Or `raw_request`? What would you rename it too? I'm not sure I'm understanding this comment.

src/lib.rs
```diff
@@ -1245,9 +1247,14 @@ impl GooseAttack {
...
         }
 
-        // Allocate a state for each of the users we are about to start.
         if !self.configuration.worker {
+            // Allocate a state for each of the users we are about to start.
             self.weighted_users = self.weight_task_set_users()?;
+
+            // Stand-alone and Manager processes can display metrics.
+            if !self.configuration.no_stats {
+                self.stats.display_metrics = true;
```
LionsAd, line 1256: Question: Do you also need to check for no_task_stats here?

LionsAd:
Looks fantastic now

src/worker.rs
jeremyandrews: It's now GooseMetric

src/lib.rs
jeremyandrews: No, this flag only gets set to True on the Manager or a Stand-Alone Goose process. However, it made me realize a different recent commit was wrong, undoing it here:
https://github.com/tag1consulting/goose/pull/140/commits/19c88928a779db85c33cdbd893855823b305ffac#diff-6a9cab52f3c4564d1db3838cb39ed73fR395

(We can't ONLY check `display_metrics`, we also have to check if `tasks` is empty or we display empty task stats.)

While testing this I also realized we were allowing `--no-task-stats` to run on the Worker, which is incorrect -- it should only be allowed on the Manager, also fixed.

src/goose.rs
LionsAd: `request` - the other invocations of `send_to_parent` use:

```rust
self.send_to_parent(GooseMetric::Request(request.clone()))?;
```

PR #146, opened 2020-08-21
standardize errors for consistency

### LINKED ISSUE #132: standardize error details
We have several variations on error details with the same intent:

- `--no-hash-check is only available to the manager`
- `--no-hash-check is only available when running in manager mode`
- `--users option only available to manager process`
- `--debug-log-file can only be enabled in stand-alone or worker mode`
- `--throttle-requests can only be enabled in stand-alone mode or worker mode`

[Fabian proposes a different wording](https://github.com/tag1consulting/goose/pull/129#pullrequestreview-459091587):

- `can only be set on the manager`

Often "only available to the manager" isn't correct, as the option is also available in stand-alone mode as well.  Maybe better, "this option can not be set on the worker" ?

jeremyandrews:
- Removed `Option<>` around `error.detail`, as there was only one error that didn't define it yet
 - Consistently defined and formatted `error.detail`; errors are now always a complete sentence
 - Updated documentation

This fixes #132.

Some sample errors:

```
$ cargo run --example simple -- --worker --host http://foo
   Compiling goose v0.10.0-dev (/home/jandrews/devel/rust/goose)
    Finished dev [unoptimized + debuginfo] target(s) in 7.87s
     Running `target/debug/examples/simple --worker --host 'http://foo'`
Error: InvalidOption { option: "--host", value: "http://foo", detail: "The --host option can not be set together with the --worker flag." }
```

```
$ cargo run --example simple -- --host foo
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/examples/simple --host foo`
Error: InvalidHost { host: "foo", detail: "There was a failure parsing the host specified with --host.", parse_error: RelativeUrlWithoutBase }
```

```
$ cargo run --example simple -- --manager --worker
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/examples/simple --manager --worker`
Error: InvalidOption { option: "--worker", value: "true", detail: "The --worker flag can not be set together with the --manager flag" }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
     o `--debug-log-format` became `--debug-format`
     o `--metrics-log-file` became `--metrics-file`
     o `--metrics-log-format` became `--metrics-format`
+ - reworded errors for consistency, made error.detail required
 
 ## 0.9.1 Aug 1, 2020
  - return `GooseStats` from `GooseAttack` `.execute()`
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ $ cargo run
    Compiling loadtest v0.1.0 (/home/jandrews/devel/rust/loadtest)
     Finished dev [unoptimized + debuginfo] target(s) in 3.56s
      Running `target/debug/loadtest`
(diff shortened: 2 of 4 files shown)


LionsAd (approved):
RTBM - I love the cleanup

PR #152, opened 2020-08-25
consistently build configuration from arguments

### LINKED ISSUE #144: not testing actual run-time flags and options
During a recent `--help` and general documentation cleanup, I reworked the Goose `--workers-expected` option to use the Structopt `required_if` attribute, as it is required if `--master` is enabled. The implementation seems to actually be Clap: [`required_if`](https://docs.rs/clap/*/clap/struct.Arg.html#method.required_if).

Possibly it's as simple as `--master` is a flag, not an option, and `required_if` only works with options -- this isn't clear to me from the documentation.

The bigger problem is that our tests did not catch this, only manually running Goose caught it. This issue is to follow up so we are properly testing configuration options to avoid something similar to this from slipping through again. (When actually running Goose, we use `GooseConfiguration::from_args()` to instantiate the configuration struct. When testing, we use `tests/common/build_configuration()` to manually construct a configuration struct.

jeremyandrews:
Consistently build configuration from arguments:
 - remove cases where parameters were manually modified directly in the configuration object
 - remove cases where defaults were automagically derived

This fixes #144.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
     o rewrote/simplified configuration descriptions to fit standard console width
  - update prelude documentation
  - increase precision of metrics for smaller values
+ - consistently build configuration from arguments
 
 ## 0.9.1 Aug 1, 2020
  - return `GooseStats` from `GooseAttack` `.execute()`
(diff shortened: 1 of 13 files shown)


LionsAd (commented):
Review

LionsAd (approved):
RTBM

PR #153, opened 2020-08-26
introduce generic set_default(), also replaces set_host()

jeremyandrews:
- replaced `GooseAttack::set_host()` with more generic `GooseAttack::set_default()`, exposing the following defaults when writing a load test:
    o default host: `GooseDefault::Host` (&str)
    o default users to start: `GooseDefault::Users` (usize)
    o default users to start per second: `GooseDefault::HatchRate` (usize)
    o default number of seconds for test to run: `GooseDefault::RunTime` (usize)
    o default log level: `GooseDefault::LogLevel` (usize)
    o default log file name: `GooseDefault::LogFile` (&str)
    o default verbosity: `GooseDefault::Verbose` (usize)
    o default to only printing final summary metrics: `GooseDefault::OnlySummary` (bool)
    o default to not resetting metrics after all users start: `GooseDefault::NoResetMetrics` (bool)
    o default to not tracking metrics: `GooseDefault::NoMetrics` (bool)
    o default to not tracking task metrics: `GooseDefault::NoTaskMetrics` (bool)
    o default metrics log file name: `GooseDefault::MetricsFile` (&str)
    o default metrics log file format: `GooseDefault::MetricsFormat` (&str)
    o default debug log file name: `GooseDefault::DebugFile` (&str)
    o default debug log file format: `GooseDefault::DebugFormat` (&str)
    o default to tracking status codes: `GooseDefault::StatusCodes` (bool)
    o default maximum requests per second: `GooseDefault::ThrottleRequests` (usize)
    o default to following redirect of base_url: `GooseDefault::StickyFollow` (bool)
    o default to enabling Manager mode: `GooseDefault::Manager` (bool)
    o default number of Workers to expect: `GooseDefault::ExpectWorkers` (usize)
    o default to ignoring load test checksum: `GooseDefault::NoHashCheck` (bool)
    o default host to bind Manager to: `GooseDefault::ManagerBindHost` (&str)
    o default port to bind Manager to: `GooseDefault::ManagerBindPort` (usize)
    o default to enabling Worker mode: `GooseDefault::Worker` (bool)
    o default host for Worker to connect to: `GooseDefault::ManagerHost` (&str)
    o default port for Worker to connect to: `GooseDefault::ManagerPort` (usize)
 - overloaded set_default() with a new GooseDefaultType trait, supporting `&str`, `usize` and `bool` type defaults


src/lib.rs
```diff
@@ -510,11 +645,14 @@ impl GooseAttack {
...
             weighted_users: Vec::new(),
-            host: None,
+            defaults: GooseDefaults::default(),
             configuration: GooseConfiguration::parse_args_default_or_exit(),
             number_of_cpus: num_cpus::get(),
             run_time: 0,
             users: 0,
+            hatch_rate: 0,
+            throttle_requests: 0,
+            attack_mode: GooseMode::Undefined,
```
jeremyandrews, line 655: @LionsAd Here are the three new things that converged into `GooseAttack`, instead of over-loading configuration.

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+                if user_count >= self.users {
+                    trace!("created {} weighted_users", user_count);
+                    return Ok(weighted_users);
+                }
+            }
+        }
+    }
+
+    // Configure which mode this GooseAttack will run in.
+    fn set_attack_mode(&mut self) -> Result<(), GooseError> {
```
jeremyandrews, line 976: @LionsAd I like having an `attack_mode`, but this could just live in the configuration struct if that's somehow better.

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+                    detail: "The --no-hash-check flag can not be set without also setting the --manager flag.".to_string(),
+                });
+            }
+        }
+
+        Ok(())
+    }
+
+    // Determine how many workers to expect.
+    fn set_expect_workers(&mut self) -> Result<(), GooseError> {
```
jeremyandrews, line 1120: @LionsAd In this case we overload the configuration array (and I suspect this is the preferred way to handle defaults).

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+            }
+        }
+
+        // Overload configuration.expect_workers to make available in Worker process.
+        self.configuration.expect_workers = expect_workers;
+
+        Ok(())
+    }
+
+    fn set_gaggle_host_and_port(&mut self) -> Result<(), GooseError> {
```
jeremyandrews, line 1169: @LionsAd Again we overload the configuration structure (this simplifies passing the information around in a Gaggle)

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+                            .to_string(),
+                });
+            }
+        }
+
+        Ok(())
+    }
+
+    // Configure how many Goose Users to hatch.
+    fn set_users(&mut self) -> Result<(), GooseError> {
```
jeremyandrews, line 1255: @LionsAd In this case we are writing information in `GooseAttack` based on `configuration` and `defaults`. This could also just also overload `configuration`, as we track how many users _actually_ start in the metrics structure.

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+
+        if self.attack_mode != GooseMode::Manager && self.attack_mode != GooseMode::Worker {
+            debug!("users = {}", self.users);
+        }
+
+        Ok(())
+    }
+
+    // Configure maximum run time if specified, otherwise run until canceled.
+    fn set_run_time(&mut self) -> Result<(), GooseError> {
```
jeremyandrews, line 1302: @LionsAd This one also ends up in `GooseAttack`, but likely could just overload `configuration`.

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+
+        if self.run_time > 0 {
+            info!("run_time = {}", self.run_time);
+        }
+
+        Ok(())
+    }
+
+    // Configure how quickly to hatch Goose Users.
+    fn set_hatch_rate(&mut self) -> Result<(), GooseError> {
```
jeremyandrews, line 1335: @LionsAd This is going into `GooseAttack`, but likely could just overload `configuration`

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+
+        if self.hatch_rate > 0 {
+            info!("hatch_rate = {}", self.configuration.hatch_rate);
+        }
+
+        Ok(())
+    }
+
+    // Configure maximum requests per second if throttle enabled.
+    fn set_throttle_requests(&mut self) -> Result<(), GooseError> {
```
jeremyandrews, line 1375: @LionsAd Another that's going into `GooseAttack` but could just overload `configuration`.

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+
+        if self.throttle_requests > 0 {
+            info!("throttle_requests = {}", self.throttle_requests);
+        }
+
+        Ok(())
+    }
+
+    // Determine if no_reset_statics is enabled.
+    fn no_reset_metrics(&self) -> bool {
```
jeremyandrews, line 1434: @LionsAd should this be named `get_no_reset_metrics`? We're just determining if this flag is set (or defaulted to `true`)

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+            } else {
+                default
+            }
+        } else {
+            false
+        }
+    }
+
+    // Determine if the status_codes flag is enabled.
+    fn set_status_codes(&mut self) {
```
jeremyandrews, line 1450: @LionsAd In this case, we're overloading the `configuration` (as we need this information on Workers)

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+                    default
+                }
+            } else {
+                false
+            };
+        }
+    }
+
+    // Determine if the no_task_metrics flag is enabled.
+    fn set_no_task_metrics(&mut self) {
```
jeremyandrews, line 1484: @LionsAd Again overloads `configuration` as it's needed on the Worker.

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+                        default
+                    }
+                } else {
+                    false
+                };
+        }
+    }
+
+    // Determine if the no_metrics flag is enabled.
+    fn set_no_metrics(&mut self) {
```
jeremyandrews, line 1502: @LionsAd Again we overload `configuration`, in this case so it's available to `GooseUser`'s

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+                    default
+                }
+            } else {
+                false
+            };
+        }
+    }
+
+    // Determine if the sticky_follow flag is enabled.
+    fn set_sticky_follow(&mut self) {
```
jeremyandrews, line 1519: @LionsAd Again we overload `configuration`, again so it's available to `GooseUser`'s.

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+                }
+            } else {
+                false
+            };
+        }
+    }
+
+    #[cfg(feature = "gaggle")]
+    // Determine if no_hash_check is enabled.
+    fn no_hash_check(&self) -> bool {
```
jeremyandrews, line 1537: @LionsAd Should this be `get_no_hash_check` or `calculate_no_hash_check`? Or should we just overload `configuration` consistently?

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+        // If GooseDefault::MetricFile is set, return it.
+        if let Some(default_metrics_file) = &self.defaults.metrics_file {
+            return Ok(Some(default_metrics_file));
+        }
+
+        // Otherwise there is no metrics file.
+        Ok(None)
+    }
+
+    // Configure metrics log format.
```
jeremyandrews, line 1574: @LionsAd Here we're overloading `configuration` again, as we need this on Workers.

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
+                    options.join(", ")
+                ),
+            });
+        }
+
+        Ok(())
+    }
+
+    // If enabled, returns the path of the metrics_file, otherwise returns None.
+    fn get_debug_file_path(&self) -> Result<Option<&str>, GooseError> {
```
jeremyandrews, line 1617: @LionsAd Should we use `get_` or `calculate_`, or overload `configuration` for consistency?

src/lib.rs
```diff
@@ -885,51 +925,747 @@ impl GooseAttack {
...
-                if user_count >= self.users {
-                    trace!("created {} weighted_users", user_count);
-                    return Ok(weighted_users);
-                }
+        // Otherwise there is no debug file.
+        Ok(None)
+    }
+
+    // Configure debug log format.
+    fn set_debug_format(&mut self) -> Result<(), GooseError> {
```
jeremyandrews, line 1638: @LionsAd Here we overload `configuration` as we need it on the Workers

src/lib.rs
```diff
@@ -1851,7 +2438,205 @@ impl GooseAttack {
...
+/// fn main() -> Result<(), GooseError> {
+///     GooseAttack::initialize()?
+///         .set_default(GooseDefault::OnlySummary, true)
+///         .set_default(GooseDefault::Verbose, 1)
+///         .set_default(GooseDefault::MetricsFile, "goose-metrics.log");
+///
+///     Ok(())
+/// }
+/// ```
+pub trait GooseDefaultType<T> {
```
jeremyandrews, line 2515: @LionsAd Here's where we use a Generic `GooseDefaultType`, and then implement for `&str`, `usize` and `bool`. In order to have friendly error messages, we have to handle the entire `GooseDefault` enum in all three implementations.

tests/defaults.rs
```diff
@@ -0,0 +1,430 @@
...
+// - GooseDefault::Manager
+// - GooseDefault::ExpectWorkers
+// - GooseDefault::NoHashCheck
+// - GooseDefault::ManagerBindHost
+// - GooseDefault::ManagerBindPort
+// - GooseDefault::Worker
+// - GooseDefault::ManagerHost
+// - GooseDefault::ManagerPort
+
+// Can't be tested:
```
jeremyandrews, line 45: @LionsAd Note that we can't test LogFile, Verbose, or LogLevel in a meaningful way because of a limitation of the logging library we're using -- I was going to explore if there are other/better libraries out there that remove this limitation.

tests/defaults.rs
```diff
@@ -0,0 +1,430 @@
...
+// - GooseDefault::LogFile (logger can only be configured once)
+// - GooseDefault::Verbose (logger can only be configured once)
+// - GooseDefault::LogLevel (can't validate due to logger limitation)
+
+// Needs followup:
+// - GooseDefault::NoMetrics:
+//     Gaggles depend on metrics, when disabled load test does not shut down clearly.
```
jeremyandrews, line 52: @LionsAd We talked about this: shutting down a Gaggle cleanly depends on metrics being enabled. If metrics are disabled, it will shut down, but only by the Manager timing out after it tells the Workers to stop and they don't reply for ~60 seconds.

tests/defaults.rs
```diff
@@ -0,0 +1,430 @@
...
+// - GooseDefault::StickyFollow
+//     Needs more complex tests
```
jeremyandrews, line 54: @LionsAd I either need to add a second test to the existing StickyFollow test, that uses defaults as a configuration, or I need to re-implement that test using defaults in this test file.

tests/defaults.rs
```diff
@@ -0,0 +1,430 @@
...
+        .execute()
+        .unwrap();
+
+    validate_test(goose_metrics, index, about, &[metrics_file], &[debug_file]);
+}
+
+#[test]
+#[cfg_attr(not(feature = "gaggle"), ignore)]
+/// Load test confirming that Goose respects configured gaggle-related defaults.
+fn test_gaggle_defaults() {
```
jeremyandrews, line 182: @LionsAd We have a test file named `gaggle`, and we have this test file named `defaults`. Should this test live here in `defaults`, or in `gaggle`? (It's a combination of both features).

tests/defaults.rs
```diff
@@ -0,0 +1,430 @@
...
+        metrics_files.push(file);
+        let file = debug_file.to_string() + &i.to_string();
+        debug_files.push(file);
+    }
+    validate_test(goose_metrics, index, about, &metrics_files, &debug_files);
+}
+
+#[test]
+/// Load test confirming that Goose respects configured defaults.
+fn test_defaults_no_metrics() {
```
jeremyandrews, line 291: @LionsAd Originally I had added a `test_defaults_gaggle_no_metrics()` test which is how I realized things don't shut down cleanly in a Gaggle if metrics are disabled. If/when we fix this, where should this test live? Here in `defaults`, or in `gaggle`?

tests/one_taskset.rs
```diff
@@ -94,70 +94,6 @@ fn test_single_taskset() {
     assert!(goose_metrics.users == config.users.unwrap());
 }
 
-#[test]
-// Load test with a single task set containing two weighted tasks. Validate
-// that setting the host in the load test is properly recognized, and doesn't
-// otherwise affect the test.
-fn test_single_taskset_empty_config_host() {
```
jeremyandrews, line 101: @LionsAd This got moved into the new `defaults` test, as that's what it's testing.

tests/defaults.rs
jeremyandrews: I'm thinking maybe I should just remove the `gaggle.rs` test file, and spread gaggle tests out into all the other tests, as in many cases functionality should be tested both in standalone mode, as well as in gaggle mode.

jeremyandrews: I've created a followup issue for this: https://github.com/tag1consulting/goose/issues/158

JordiPolo:
I do not quite like this to be in the GooseAttack itself, it is already pretty big as it is.
Maybe unintended but on 0.9.0 it works pretty well to [setup a GooseConfiguration](https://github.com/JordiPolo/minos/blob/master/src/command_performance.rs#L85) struct. And then simply pass it to `GooseAttack::initialize_with_config`.

This kind of approach allows to put the code dealing with the config in a different place than the one initializing the attack. Also does not overload GooseAttack with too many responsibilities

src/lib.rs
jeremyandrews: I re-removed `hatch_rate` and `throttle_requests` and those are fully managed in `configuration`. Now only `attack_mode` is newly added to `GooseAttack`.

src/lib.rs
jeremyandrews: I re-removed this from `GooseAttack` and fully manage it with `configuration`.

src/lib.rs
jeremyandrews: I re-removed this from `GooseAttack` and manage it entirely in `configuration`.

jeremyandrews:
> I do not quite like this to be in the GooseAttack itself, it is already pretty big as it is.

All that's introduced is a `defaults` structure, which allows defaults to be set with proper validation.

> Maybe unintended but on 0.9.0 it works pretty well to setup a GooseConfiguration struct.
> And then simply pass it to GooseAttack::initialize_with_config.

Prior to `0.10` we were doing the same things in integration tests, but with this PR we're now properly configuring through `Gumdrop`. This has proved important as directly editing the `configuration` struct allowed unintentional regression as it can bypass things like bounds checking. 

> This kind of approach allows to put the code dealing with the config in a different place than the one
> initializing the attack. Also does not overload GooseAttack with too many responsibilities

The changes in this PR don't prevent you from doing things this way, however I strongly recommend you use `set_default()` if at all possible so Goose can validate your configuration.

src/lib.rs
jeremyandrews: Actually this can not use `configuration` as Users is optional. The final value lives in `GooseAttack` so we can detect if someone incorrectly tries to configure `--users` on a Worker.

src/lib.rs
jeremyandrews: Actually, this can not live in `configuration` as the Type changes from `&str` to `usize`. When setting `--run-time` it supports a string such as `1h30m` -- this is converted and stored as a `usize` in `GooseAttack`.

CHANGELOG.md
```diff
@@ -23,6 +23,33 @@
...
+    o default metrics log file name: `GooseDefault::MetricsFile` (&str)
+    o default metrics log file format: `GooseDefault::MetricsFormat` (&str)
+    o default debug log file name: `GooseDefault::DebugFile` (&str)
+    o default debug log file format: `GooseDefault::DebugFormat` (&str)
+    o default to tracking status codes: `GooseDefault::StatusCodes` (bool)
+    o default maximum requests per second: `GooseDefault::ThrottleRequests` (usize)
+    o default to following redirect of base_url: `GooseDefault::StickyFollow` (bool)
+    o default to enabling Manager mode: `GooseDefault::Manager` (bool)
+    o default number of Workers to expect: `GooseDefault::ExpectWorkers` (usize)
+    o default to ignoring load test checksum: `GooseDefault::NoHashCheck` (bool)
```
JordiPolo, line 47: I find it mildly confusing that there are a few "No..." together with the others.

CHANGELOG.md
```diff
@@ -23,6 +23,33 @@
...
+    o default host to bind Manager to: `GooseDefault::ManagerBindHost` (&str)
```
JordiPolo, line 48: I wish all these were their own thing, then maybe `GooseDefault::Manager` could be an `Option<ManagerDefaults>` so when None, Goose does not run in Manager mode.

src/lib.rs
```diff
@@ -468,6 +471,132 @@ impl From<io::Error> for GooseError {
...
+    StandAlone,
+    /// The controlling process in a Gaggle distributed load test.
+    Manager,
+    /// One of one or more working processes in a Gaggle distributed load test.
+    Worker,
+}
+
+/// Optional default values for Goose run-time options.
+#[derive(Clone, Debug, Default)]
+pub struct GooseDefaults {
```
JordiPolo, line 489: Why not call this `GooseConfig` and then allow to set this config as default for an attack or as configuration per task? Just reuse the same thing. Or maybe that's the idea and it is just the name confusing me.

src/lib.rs
```diff
@@ -512,11 +643,12 @@ impl GooseAttack {
             test_stop_task: None,
             task_sets: Vec::new(),
             weighted_users: Vec::new(),
-            host: None,
-            configuration,
+            defaults: GooseDefaults::default(),
+            configuration: GooseConfiguration::parse_args_default_or_exit(),
```
JordiPolo, line 647: Parsing the config already happens in this funcion, some lines above.

JordiPolo: What's the difference between the GooseAttack configuration from the arguments and the default? It is unclear to me which one should be used where

src/lib.rs
jeremyandrews: updated to `set_no_reset_metrics()` which updates `configuration` with default if set

src/lib.rs
jeremyandrews: updated to `set_no_hash_check()` which updates `configuration` with default if set

CHANGELOG.md
jeremyandrews: With configuration libraries, all flags default to `false` (which was the case both with `StructOpt` which we used before, and `Gumdrop` which we're using now). That said, you can `set_default()` to over-ride this default to be `true` if desired.

If running a Gaggle, then we default to performing a hash check as it's important all members of the Gaggle are running the same load test, meaning the flag has to be `--no-hash-check' as it defaults to `false`. If you wanted to disable the hash check by default in your load test, you'd need to `set_default(GooseDefault::NoHashCheck, true)`.

src/lib.rs
jeremyandrews: I think the fundamental confusion is you're trying to modify Goose state directly, instead of using the provided functions. This was necessary prior to `0.10`, but I recommend once you upgrade that you use `set_default()` to modify Goose's configuration state.

You should not be editing `goose_attack.configuration` nor `goose_attack.defaults` directly.

src/lib.rs
jeremyandrews: It's orders of precedence:

- if a cli flag/option is set, we always use this
- otherwise, if a custom default is defined, we use this
- otherwise, we use Goose's base defaults

The rules are slightly more complex for something like the host name:
- if the URL itself defines a host, we use that
- otherwise, if the --host flag is set, we use that
- otherwise, if a host is defined for the task set, we use that
- otherwise, if a default host is defined, we use that
- otherwise we fail on error, as a host is required

jeremyandrews: Also note that Gaggle's complicate things a little more.

Most flags/options an only be set on the Manager, so `set_default()` can define a default and the underlying logic only sets these defaults on the Manager, and not on the Worker (as the Worker must instead inherit these settings from the Manager).

That said, there are a few options that can only be set on the Worker, and again `set_default()` will only set these defaults on the Worker, not on the Manager.

CHANGELOG.md
jeremyandrews: One of the improvements made in `0.10` is a more logical grouping of options in the help text when you run `-h`. Does that address your concern? Or what would be solving by moving Gaggle options into their own struct?

```rust
Options available when launching a Goose load test.


Optional arguments:
  -h, --help                 Displays this help
  -V, --version              Prints version information
  -l, --list                 Lists all tasks and exits

  -H, --host HOST            Defines host to load test (ie http://10.21.32.33)
  -u, --users USERS          Sets concurrent users (default: number of CPUs)
  -r, --hatch-rate RATE      Sets per-second user hatch rate (default: 1)
  -t, --run-time TIME        Stops after (30s, 20m, 3h, 1h30m, etc)
  -g, --log-level            Sets log level (-g, -gg, etc)
  -L, --log-file NAME        Enables log file and sets name
  -v, --verbose              Sets debug level (-v, -vv, etc)

Metrics:
  --only-summary             Only prints final summary metrics
  --no-reset-metrics         Doesn't reset metrics after all users have started
  --no-metrics               Doesn't track metrics
  --no-task-metrics          Doesn't track task metrics
  -m, --metrics-file NAME    Sets metrics log file name
  --metrics-format FORMAT    Sets metrics log format (csv, json, raw)
  -d, --debug-file NAME      Sets debug log file name
  --debug-format FORMAT      Sets debug log format (json, raw)
  --status-codes             Tracks additional status code metrics

Advanced:
  --throttle-requests VALUE  Sets maximum requests per second
  --sticky-follow            Follows base_url redirect with subsequent requests

Gaggle:
  --manager                  Enables distributed load test Manager mode
  --expect-workers VALUE     Sets number of Workers to expect
  --no-hash-check            Tells Manager to ignore load test checksum
  --manager-bind-host HOST   Sets host Manager listens on (default: 0.0.0.0)
  --manager-bind-port PORT   Sets port Manager listens on (default: 5115)
  --worker                   Enables distributed load test Worker mode
  --manager-host HOST        Sets host Worker connects to (default: 127.0.0.1)
  --manager-port PORT        Sets port Worker connects to (default: 5115)
```

LionsAd (approved):
RTBM - let’s get this in

PR #159, opened 2020-09-10
Use info not println

JordiPolo:
I think we are doing logging on the section and also it is good so users can control if they want to see this or not.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,7 +59,7 @@
 //!   let _goose = user.get("/path/to/foo").await?;
 //!
 //!   Ok(())
-//! }   
+//! }
 //! ```
 //!
 //! In the above example, we're using the GooseUser helper method `get` to load a path
@@ -78,7 +78,7 @@
 //!     let _goose = user.goose_send(request_builder.timeout(time::Duration::from_secs(3)), None).await?;
 //!
 //!     Ok(())
-//! }   
+//! }
 //! ```
 //!
 //! We pass the `request_builder` object to `goose_send` which builds and executes it, also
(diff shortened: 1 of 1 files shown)


jeremyandrews:
If you do not want the statistics displayed, the intent is that you don't invoke `.print()`.

If you invoke GooseAttack as follows, it will print the statistics:
```rust
    GooseAttack::initialize()?
        .register_taskset(taskset!("LoadtestTasks")
            .register_task(task!(loadtest_index))
        )
        .execute()?
        .print();
```

But if instead you invoke GooseAttack as follows, it will simply return the statistics to you, and you can do with them as you please:
```rust
    let goose_statistics = GooseAttack::initialize()?
        .register_taskset(taskset!("LoadtestTasks")
            .register_task(task!(loadtest_index))
        )
        .execute()?;
```

JordiPolo:
Yes, but we may not be talking about the same thing. 
If you see the diff, there are a few `println!` in the code parsing the options so these will always appear in the output which could be surprising and it is not consistent with other similar areas which use info!

jeremyandrews:
The things that are using `info!()` are debug logs that require setting `-v` to see.

There are things that are not debug logs, but fundamental to the load test that do intentionally do not require enabling debug logs to see. In particular: 1) when all Goose Users are up and running and the load test is generating maximum load, and 2) if we reset statistics after all users stop (which we do by default) because this info is permanently deleted at this point.

Currently there's no way to hide #1, and you can only disable #2 by setting `--no-reset-stats`. If this is somehow problematic, perhaps we could consider a `--quiet` or `--silent` flag which prevents any/all output (essentially silencing the above two outputs)?

JordiPolo:
Yeah, unless I really want control and do not print anything unless I do it by myself. I think that's the most flexible for any use case.

PR #165, opened 2020-09-14
rework hatch_rate with Option<usize>

jeremyandrews:
Rework `hatch_rate` implementation so it's stored in an `Option<usize>` as it can be `None` on the Worker (where it then inherits the setting from the Manager). We use an `Option` instead of just a `usize` that defaults to `0` so we can detect even an invalid configuration of `--hatch-rate 0` on the Worker.

If not otherwise configured and not on a Worker, `hatch_rate` defaults to `Some(1)`. 

Fixes #163.

PR #167, opened 2020-09-14
remove `GooseAttack.users`

### LINKED ISSUE #166: Move `users` out of `GooseAttack` so there's only a copy in `configuration`
As `users` is wrapped in `Option<>` we were storing the final value in the `GooseAttack`. This is confusing, as in nearly all cases configuration lives in the `configuration` object.

Unrelated, in #165 I reworked `hatch_rate` which is now also stored in an `Option<>` and realized the same can be done with `users`. This issue is to remove the extra copy of `users` out of `GooseAttack`, simplifying the underlying data model.

jeremyandrews:
Fixes #166. The `GooseAttack.users` field was redundant and unnecessary.

(This also adds `defaults` validation for `users` that should also be added in #165 for `hatch_rate` -- I'll update that PR shortly.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.10.1-dev
  - rework `hatch_rate` to be stored in an `Option<usize>` as it can be `None` on a Worker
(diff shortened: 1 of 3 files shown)


LionsAd (commented):
RTBM - retro

PR #169, opened 2020-09-16
return error if setting invalid defaults, bounds checking of defaults

### LINKED ISSUE #168: perform bounds checking on defaults
With 0.10 Goose gained the ability to define custom defaults for all run-time options. However, it's currently impossible to define non-valid defaults. What bounds checking we're currently doing leads to panics.

This issue is to do more exhaustive bounds checking, and to return more helpful errors instead of panicking wherever it makes sense.

jeremyandrews:
Returns a Result<> from the `GooseDefaultType`, which requires wrapping it in a `Box<>`. This code is not in the fast-path, it's just one time configuration at the start of a load test.

Performs bounds checking whether values are set by flags, custom defaults, or Goose defaults.

Fixes #168


LionsAd (commented):
Approach looks good to me - I love that you can continue chaining via ? Operator

src/lib.rs
```diff
@@ -1244,45 +1249,59 @@ impl GooseAttack {
...
             // Otherwise use default.
             } else {
+                key = "set_default(GooseDefault::Users)";
+                value = default_users;
+
                 self.configuration.users = Some(default_users);
             }
         // If not and if not running on Worker, default to 1.
         } else if self.attack_mode != GooseMode::Worker {
+            value = self.number_of_cpus;
```
LionsAd, line 1274: I think this should not be needed.

src/lib.rs
```diff
@@ -1324,43 +1352,52 @@ impl GooseAttack {
...
                 self.configuration.hatch_rate = None;
             // Otherwise use default.
             } else {
+                key = "set_default(GooseDefault::HatchRate)";
+                value = default_hatch_rate;
                 self.configuration.hatch_rate = Some(default_hatch_rate);
             }
         // If not and if not running on Worker, default to 1.
         } else if self.attack_mode != GooseMode::Worker {
+            value = 1;
```
LionsAd, line 1376: I think this can never lead to an error.

LionsAd (approved):
Generally looks good - just two code paths are confusing that only value is set. Might want to add a comment for why that is the case.

src/lib.rs
jeremyandrews: I don't understand?

If `--users` is set, we use that.
Otherwise, if `set_default(Users)` is set, we use that.
Otherwise we default to the number of cpus. 

Are you saying we shouldn't default to the number of cpus? Or that the `self.number_of_cpus` variable isn't needed? (The latter is correct: this was originally added because I was unable to call the function "on the fly" from a StructOpt structure -- now that we're not doing that, it can indeed be removed. Feels like it's better handled in another PR though?)

jeremyandrews: Oh, the `value` shouldn't be needed. I agree it shouldn't be, but: 1) in Rust we have to set `value` to something in all code-paths, and 2) what if the library that tells us the number of cpus returns an invalid number (ie, 0)?

src/lib.rs
jeremyandrews: Agreed -- my thinking was more what happens if we make later code-changes and introduce an error.

src/lib.rs
jeremyandrews: I've added the missing `key`, and documented why we set up debug here even though failure isn't currently possible.

src/lib.rs
jeremyandrews: I've added the missing `key`, and documented why we set up debug here even though failure currently isn't possible.

jeremyandrews:
> Generally looks good - just two code paths are confusing that only value is set. Might want to add a comment for why that is the case.

Thanks, I've added comments and the missing keys.

PR #175, opened 2020-09-18
remove `gaggle.rs`, instead adding gaggle-versions of existing integration tests

### LINKED ISSUE #158: remove `gaggle.rs` test file, and include gaggle tests in other test files
In the beginning, there was just one Gaggle integration test, so it made sense for it to live in its own file. But as we improving our testing framework, it's clear we need more Gaggle tests duplicating non-Gaggle tests to confirm the functionality works in both modes. As individual integration tests can be skipped when certain features aren't compiled in, there's no particular advantage to keeping Gaggle tests in a single file, instead it seems advantageous for similar tests to live together.

jeremyandrews:
It's important that all functionality is tested in standalone and gaggle mode to be sure it works in both cases. Toward this goal, we remove `tests/gaggle.rs` and instead add gaggle variations of the existing integration tests.

Fixes #158


tests/closure.rs
```diff
@@ -0,0 +1,390 @@
...
+#[test]
+#[cfg_attr(not(feature = "gaggle"), ignore)]
+// Spawn a gaggle of 1 manager and 2 workers each simulating one user. Run a load test,
+// with a single task set containing two weighted tasks setup via closure. Validate
+// that weighting and metrics are correctly merged to the Manager.
+fn test_single_taskset_closure_gaggle() {
+    use std::sync::Arc;
+
+    // Configure endpoints to test.
+    let test_endpoints = vec![
```
LionsAd, line 183: Uh - no ;) -- we really should not be duplicating the code here, but instead call a common helper function.

Especially as we re-organize it here, this is perfect to add the helper function. (can first be in this file)

LionsAd (changes requested):
Before we do that, we should re-work the way we setup the tests. This duplicates way too much code.

LionsAd:
Especially as the goal is to test every test in normal and gaggle mode => Do something that makes it simple to reach that goal even for new tests -- duplicating all code ever written is not scalable way to reach that goal

tests/closure.rs
jeremyandrews: My main concern is currently all integration tests are excellent tutorials. The more we abstract them, the less useful they become for this purpose. In any case, I do understand we make life easier for ourselves as we abstract.

jeremyandrews: @LionsAd I've reworked all tests affected by this PR, moving duplicated code into helper functions.

LionsAd (approved):
RTBM - Much better now!

LionsAd:
Looks great

Issue #182, opened 2020-09-21
`test_start` and `test_stop` get run on each Worker instead of just one time


jeremyandrews:
The `test_start` functions are intended to be run one time when the load tests starts, and the `test_stop` function one time when the load test stops.

While this happens correctly in standalone mode, when running a load test as a Gaggle, they incorrectly end up running once per Worker. One possible solution would be to run `on_start` and `on_stop` tasks from the Manager, instead of coordinating them with the Workers.

The attached patch adds a test which illustrates the problem:

[test.txt](https://github.com/tag1consulting/goose/files/5253504/test.txt)

jeremyandrews:
Address the @FIXME in this test when fixing this bug:
https://github.com/tag1consulting/goose/blob/master/tests/redirect.rs#L281

PR #185, opened 2020-09-23
deduplicate setup_teardown test, add gaggle versions

jeremyandrews:
Continuing what was started in #175, this de-duplicates and adds Gaggle tests for `tests/setup_teardown.rs`.

There is a big @FIXME in the code as until #182 lands the Gaggle functionality isn't working correctly.

### DIFF
--- a/tests/setup_teardown.rs
+++ b/tests/setup_teardown.rs
@@ -1,13 +1,29 @@
 use httpmock::Method::{GET, POST};
-use httpmock::{Mock, MockServer};
+use httpmock::{Mock, MockRef, MockServer};
 
 mod common;
 
 use goose::prelude::*;
+use goose::GooseConfiguration;
 
 const INDEX_PATH: &str = "/";
 const SETUP_PATH: &str = "/setup";
 const TEARDOWN_PATH: &str = "/teardown";
+
+const INDEX_KEY: usize = 0;
+const SETUP_KEY: usize = 1;
+const TEARDOWN_KEY: usize = 2;
+
+// Defines the different types of tests.
+#[derive(Clone)]
+enum TestType {
+    // Testing on_start alone.
+    Start,
+    // Testing on_stop alone.
+    Stop,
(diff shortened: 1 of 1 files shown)


tests/setup_teardown.rs
```diff
@@ -26,130 +46,334 @@ pub async fn get_index(user: &GooseUser) -> GooseTaskResult {
...
+
+// Common validation for the load tests in this file.
+// @FIXME: remove the `is_gaggle` flag once issue #182 lands.
+fn validate_test(test_type: &TestType, mock_endpoints: &Vec<MockRef>, is_gaggle: bool) {
+    // Confirm the load test ran.
+    assert!(mock_endpoints[INDEX_KEY].times_called() > 0);
+
+    // Now confirm TestType-specific counters.
+    if is_gaggle {
+        // @FIXME: when https://github.com/tag1consulting/goose/issues/182 lands
```
jeremyandrews, line 115: The `is_gaggle` flag will need to be removed when #182 lands (and was setup so it will start failing when Gaggle's handling of `test_start` and `test_stop` are fixed so we don't forget).

tests/setup_teardown.rs
```diff
@@ -26,130 +46,334 @@ pub async fn get_index(user: &GooseUser) -> GooseTaskResult {
...
+                assert!(mock_endpoints[TEARDOWN_KEY].times_called() == 0);
+            }
+            TestType::StartAndStop => {
+                // @FIXME: setup and teardown should have run.
+                assert!(mock_endpoints[SETUP_KEY].times_called() == 0);
+                assert!(mock_endpoints[TEARDOWN_KEY].times_called() == 0);
+            }
+        }
+    } else {
+        match test_type {
```
jeremyandrews, line 136: Only the `else` block will remain when we can remove `is_gaggle`, as the test should have identical results whether running in standalone or gaggle mode.

LionsAd (approved):
RTBM

PR #191, opened 2020-09-29
add a test for task sequencing

jeremyandrews:
Add a test to confirm `set_sequence()` on tasks works as expected, affecting the order tasks are executed. This leverages that by default Goose executes tasks in the order they are declared, combined with a task that sleeps for the duration of the test after which no other tasks run. Without sequences all tasks run, as the delay task runs last. With sequences only two tasks run, as the delay task runs sooner.

Also confirms that sequences do not affect `test_start` and `test_stop` functions, which should always run regardless.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.10.3-dev
  - fixup sticky redirect tests to properly test functionality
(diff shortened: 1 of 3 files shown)


jeremyandrews:
This worked locally, changing parameters I can unexpectedly duplicate -- apparently needs work.

PR #200, opened 2020-10-23
get language-appropriate random words from titles

jeremyandrews:
- move logic to get a random language-appropriate word from titles (originally used for search) into a function
 - search for multiple random words instead of just 1 word
 - randomly generate name, email, subject and message for content form using the same logic
 - remove `info!()` when contact form is throttled as this happens too often for the information to be helpful when running a load test

### DIFF
--- a/examples/umami/common.rs
+++ b/examples/umami/common.rs
@@ -1,7 +1,6 @@
 use goose::goose::GooseResponse;
 use goose::prelude::*;
 
-use log::info;
 use rand::prelude::IteratorRandom;
 use rand::seq::SliceRandom;
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


examples/umami/common.rs
```diff
@@ -521,12 +557,15 @@ pub async fn anonymous_contact_form(user: &GooseUser, english: bool) -> GooseTas
                     }
 
                     // Build contact form parameters.
-                    // @TODO: dynamically generate the content.
+                    let name = random_words(2, english).join(" ");
+                    let email = format!("{}@example.com", random_words(1, english).pop().unwrap());
```
LionsAd, line 561: Could this be invalid and lead to an error?

LionsAd (approved):
Except for a new error of email being invalid this looks good ...

examples/umami/common.rs
jeremyandrews: I have not triggered any errors in my testing. It seems validation of the email address is done by this library:
https://github.com/egulias/EmailValidator/blob/master/src/Validation/RFCValidation.php

In any case, even if it _can_ generate an error, an infrequent error condition still seems like a valid test?

jeremyandrews: I let the load test run for an hour and saw no errors:
```
 POST anon /en/contact    |         9,200 |         0 (0%) |     2.56 |    0.00
 POST anon /en/search     |         9,198 |         0 (0%) |     2.56 |    0.00
 POST anon /es/contact    |         3,123 |         0 (0%) |     0.87 |    0.00
 POST anon /es/search     |         3,123 |         0 (0%) |     0.87 |    0.00
```

PR #209, opened 2020-10-30
add logged in admin user to umami load test

jeremyandrews:
Each admin load test user logs in one time at the start of the test and then runs the following tasks (and also load all static elements on any pages loaded):
 - Loads the front page;
 - Loads the article listing page;
 - Loads an "article", edits (not making any actual changes), and saves it (flushing all caches).

### DIFF
--- a/examples/umami/admin.rs
+++ b/examples/umami/admin.rs
@@ -1,0 +1,210 @@
+use goose::prelude::*;
+
+use crate::common;
+
+use rand::seq::SliceRandom;
+use std::env;
+
+/// Log into the website.
+pub async fn log_in(user: &GooseUser) -> GooseTaskResult {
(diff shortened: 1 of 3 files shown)


examples/umami/admin.rs
```diff
@@ -0,0 +1,210 @@
...
+                            Some(&html),
+                        );
+                    }
+
+                    // Build node form with random word from title.
+                    let params = [
+                        ("form_build_id", &form_build_id.unwrap()),
+                        ("form_token", &form_token.unwrap()),
+                        ("form_id", &"node_article_edit_form".to_string()),
+                        ("op", &"Save (this translation)".to_string()),
```
jeremyandrews, line 170: @LionsAd Is this sufficient to trigger a D9 cache flush? The logged-in user simply loads the edit page and submits the node edit form.

PR #215, opened 2020-11-04
support f32 hatch rate

jeremyandrews:
Currently the slowest rate to hatch users is 1 per second. With this PR, hatch rate can be specified as a f32, so for example you can say `-r.5` and it will launch 1 user every 2 seconds.


LionsAd (approved):
RTBM - Looks good to me

PR #225, opened 2020-11-06
expose GooseScheduler for allocating GooseTaskSets

### LINKED ISSUE #221: Make the order users are launched configurable
Currently users are launched in the order they’re defined. If a user has a weight of 10 the ten of those users are launched before the next type of user. 

Make .round_robin() an optional argument to GooseAttack so one or each type is created instead (respecting weights so we don’t change the total number if each type if user created)

Another option could be .rand()

jeremyandrews:
- allow configuration of the algorithm used when allocating `GooseTaskSet`s to starting `GooseUser`s:
   - `GooseScheduler::RoundRobin` allocates 1 of each available `GooseTaskSet` at a time (new default)
   - `GooseScheduler::Serial` allocates all of each available `GooseTaskSet` in the order they are defined
   - `GooseScheduler::Random` randomly allocates 1 `GooseTaskSet` at a time from all available
 - fixes #221

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1062,13 +1062,13 @@ impl GooseAttack {
                 // Allocate task sets round robin.
                 let task_sets_len = available_task_sets.len();
                 loop {
-                    for (i, task_sets) in available_task_sets
+                    for (task_set_index, task_sets) in available_task_sets
                         .iter_mut()
                         .enumerate()
                         .take(task_sets_len)
                     {
                         if let Some(task_set) = task_sets.pop() {
-                            debug!("allocating 1 user from TaskSet {}", i);
+                            debug!("allocating 1 user from TaskSet {}", task_set_index);
                             weighted_task_sets.push(task_set);
                         }
                     }
@@ -1079,11 +1079,11 @@ impl GooseAttack {
             }
             GooseTaskSetScheduler::Serial => {
                 // Allocate task sets serially in the weighted order defined.
-                for (task_set, task_sets) in available_task_sets.iter().enumerate() {
+                for (task_set_index, task_sets) in available_task_sets.iter().enumerate() {
                     debug!(
                         "allocating all {} users from TaskSet {}",
                         task_sets.len(),
-                        task_set
+                        task_set_index
                     );
                     weighted_task_sets.append(&mut task_sets.clone());
                 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -864,11 +864,11 @@ impl GooseAttack {
     /// Note that the following pattern is repeated:
     ///  A, B, A, B, A, B, A, A
     ///
-    /// If reconfigured to schedule serially, then in the above configuration they
-    /// will be allocated as follows:
+    /// If reconfigured to schedule serially, then they will instead be allocated in
+    /// the following order:
     /// A, A, A, A, A, B, B, B, A, A, A, A, A, B, B, B, A, A, A, A
     ///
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -956,8 +1032,9 @@ impl GooseAttack {
         // 'u' will always be the greatest common divisor
         debug!("gcd: {}", u);
 
-        // Build a weighted lists of task sets (identified by index)
-        let mut weighted_task_sets = Vec::new();
+        // Build a vector of vectors to be used to schedule users.
+        let mut available_task_sets = Vec::new();
```
LionsAd, line 1036: micro-optimization (probably not worth it): 

You could already know the total_tasks_set at this point so reserve capacity in the Vec.

src/lib.rs
```diff
@@ -968,8 +1045,66 @@ impl GooseAttack {
...
+                        }
+                    }
+                    if weighted_task_sets.len() >= total_task_sets {
+                        break;
+                    }
+                }
+            }
+            GooseScheduler::Serial => {
+                // Allocate task sets serially in the weighted order defined.
+                for (task_set, task_sets) in available_task_sets.iter().enumerate() {
```
LionsAd, line 1082: Should use `i` here for index like above or even better use: `task_set_index, task_sets`

src/lib.rs
```diff
@@ -837,6 +853,59 @@ impl GooseAttack {
...
+    /// are launched.
+    ///
+    /// By default, GooseTaskSets are allocated to new GooseUser's in a round robin
+    /// style. For example, if TaskA has a weight of 5, TaskB has a weight of
+    /// 4, and you launch 20 users, they will be launched in the following order:
+    ///  A, B, A, B, A, B, A, B, A, A, B, A, B, A, B, A, B, A, A, B
+    ///
+    /// If reconfigured to allocate serially, then in the above configuration they
+    /// will be allocated as follows:
+    /// A, A, A, A, A, B, B, B, B, A, A, A, A, A, B, B, B, B, A, B
```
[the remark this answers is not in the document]
jeremyandrews: This documentation changed in a later commit. I think it’s correct in the latest code. Can you recheck?

src/lib.rs
```diff
@@ -765,6 +780,7 @@ impl GooseAttack {
             configuration,
             run_time: 0,
             attack_mode: GooseMode::Undefined,
+            scheduler: GooseScheduler::RoundRobin,
```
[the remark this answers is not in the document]
jeremyandrews: Yes, I like that. And maybe as a followup we should add a GooseTaskScheduler with the same options (currently it uses the random method, but it would be nice for it also to be configurable). Fixed.

src/lib.rs
jeremyandrews: You're absolutely correct, and [the documentation makes it clear](https://doc.rust-lang.org/std/vec/struct.Vec.html#capacity-and-reallocation) it is recommended to specify capacity if you know it, otherwise the vector may get moved around in memory as it grows. This would be more important with larger load tests. The `available_task_sets` vector will always be the same length as our existing vector of `task_sets`, fixed.

src/lib.rs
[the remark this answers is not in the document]
jeremyandrews: I've triple-checked the documentation is correct (and made it more clear) now.

src/lib.rs
jeremyandrews: Yes, I use `task_set_index` elsewhere in the code, so using that consistently now here. Fixed.

PR #229, opened 2020-11-11
account for time spent not sleeping, looping every second

jeremyandrews:
- when setting `--running-metrics <usize>` the intent is that the running metrics are displayed every usize seconds
 - the Goose parent thread collects metrics from all GooseUser threads, and this takes some time (especially on bigger load tests)
 - it also displays running metrics when enabled
 - this PR subtracts the time spent doing these "other things" before running through the loop and doing it all over again, avoiding all but the most extreme time drifts
 - this also seems to fix some intermittent integration test failures that were inconsistently happening

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.10.7-dev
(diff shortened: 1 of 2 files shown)


LionsAd (approved):
RTBM - nice!

Issue #220, opened 2020-11-05
Long Lines in debug log are truncated 


jeremyandrews:
This makes reviewing the debug log difficult. At minimum we should add the new line at the start if the line instead if the end if the line. 

Tracking down and fixing how/why it’s getting truncated would also be helpful.

PR #236, opened 2020-11-13
increase debug log buffer size

jeremyandrews:
- increases the buffer size for the debug log from 8K to 8M
 - adds line feed at start of each debug file line instead of the end in case a web page is ever larger than the buffer
 - fixes https://github.com/tag1consulting/goose/issues/220

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -25,7 +25,8 @@ pub async fn logger_main(
         debug_file = match File::create(&file_path).await {
             Ok(f) => {
                 info!("writing errors to debug_file: {}", &file_path);
-                Some(BufWriter::new(f))
+                // Increase from default capacity of 8K to 8M so large pages can be captured.
+                Some(BufWriter::with_capacity(8 * 1024 * 1024, f))
             }
             Err(e) => {
                 panic!("failed to create debug_file ({}): {}", file_path, e);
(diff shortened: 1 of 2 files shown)


src/logger.rs
```diff
@@ -25,7 +25,8 @@ pub async fn logger_main(
         debug_file = match File::create(&file_path).await {
             Ok(f) => {
                 info!("writing errors to debug_file: {}", &file_path);
-                Some(BufWriter::new(f))
+                // Increase from default capacity of 8K to 8M so large pages can be captured.
+                Some(BufWriter::with_capacity(8 * 1024 * 1024, f))
```
jeremyandrews, line 29: @LionsAd This buffer gets allocated at start time only if the debug logger is enabled. It was using Tokio's default buffer size of 8K, I've increased it to 8M. I think this is acceptable overhead, but I'm flagging it in case you disagree.

Possibly it's worth exposing this as an environment variable or compile time option, so it's relatively easy to manually adjust the buffer size? Maybe if you're load testing a web site with extremely large web pages that would be important?

LionsAd: Can't you just write buffered several times to the log when the message arrives? Feels easier to me ...

jeremyandrews: I would have to split the message into 8K chunks, writing each chunk to the buffer -- this feels more complex to me? Plus, it makes the buffer less effective, as it has to be flushed each time we fill it.  A larger buffer makes the debug logger more efficient.

LionsAd: Okay -
Makes sense.

LionsAd (approved):
RTBM - then

Issue #230, opened 2020-11-12
optionally disable logging of bodies


jeremyandrews:
Add a flag to optionally disable the logging of response bodies in the debug_log, as when not needed this can be very large.

PR #237, opened 2020-11-14
optionally don't log response body in debug log

jeremyandrews:
- introduce `--no-debug-body` to optionally not log the response body in the debug log
 - allocate less memory to the debug log buffer when not logging response bodies (64K versus 8M)
 - fixes https://github.com/tag1consulting/goose/issues/230

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -22,11 +22,17 @@ pub async fn logger_main(
     // If debug file is configured, prepare an asynchronous buffered file writer.
     let mut debug_file = None;
     if let Some(file_path) = debug_file_path {
+        // Allocate a smaller buffer (64K) when not logging response bodies.
+        let buffer_capacity = if configuration.no_debug_body {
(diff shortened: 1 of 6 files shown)


src/logger.rs
```diff
@@ -22,11 +22,17 @@ pub async fn logger_main(
     // If debug file is configured, prepare an asynchronous buffered file writer.
     let mut debug_file = None;
     if let Some(file_path) = debug_file_path {
+        // Allocate a smaller buffer (64K) when not logging response bodies.
+        let buffer_capacity = if configuration.no_debug_body {
+            64 * 1024
```
jeremyandrews, line 27: The default size is 8K, but it I don't want to risk truncating debug logs artificially -- for example, if someone is dumping a large POST array, or there's lots of response headers.

LionsAd (approved):
RTBM

PR #238, opened 2020-11-14
rename metrics file to requests file

### LINKED ISSUE #226: the `--metrics-file` should be renamed to `---requests-file`
The metrics file optionally created when running a GooseAttack logs all requests made during a load test. While this data is used by Goose to generate metrics, it would more logically be referred to as a requests file as that's precisely what it's logging. This would avoid confusion with it and the optional metrics displayed while running a load test.

jeremyandrews:
- Goose can optionally generate a log of all requests it makes, which is confusingly referred to as the "metrics file"; this PR renames it more accurately to the "requests file"
 - rename `--metrics-file` to `--requests-file`
 - rename `--metrics-format` to `--requests-format`
 - fixes #226


LionsAd (approved):
RTBM - Thanks for only making the needed changes.

jeremyandrews:
> Thanks for only making the needed changes.

The beauty of search and replace! :)

PR #244, opened 2020-12-01
[WIP] optionally write a summary HTML report

### LINKED ISSUE #232: Create an HTML report

jeremyandrews:
- introduce `--report-file` option (and `GooseDefault::ReportFile`)
 - introduce a report template, copied from the Locust report template
 - add dependency on https://docs.rs/chrono to add human readable time to reports
 - write Request metrics to report
 - write Response time metrics to report
 - optionally write Task metrics to report if enabled (if `--no-task-metrics` not enabled)
 - optionally write Status Code metrics to report if enabled (if `--status-codes` is enabled)
 - fixes #232 

--

 *@TODO:*
 - add error summary to report
 - add charts to HTML report


src/lib.rs
```diff
@@ -2972,6 +3046,162 @@ impl GooseAttack {
...
+                aggregate_total_count,
+                aggregate_response_time_minimum,
+                aggregate_response_time_maximum,
+            ));
+
+            // Compile the request metrics template.
+            let mut request_template = Vec::new();
+            for metric in request_metrics {
+                request_template.push(report::request_metrics_template(metric));
+            }
```
jeremyandrews, line 3168: @LionsAd Here's the "inner template", the rows of a table within the template. I'm not sure if "compiling" is the correct term to describe what we're doing here: but this is compiling/formatting/search-and-replacing the template -- we build a vector of strings, one entry per row in the table.

src/lib.rs
```diff
@@ -2972,6 +3046,162 @@ impl GooseAttack {
...
+
+            // Compile the report template.
+            let report = report::build_report(
+                &start_time,
+                &end_time,
+                &host,
+                &request_template.join("\n"),
+                &response_template.join("\n"),
+            );
```
jeremyandrews, line 3183: @LionsAd And here's the "outer" template -- this builds the actual HTML report, and inserts the inner templates we built earlier.

src/report.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub percentile_70: String,
+    pub percentile_80: String,
+    pub percentile_90: String,
+    pub percentile_95: String,
+    pub percentile_99: String,
+    pub percentile_100: String,
+}
+
+/// Helper to generate a single response metric.
+pub fn get_response_metric(
```
jeremyandrews, line 40: I like this function, but I'm open to better ways to do this. I originally was doing all this work each time directly, and it was a lot of code duplication. It's perhaps a little odd how I build a Vector and then `take` the Strings out of the vector, but it saves having to do it more manually for each column.

src/report.rs
```diff
@@ -0,0 +1,256 @@
...
+        percentile_70: mem::take(&mut percentiles[2]),
+        percentile_80: mem::take(&mut percentiles[3]),
+        percentile_90: mem::take(&mut percentiles[4]),
+        percentile_95: mem::take(&mut percentiles[5]),
+        percentile_99: mem::take(&mut percentiles[6]),
+        percentile_100: mem::take(&mut percentiles[7]),
+    }
+}
+
+pub fn request_metrics_template(metric: RequestMetric) -> String {
```
jeremyandrews, line 75: @LionsAd This is how we build the "inner" templates without using something like Handlebars to loop. format!() requires that the template itself has to be hard coded, we can't use a variable (or even a constant).

src/report.rs
```diff
@@ -0,0 +1,256 @@
...
+        percentile_70 = metric.percentile_70,
+        percentile_80 = metric.percentile_80,
+        percentile_90 = metric.percentile_90,
+        percentile_95 = metric.percentile_95,
+        percentile_99 = metric.percentile_99,
+        percentile_100 = metric.percentile_100,
+    )
+}
+
+pub fn build_report(
```
jeremyandrews, line 127: @LionsAd And this is the actual file template. It required some extra escaping (ie `{{` to escape `{`) but otherwise is the same.

LionsAd (approved):
RTBM - Overall looks great to me now.

src/lib.rs
```diff
@@ -2972,6 +3006,66 @@ impl GooseAttack {
...
+    async fn write_html_report(&mut self) -> Result<(), GooseError> {
+        // Only write the report if enabled.
+        if let Some(report_file) = self.get_report_file_path()? {
+            // create the handlebars registry
+            let mut handlebars = Handlebars::new();
+
+            // Load report template into memory.
+            handlebars
+                .register_template_string("report", report::TEMPLATE)
+                .expect("failed to parase built-in template");
```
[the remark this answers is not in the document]
jeremyandrews: @LionsAd I'm not sure how this ended up on your review, this is from an earlier version of the code before I removed Handlebars...?

PR #256, opened 2021-03-20
respect hatch rate when launching gaggle

### LINKED ISSUE #253: Distributed load test not respecting hatch rate
When spinning up a Gaggle, Goose Workers are not respecting the hatch rate that's set on the Manager. 

If the Manager is configured with a hatch rate of 100 and there are 4 Workers, they should each launch Users at the rate of 25 per second. However instead each Worker seems to launch Users at around ~2/second.

jeremyandrews:
I originally assumed (incorrectly) that we weren't properly setting `hatch_rate` on the Workers. Instead what's happening is we're pushing metrics from the Workers to the Manager too frequently, and when running as a Gaggle we end up blocking waiting for the Manager to reply.

To solve:
- We make the Manager wake every 100ms instead of ever 500ms (so we minimize how often we block)
- We only synchronize the metrics at least as often as we display metrics, or at least once every five seconds
- We optimize serialization with `serde_cbor` by using a BufWriter

Fixes #253

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.10.9-dev
  - avoid unnecessary work on Manager when starting a Gaggle
+ - respect `--hatch-rate` when starting a Gaggle mode
 
 ## 0.10.8 Feb 13, 2021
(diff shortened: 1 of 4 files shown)


LionsAd (commented):
RTBM (retro)

PR #263, opened 2021-04-07
prepare for Rust 2021

### LINKED ISSUE #262: naming conventions style issue
An update to Clippy has triggered a new style error:
https://github.com/tag1consulting/goose/runs/2286275958

The specific Clippy documentation can be found here:
https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

And the Rust naming conventions it's enforcing:
https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case

jeremyandrews:
- fixes #262, following the Rust naming conventions in regards to case as defined in https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case
 - uses a custom display format to display methods so the name continues to be all caps as expected, for example `GET` instead of `Get`
 - removes a lingering `panic!()` that was incorrectly formatted for Rust 2021, replacing it with a correctly formatted error
 - optimize vector creation by avoiding creating a mutable vector then pushing data multiple times, instead using the more performant and easier to read `vec![]` macro (https://rust-lang.github.io/rust-clippy/master/index.html#vec_init_then_push)
 - call `format!` directly in closure test for improved readability

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.11.0-dev
  - capture errors and count frequency for each, including summary in metrics report; optionally disable with `--no-error-summary`
- - clippy cleanups:
+ - clippy cleanups (prepare for Rust 2021 https://blog.rust-lang.org/inside-rust/2021/03/04/planning-rust-2021.html):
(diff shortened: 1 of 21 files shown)


LionsAd (commented):
RTBM (retro)

PR #264, opened 2021-05-04
respect GooseScheduler when allocating GooseTasks

jeremyandrews:
Prior to this PR, Goose's Task scheduler is a bit complex: it maintains pointers to multiple buckets (tracking sequenced tasks versus unsequenced tasks), and initially allocates `GooseTask`s sequentially as they're defined, and subsequently shuffles them each time they're re-run.

With this PR, each `GooseUser` schedules their `GooseTask`s one time at startup, and then repeatedly runs the tasks over and over in the same order. It allocates `GooseTask`s based on the configured `GooseScheduler`, so either `RoundRobin`, `Sequential`, or `Random`, defaulting to `RoundRobin` if not otherwise configured. The primary goal of this change is to ensure that load tests are consistent from run to run; general simplification was just a bonus.

(When writing https://github.com/tag1consulting/loadtest-blog I initially had all tasks within a single taskset, and was annoyed that it was _first_ loading the blog 95 times, then loading the front page 4 times and so on: I had to [split it like this into three task sets](https://github.com/tag1consulting/loadtest-blog/blob/main/src/main.rs#L12) to properly round robin between the tasks. With this PR, I could have created a single task set and it still would have round robined from blog to front page to the blog listing page as desired.)

Changes in this PR:
 - allocate `GooseTask`s with the same `GooseScheduler` as is used to allocate `GooseTaskSet`s
 - simplify how `GooseUser`s run tasks, scheduling only once at start and then re-running the same progression each time; no-longer requires tracking multiple buckets just a single position (removing an atomic integer from the `GooseUser` object)
 - default to launching `GooseTask`s with the `GooseScheduler::RoundRobin` scheduler, instead of a variation on the `GooseScheduler::Serial` and `GooseScheduler::Random` scheduler (in which tasks were run in the order defined then shuffled each time they were run)
 - simplifying how tasks are scheduled provides a ~2% speedup in performance: https://docs.google.com/spreadsheets/d/1zRrNQDK2DBd7Dxi4hBvlOjiI-WYXnyEoMJp9baus8T0/edit#gid=0
 - add test coverage for scheduling `GooseTask`s, and for scheduling Sequenced `GooseTask`s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
  - update dependencies: `itertools` to `0.10`, `simplelog` to `0.10`, `url` to `2`
  - update `nng` dependency for optional `gaggle` feature
  - simplify `examples/umami` regex when parsing form
- - allow configuration of algorithm used when allocating `GooseTask`s, in addition to when allocating `GooseTaskSet`s; `GooseTaskSetScheduler` becomes more generically `GooseScheduler`
+ - allow configuration of algorithm for allocating `GooseTask`s the same as `GooseTaskSet`s; `GooseTaskSetScheduler` becomes more generically `GooseScheduler`
 
 ## 0.11.0 April 9, 2021
  - capture errors and count frequency for each, including summary in metrics report; optionally disable with `--no-error-summary`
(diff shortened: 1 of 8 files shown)


LionsAd (commented):
RTBM (retro) - nice solution

PR #267, opened 2021-05-08
update rand dependency to 0.8 branch

jeremyandrews:
- Update `rand` to the latest `0.8` branch
- Requires update to how the [`gen_range`](https://docs.rs/rand/0.8.*/rand/trait.Rng.html#method.gen_range) method is invoked


LionsAd (approved):
RTBM

PR #279, opened 2021-06-02
[WIP] add traits to improve compile-time validation

jeremyandrews:
- use traits to de-duplicate code where possible, and to gain compile-time validation that both Controllers are properly handling all defined commands
 - add `async_trait` dependency as Stable Rust doesn't otherwise support async traits
 - add `GooseController` trait:
     o add `get_command_string()` providing controller-specific logic for extracting a command from the protocol-specific request
 - add `GooseControllerExecuteCommand` trait:
     o add `execute_command()` to run the command received from a Controller request
     o add `write_to_socket()` to implement the per-protocol logic for sending a message back to the controller client
 - introduce `GooseControllerState` created in the main Controller thread and then passed to the per-client thread as new requests come in, also provides the following functions:
    o `accept_connections` starts an appropriate control loop for supported protocols
    o `get_match` determines what command the controller client sent
    o `process_command` creates a temporary one-shot channel, sends a request to the parent, and waits for a reply
    o `process_local_command` handles requests that don't involve the parent process
    o `process_response` prepares a client response based on the parent process response
 - removed `GooseCommand::Echo` which was just an initial proof of concept
 - removed `GooseControllerRequestMessage` enum instead using the old `GooseControllerCommandAndValue` struct renamed as `GooseControllerRequestMessage`, making the value optional
 - simplified `write_to_controller` to process a `Result<String, String>` directly, removed `error` from `GooseControllerWebSocketResponse` to avoid special-handling for the WebSocket Controller (the error text instead goes in the `response` field)

### DIFF
--- a/src/controller.rs
+++ b/src/controller.rs
@@ -117,7 +117,10 @@ pub struct GooseControllerWebSocketResponse {
 /// Return type to indicate whether or not to exit the Controller thread.
 type GooseControllerExit = bool;
 
-/// Simplify the GooseController trait definition for WebSockets.
+/// The telnet Controller message buffer.
+type GooseControllerTelnetMessage = [u8; 1024];
+
+/// The WebSocket Controller message buffer.
 type GooseControllerWebSocketMessage =
     std::result::Result<tungstenite::Message, tungstenite::Error>;
 
@@ -132,6 +135,8 @@ type GooseControllerWebSocketSender = futures::stream::SplitSink<
 pub struct GooseControllerState {
     /// Track which controller-thread this is.
     thread_id: u32,
+    /// Track the ip and port of the connected TCP client.
+    peer_address: String,
     /// A shared channel for communicating with the parent process.
     channel_tx: flume::Sender<GooseControllerRequest>,
     /// A compiled set of regular expressions used for matching commands.
@@ -144,17 +149,13 @@ pub struct GooseControllerState {
 // Defines functions shared by all Controllers.
 impl GooseControllerState {
     async fn accept_connections(self, mut socket: tokio::net::TcpStream) {
-        let peer_addr = socket
-            .peer_addr()
-            .map_or("UNKNOWN ADDRESS".to_string(), |p| p.to_string());
         info!(
             "{:?} client [{}] connected from {}",
-            self.protocol, self.thread_id, peer_addr
+            self.protocol, self.thread_id, self.peer_address
         );
-
         match self.protocol {
             GooseControllerProtocol::Telnet => {
-                let mut buf: [u8; 1024] = [0; 1024];
+                let mut buf: GooseControllerTelnetMessage = [0; 1024];
 
                 // Display initial goose> prompt.
                 write_to_socket_raw(&mut socket, "goose> ").await;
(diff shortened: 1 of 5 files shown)


src/controller.rs
```diff
@@ -117,15 +112,588 @@ pub struct GooseControllerWebSocketResponse {
...
+
+                // Display initial goose> prompt.
+                write_to_socket_raw(&mut socket, "goose> ").await;
+
+                loop {
+                    // Process data received from the client in a loop.
+                    let n = match socket.read(&mut buf).await {
+                        Ok(data) => data,
+                        Err(e) => {
+                            warn!("failed to read data from socket: {}", e);
```
jeremyandrews, line 167: The client disconnects here. Use same info message as elsewhere we break.

src/controller.rs
```diff
@@ -117,15 +112,588 @@ pub struct GooseControllerWebSocketResponse {
...
+        }
+
+        // Await response from parent.
+        match response_rx.await {
+            Ok(value) => Ok(value.response),
+            Err(e) => Err(format!("one-shot channel dropped without reply: {}", e)),
+        }
+    }
+
+    fn process_local_command(
```
jeremyandrews, line 361: Add missing comment.

src/controller.rs
```diff
@@ -117,15 +112,588 @@ pub struct GooseControllerWebSocketResponse {
...
+
+/// Controller-protocol-specific functions, necessary to manage the different way each
+/// Controller protocol communicates with a client.
+#[async_trait]
+trait GooseController<T> {
+    // Extract the command string from a Controller client request.
+    async fn get_command_string(&self, raw_value: T) -> Result<String, String>;
+}
+#[async_trait]
+impl GooseController<[u8; 1024]> for GooseControllerState {
```
jeremyandrews, line 454: Add type for this, similar to WebSocket socket type.

src/controller.rs
```diff
@@ -117,15 +112,588 @@ pub struct GooseControllerWebSocketResponse {
...
+                    let command_string: GooseControllerWebSocketRequest =
+                        match serde_json::from_str(&request) {
+                            Ok(c) => c,
+                            Err(_) => {
+                                return Err("unrecognized json request, refer to Goose README.md"
+                                    .to_string())
+                            }
+                        };
+                    return Ok(command_string.request);
+                }
```
jeremyandrews, line 495: Add else

src/controller.rs
```diff
@@ -117,15 +112,588 @@ pub struct GooseControllerWebSocketResponse {
...
+            }
```
jeremyandrews, line 496: Add else

src/controller.rs
```diff
@@ -117,15 +112,588 @@ pub struct GooseControllerWebSocketResponse {
...
+        }
```
jeremyandrews, line 497: Add else

src/controller.rs
```diff
@@ -117,15 +112,588 @@ pub struct GooseControllerWebSocketResponse {
...
+        socket: &mut tokio::net::TcpStream,
+        message: Result<String, String>,
+    ) {
+        // Send result to telnet Controller client, whether Ok() or Err().
+        let response_message = match message {
+            Ok(m) => m,
+            Err(e) => e,
+        };
+        if socket
+            // Add a linefeed to the end of the message.
```
jeremyandrews, line 571: and a prompt

PR #284, opened 2021-06-10
make internal shared funcs and structs pub(crate)

jeremyandrews:
While reviewing the latest [Goose docs](https://docs.rs/goose) I noticed it's a bit cluttered with functions and structures that aren't useful to consumers of the Goose library. In those cases I've changed their scope from `pub` to `pub(crate)` thereby removing them from the documentation.

 - documentation cleanup
 - de-scope internal-only documentation from docs.rs
 - improve public-facing documentation and provide more and better examples

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.11.3-dev
+ - remove internal-only functions and structures from documentation, exposing only what's useful to consumers of the Goose library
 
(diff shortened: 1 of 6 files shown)


jeremyandrews:
@LionsAd In theory everything I've re-scoped was useless to libraries using Goose, but technically this changes the exposed UI. Is it important I increment the version to `0.12` to reflect this?

PR #288, opened 2021-06-16
coordinated omission mitigation

### LINKED ISSUE #287: RFC: Conrolling/preventing Coordinated Ommission
This is to discuss what it might look like for a user to prevent CO in Goose.  
Strictly, user focused, or a write the code you wish you had approach.  Implementation flows from here.

1. `taskset` and `task`: These would seem to need an input defining the request-start interval.  The existing behavior can be preserved by renaming to `taskset_with_co` or have the default be "with-co" and create a new `taskset_without_co`, etc.  Question is could `taskset` define the interval for `task` when it is repeated.
1. `set_host`: These would seem to need an input defining the request-start interval - as a convenient way of setting the request interval for any `taskset` and `task` that don't have CO guards.  This comes down to: Does Goose want to support uses that are not guarded against CO?
1. `set_weight`: Setting a weight should not be incompatible with guarding against CO.  Should this blow up?: `task!(b_task_function, interval).set_weight(3)?`.  I don't think so.
1. `set_sequence` shouldn't need to change, as long a sequence means scheduled-sequence and not experienced-sequence.
1. `set_wait_time`: This probably is incompatible with trying to guard against CO, at least in its current usage, so this probably should raise an error about incompatible use: `taskset!("WebsiteUser", interval).set_wait_time(5,15)?`.  There are sane use cases for `set_wait_time` which means we probably also want to keep a CO vulnerable versions of `taskset` and `task`?

jeremyandrews:
- Closes #287
 - Introduces the `--co-mitigation` which supports the following options: "disabled" (earlier default, pre-0.12.0), "average" (current default), "minimum", "maximum", "median"
 - Coordinated omission mitigation compares the current request response-time to the configured baseline (ie, for `--co-mitigation average` it compares against the average response-time for the URL being requested, whereas for `--co-mitigation minimum` it compares against the minimum response-time for the URL being requested, etc)
 - Goose's coordinated omission mitigation is based on the general implementation from [HdrHistogram](https://github.com/HdrHistogram/HdrHistogram_rust/blob/master/src/lib.rs#L916)
 - As explained in the code, coordinate omission mitigation works as follows:

> Mitigate the loss of data (coordinated omission) due to stalls on the upstream server.
>
> Stalling can happen for lots of reasons, for example garbage collection, a cache stampede, even unrelated load affecting the server. Without any mitigation, Goose would lose statistically relevant information as it is unable to make additional requests while blocked by an upstream stall. It attempts to mitigate this by back-filling the requests that would have been made during that time.
>
> By default, Goose assumes that it should expect requests to return in the Average `response_time`. However, different server configurations and testing plans could work on different assumptions.

      

@TODO:
 - Add test(s) for coordinated omission mitigation
 - In a followup ticket explore adding similar support for GooseTaskMetricAggregate (which will likely require adding a `--tasks-file` similar to the `--requests-file` so we can analyze and confirm the change works as intended)

Here's a  graph of the first 250,000 entries in the requests log:
image

And here's the percentile graph showing more of the curve we're looking for:
image

To test, I ran Goose with the following parameters:
```
   cargo run --release --example drupal_loadtest -- -H http://apache -v -u12 --requests-file requests.log --requests-format json -t60
```

After the test was running 10-15 seconds, I used `ctrl-z` to pause Apache upstream for ~15 seconds, and then I resumed it. Once Goose finished I extracted just the response times as follows:
```
   cat requests.log | jq '.response_time' > response_time.log
```


jeremyandrews:
If instead using the average response time for a given request, the graph looks like:
image

jeremyandrews:
If instead the average response time is used, the graph changes to look as follows with the same basic test:
image

jeremyandrews:
If instead the average response time is used, the graph changes to look as follows with the same basic test:
image

jeremyandrews:
And with a more complex load test (ie, more simulated users and delays, so each user tends to hit the upstream hang at a different time):
image

bbros-dev:
Fascinating and a relief the workaround PoC hasn't failed miserably :)
I'm still not sure what a fix would look like, but back-filling observed data is a reasonable interim workaround.

We should also note this approach makes a very optimistic assumption - that all the missing/back-filled requests would have completed at the same time the blocking/CO response completed.

I'll refer to the empirical cumulative distribution functions (eCDF) images as #1, #2, #3, reading from top to bottom. 

As a side note, it'd be correct to plot these as step functions, making some additional assumptions (Cantelli theorem from memory) you could claim they should 'become' curvilinear as your sample size increases indefinitely.  I'm not convinced these systems satisfy those requirements, but that's just a hunch - and one I don't have time to look into.

IIUC, these plots relate to the 'same' experiment shown [here](https://user-images.githubusercontent.com/402892/121796746-b7840d00-cc1b-11eb-89bd-ecf24c710bb8.png). That's my assumption.

eCDF #2 looks to me like a classic case of CO, not sure if you agree?  I think that rules out the average, median or any other measure of central tendency - it obscures too much meaningful data (bad), as well as noise (good).

Does the case shown in eCDF #3 use the average or minimum response time to back fill data?  If it uses the average, it'd be interesting to see the back-filled data using the minimum. *You should be able to adjust this data without re-coding and rerunning the tests if that makes it easier?*

The eCDF #1 is roughly what we would expect if Goose generated data was immune to CO. 
The reasoning for the minimum went something like this, and now having looked at the Goose code a little more, I'm pretty sure this is correct - but happy to be corrected:

Let x period be the minimum/best response time.
Let the tested system be perfect and every response time be the best response time, under all conditions. Then (legacy) Goose (without developer injected pauses between requests) would issue a request every x period.  The response times would be a degenerate random variable and the histogram would be a point mass at x, the eCDF would likewise be a point.

This 'perfection' scenario is identical to configuring a CO immune system to make requests at a fixed rate of y RPS, such that the interval between requests is x period.  The data from this CO immune system would be identical to the data from Goose produced in the perfection scenario.

Now subject the 'perfection' tested system to the ctl-z pause. The CO immune system will produce the same data as Goose, if Goose uses a CO-correction interval of x period of time. 

Okay that also applies to some other metric, say 2x. Except you start to run the risk of missing CO-events with lifetime (ctl-z time) of 0.5x.
Furthermore, if you think its reasonable to discover how much your latency distribution deviates from the best case scenario (which generates a point mass), then you should, in a CO immune system, schedule requests at the rate of y RPS or, equivalently, an interval x period apart.
The difficulty in rationalizing the use use of another metric, say median, is being able to state the distribution you are comparing your results to - you know it won't be a point mass, because when the median exists, half the observations are below and half above. But are they uniformly distributed on both sides? Exponential on one side and normal on the other? Something else?  You'd have to make a CO-immune test of your system first. Which isn't our scenario.

Final remark.  Looking into this a little more, the workaround can be even less invasive, and can be something that won't necessarily need to be torn out when the eventual fix lands.  Just make the adjustments after the test run is finished.  This has some added benefit/advantages:
1. You preserve the raw data
2. You allow the user to choose what interval to use for the back filling (CO-correction), mean, median, average, user defined
3. You could also allow the user to adjust the CO-event type or response time assumption.

This last point comes from the fact the HDRHistogram correction assumes the CO-event is of the type where all scheduled requests return at the same point in time (or very near the same point in time) - this is true of the ctl-z type of experiments -  same code path = same response timestamp (roughly).
A different category of CO-events are those where the CO-event results in all scheduled requests flowing through the same code path **and** experience the same total delay. Meaning they would produce data with the same interval between request and response timestamps.
While there are many CO-event categories, this same code path = same total delay, is, I think, reasonable.  
It is also trivial to code - in fact users could even adjust the data themselves.  But is nice to have the convenience of a tool that could automatically produce these adjusted data.

I say these could survive a fix landing only because it is not yet clear to me if it will be desirable to preserve CO-vulnerable behavior.  Instinctively I don't think so, but I also know I haven't thought about this too much.

HTH?

src/metrics.rs
```diff
@@ -47,6 +49,68 @@ pub enum GooseMetric {
...
+/// stampede, even unrelated load affecting the server. Without any mitigation, Goose
+/// would lose statistically relevant information as it is unable to make additional
+/// requests will blocked by an upstream stall. It attempts to mitigate this by
+/// backfilling the requests that would have been made during that time.
+///
+/// By default, Goose assumes that it should expect requests to return in the Median
+/// response_time. However, different server configurations and testing plans could
+/// work on different assumptions.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum GooseCoordinatedOmissionMitigation {
```
bbros-dev, line 64: Great idea.

jeremyandrews:
As you've noted, the coordinate omission mitigation strategy is now configurable in this PR. It currently supports a few options: Disabled (pre-0.12 default), Average (current default), Minimum, Median, and Maximum (I'm not sure maximum has much value, but it was trivial to add as we already track the stats).

I continue to lean toward Average as the default, as it makes the most logical sense for what we're trying to achieve: we want to back-fill "missing" requests with "what should have happened", and logically that would be closest to whatever is average for a given request.

The different options need more analysis though.

** Disabled
image

```
     Running `target/release/examples/drupal_loadtest -H 'http://apache' -v -u12 --requests-file requests.log --requests-format json -t60 --co-mitigation dis`
10:53:33 [INFO] Output verbosity level: INFO
10:53:33 [INFO] Logfile verbosity level: WARN
10:53:33 [INFO] users = 12
10:53:33 [INFO] run_time = 60
10:53:33 [INFO] hatch_rate = 1
10:53:33 [INFO] co_mitigation = Some(Disabled)

 === PER REQUEST METRICS ===
 Name                     |        # reqs |        # fails |    req/s |  fail/s
 GET (Anon) front page    |        12,040 |         0 (0%) |   200.67 |    0.00
 GET (Anon) node page     |         8,036 |         0 (0%) |   133.93 |    0.00
 GET (Anon) user page     |         2,410 |         0 (0%) |    40.17 |    0.00
 GET (Auth) comment form  |           272 |         0 (0%) |     4.53 |    0.00
 GET (Auth) front page    |         1,352 |         0 (0%) |    22.53 |    0.00
 GET (Auth) node page     |           902 |         0 (0%) |    15.03 |    0.00
 GET (Auth) user page     |           271 |         0 (0%) |     4.52 |    0.00
 GET static asset         |        80,358 |         0 (0%) |     1339 |    0.00
 POST (Auth) comment form |           272 |         0 (0%) |     4.53 |    0.00
 -------------------------+---------------+----------------+----------+--------
 Aggregated               |       105,913 |         0 (0%) |     1765 |    0.00
 Name                     |    Avg (ms) |        Min |        Max |      Median
 GET (Anon) front page    |        0.09 |          1 |          12 |          1
 GET (Anon) node page     |       57.41 |          1 |      15,257 |         49
 GET (Anon) user page     |       12.31 |          1 |          73 |         14
 GET (Auth) comment form  |       57.67 |         31 |         116 |         54
 GET (Auth) front page    |       40.84 |         19 |      15,180 |         29
 GET (Auth) node page     |       90.87 |         29 |      15,209 |         54
 GET (Auth) user page     |       14.68 |          9 |          33 |         14
 GET static asset         |        0.07 |          0 |          15 |          0
 POST (Auth) comment form |       56.43 |         30 |         138 |         55
 -------------------------+-------------+------------+-------------+-----------
 Aggregated               |        6.33 |          1 |      15,257 |          1
 Slowest page load within specified percentile of requests (in ms):
 Name                     |    50% |    75% |    98% |    99% |  99.9% | 99.99%
 GET (Anon) front page    |      1 |      1 |      1 |      1 |      4 |      4
 GET (Anon) node page     |     49 |     69 |     96 |    100 | 15,000 | 15,000
 GET (Anon) user page     |     14 |     17 |     25 |     28 |     59 |     59
 GET (Auth) comment form  |     54 |     68 |     94 |     97 |    116 |    116
 GET (Auth) front page    |     29 |     32 |     44 |     50 |     79 |     79
 GET (Auth) node page     |     54 |     68 |     92 |     97 | 15,000 | 15,000
 GET (Auth) user page     |     14 |     16 |     24 |     24 |     33 |     33
 GET static asset         |      0 |      0 |      1 |      1 |      4 |      4
 POST (Auth) comment form |     55 |     65 |     86 |     88 |    138 |    138
 -------------------------+--------+--------+--------+--------+--------+-------
 Aggregated               |      1 |      1 |     71 |     80 |    100 | 15,000
```

** Average (current default)
image

```
     Running `target/release/examples/drupal_loadtest -H 'http://apache' -v -u12 --requests-file requests.log --requests-format json -t60 --co-mitigation ave`
10:46:15 [INFO] Output verbosity level: INFO
10:46:15 [INFO] Logfile verbosity level: WARN
10:46:15 [INFO] users = 12
10:46:15 [INFO] run_time = 60
10:46:15 [INFO] hatch_rate = 1
10:46:15 [INFO] co_mitigation = Some(Average)

 === PER REQUEST METRICS ===
 Name                     |        # reqs |        # fails |    req/s |  fail/s
 GET (Anon) front page    |        12,453 |         0 (0%) |   207.55 |    0.00
 GET (Anon) node page     |         8,614 |         0 (0%) |   143.57 |    0.00
 GET (Anon) user page     |         2,858 |         0 (0%) |    47.63 |    0.00
 GET (Auth) comment form  |           270 |         0 (0%) |     4.50 |    0.00
 GET (Auth) front page    |         1,561 |         0 (0%) |    26.02 |    0.00
 GET (Auth) node page     |           897 |         0 (0%) |    14.95 |    0.00
 GET (Auth) user page     |           342 |         0 (0%) |     5.70 |    0.00
 GET static asset         |        82,728 |         0 (0%) |     1379 |    0.00
 POST (Auth) comment form |           329 |         0 (0%) |     5.48 |    0.00
 -------------------------+---------------+----------------+----------+--------
 Aggregated               |       110,052 |         0 (0%) |     1834 |    0.00
 Name                     |    Avg (ms) |        Min |        Max |      Median
 GET (Anon) front page    |        0.12 |          0 |          12 |          0
 GET (Anon) node page     |      358.17 |         40 |      14,941 |         49
 GET (Anon) user page     |        1005 |         13 |      14,872 |         15
 GET (Auth) comment form  |       56.95 |         31 |         104 |         54
 GET (Auth) front page    |        1075 |         20 |      14,890 |         30
 GET (Auth) node page     |       58.52 |         27 |         111 |         55
 GET (Auth) user page     |        1635 |          9 |      14,878 |         15
 GET static asset         |        0.10 |          0 |          12 |          0
 POST (Auth) comment form |        1444 |         30 |      14,920 |         60
 -------------------------+-------------+------------+-------------+-----------
 Aggregated               |       79.49 |          9 |      14,941 |          9
 Slowest page load within specified percentile of requests (in ms):
 Name                     |    50% |    75% |    98% |    99% |  99.9% | 99.99%
 GET (Anon) front page    |      0 |      0 |      1 |      2 |      7 |      7
 GET (Anon) node page     |     49 |     71 |  8,000 | 12,000 | 14,941 | 14,941
 GET (Anon) user page     |     15 |     19 | 13,000 | 14,000 | 14,872 | 14,872
 GET (Auth) comment form  |     54 |     66 |     90 |     92 |    100 |    100
 GET (Auth) front page    |     30 |     35 | 13,000 | 14,000 | 14,890 | 14,890
 GET (Auth) node page     |     55 |     70 |     93 |    100 |    110 |    110
 GET (Auth) user page     |     15 |     21 | 13,000 | 14,000 | 14,878 | 14,878
 GET static asset         |      0 |      0 |      1 |      2 |      6 |      6
 POST (Auth) comment form |     60 |     74 | 13,000 | 14,000 | 14,920 | 14,920
 -------------------------+--------+--------+--------+--------+--------+-------
 Aggregated               |      9 |      9 |     79 |    120 | 13,000 | 14,941
```

** Minimum
image

```
     Running `target/release/examples/drupal_loadtest -H 'http://apache' -v -u12 --requests-file requests.log --requests-format json -t60 --co-mitigation min`
08:59:19 [INFO] Output verbosity level: INFO
08:59:19 [INFO] Logfile verbosity level: WARN
08:59:19 [INFO] users = 12
08:59:19 [INFO] run_time = 60
08:59:19 [INFO] hatch_rate = 1
08:59:19 [INFO] co_mitigation = Some(Minimum)

 === PER REQUEST METRICS ===
 Name                     |        # reqs |        # fails |    req/s |  fail/s
 GET (Anon) front page    |        11,757 |         0 (0%) |   195.95 |    0.00
 GET (Anon) node page     |        21,158 |         0 (0%) |   352.63 |    0.00
 GET (Anon) user page     |         5,089 |         0 (0%) |    84.82 |    0.00
 GET (Auth) comment form  |           338 |         0 (0%) |     5.63 |    0.00
 GET (Auth) front page    |         1,391 |         0 (0%) |    23.18 |    0.00
 GET (Auth) node page     |         2,206 |         0 (0%) |    36.77 |    0.00
 GET (Auth) user page     |           295 |         0 (0%) |     4.92 |    0.00
 GET static asset         |        78,652 |         0 (0%) |     1311 |    0.00
 POST (Auth) comment form |           934 |         0 (0%) |    15.57 |    0.00
 -------------------------+---------------+----------------+----------+--------
 Aggregated               |       121,820 |         0 (0%) |     2030 |    0.00
 Name                     |    Avg (ms) |        Min |        Max |      Median
 GET (Anon) front page    |        0.10 |          0 |           9 |          0
 GET (Anon) node page     |      975.93 |         55 |      15,777 |         55
 GET (Anon) user page     |        2028 |         18 |      15,699 |         18
 GET (Auth) comment form  |       55.04 |         33 |         124 |         51
 GET (Auth) front page    |       29.33 |         20 |          76 |         28
 GET (Auth) node page     |        3249 |         27 |      15,755 |         69
 GET (Auth) user page     |       14.47 |          9 |          34 |         13
 GET static asset         |        0.08 |          0 |          12 |          0
 POST (Auth) comment form |        4605 |         28 |      15,743 |      2,000
 -------------------------+-------------+------------+-------------+-----------
 Aggregated               |      348.94 |          9 |      15,777 |          9
 Slowest page load within specified percentile of requests (in ms):
 Name                     |    50% |    75% |    98% |    99% |  99.9% | 99.99%
 GET (Anon) front page    |      0 |      0 |      1 |      2 |      5 |      5
 GET (Anon) node page     |     55 |     69 | 13,000 | 14,000 | 15,777 | 15,777
 GET (Anon) user page     |     18 |    440 | 14,000 | 15,000 | 15,699 | 15,699
 GET (Auth) comment form  |     51 |     67 |     86 |     88 |    120 |    120
 GET (Auth) front page    |     28 |     32 |     44 |     48 |     72 |     72
 GET (Auth) node page     |     69 |  6,000 | 15,000 | 15,000 | 15,755 | 15,755
 GET (Auth) user page     |     13 |     16 |     24 |     25 |     34 |     34
 GET static asset         |      0 |      0 |      1 |      2 |      4 |      4
 POST (Auth) comment form |  2,000 |  9,000 | 15,000 | 15,000 | 15,743 | 15,743
 -------------------------+--------+--------+--------+--------+--------+-------
 Aggregated               |      9 |      9 |  8,000 | 12,000 | 15,000 | 15,777
```

** Median
image

```
     Running `target/release/examples/drupal_loadtest -H 'http://apache' -v -u12 --requests-file requests.log --requests-format json -t60 --co-mitigation med`
10:36:46 [INFO] Output verbosity level: INFO
10:36:46 [INFO] Logfile verbosity level: WARN
10:36:46 [INFO] users = 12
10:36:46 [INFO] run_time = 60
10:36:46 [INFO] hatch_rate = 1
10:36:46 [INFO] co_mitigation = Some(Median)

 === PER REQUEST METRICS ===
 Name                     |        # reqs |        # fails |    req/s |  fail/s
 GET (Anon) front page    |        11,757 |         0 (0%) |   195.95 |    0.00
 GET (Anon) node page     |         7,845 |         0 (0%) |   130.75 |    0.00
 GET (Anon) user page     |         2,352 |         0 (0%) |    39.20 |    0.00
 GET (Auth) comment form  |           269 |         0 (0%) |     4.48 |    0.00
 GET (Auth) front page    |         1,322 |         0 (0%) |    22.03 |    0.00
 GET (Auth) node page     |           884 |         0 (0%) |    14.73 |    0.00
 GET (Auth) user page     |           269 |         0 (0%) |     4.48 |    0.00
 GET static asset         |        78,469 |         0 (0%) |     1308 |    0.00
 POST (Auth) comment form |           269 |         0 (0%) |     4.48 |    0.00
 -------------------------+---------------+----------------+----------+--------
 Aggregated               |       103,436 |         0 (0%) |     1724 |    0.00
 Name                     |    Avg (ms) |        Min |        Max |      Median
 GET (Anon) front page    |        0.08 |          0 |           8 |          0
 GET (Anon) node page     |       59.23 |         52 |      15,495 |         52
 GET (Anon) user page     |       12.24 |         16 |          44 |         16
 GET (Auth) comment form  |       57.90 |         33 |         119 |         55
 GET (Auth) front page    |       53.48 |         21 |      15,435 |         29
 GET (Auth) node page     |       57.96 |         27 |         109 |         55
 GET (Auth) user page     |       14.86 |          9 |          29 |         14
 GET static asset         |        0.07 |          0 |          11 |          0
 POST (Auth) comment form |      115.32 |         34 |      15,440 |         56
 -------------------------+-------------+------------+-------------+-----------
 Aggregated               |        6.50 |          9 |      15,495 |          9
 Slowest page load within specified percentile of requests (in ms):
 Name                     |    50% |    75% |    98% |    99% |  99.9% | 99.99%
 GET (Anon) front page    |      0 |      0 |      1 |      1 |      3 |      3
 GET (Anon) node page     |     52 |     70 |     96 |    100 | 15,000 | 15,000
 GET (Anon) user page     |     16 |     17 |     27 |     28 |     34 |     34
 GET (Auth) comment form  |     55 |     67 |     94 |     95 |    119 |    119
 GET (Auth) front page    |     29 |     33 |     46 |     52 | 15,000 | 15,000
 GET (Auth) node page     |     55 |     69 |     92 |     95 |    109 |    109
 GET (Auth) user page     |     14 |     16 |     23 |     23 |     29 |     29
 GET static asset         |      0 |      0 |      1 |      1 |      3 |      3
 POST (Auth) comment form |     56 |     65 |     91 |     97 | 15,000 | 15,000
 -------------------------+--------+--------+--------+--------+--------+-------
 Aggregated               |      9 |      9 |     72 |     80 |    100 | 15,000
```

jeremyandrews:
> Let the tested system be perfect and every response time be the best response time, under all conditions.

Typically we use Goose in imperfect conditions (ie, websites on the internet), and where some fluctuation is expected. For my typical testing I would primarily want to protect against significant variations from the average. That said, I intend to leave this as configurable: it really just comes down to 1) what is the best default, and 2) are there any other variations that need to be supported (ie, average of 99%, stripping out the outliers which incorrectly inflate the average).

> Looking into this a little more, the workaround can be even less invasive, and can be something that won't necessarily need to be torn out when the eventual fix lands. Just make the adjustments after the test run is finished.

At this time the raw data is not written to disk (although enabled `--requests-file` is close as this writes a log of ALL query and response metrics). The raw data can be trivially written to disk when you create a loadtest, instead of calling `.print()` simply write out the raw data: but there's no process for reading this back in. Feel free to open a ticket and or to submit a PR.

Also: if we post-process the data, then it can't be included in the `request-file` as this is generated in real time.

bbros-dev:
> I intend to leave this as configurable: it really just comes down to 1) what is the best default, 

Yes and this is use case specific.  Over time we can document the effect of different settings, and provide lessons learned/guidance.

Overtime some rules of thumb around best practice will emerge.

jeremyandrews:
Analyzing the above numbers, this is a step forward but isn’t mitigating coordinated omissions sufficiently. I believe the logic has to be moved into goose_send and we need to track the cadence between all requests. Currently we solve for the specific request that got blocked, but not any other request in the task/taskset that never even runs during the time we’re blocked.

jeremyandrews:
I've updated the Coordinated Omission Mitigation to look at the cadence between requests so that all requests get back-filled, so not just those blocked directly by the upstream server, but also those blocked indirectly as no requests are even made. (And realized this logic can be greatly simplified: I'm currently tracking the cadence of each individual request, but really only need 1 cadence per `GooseUser`).

Here's the response times table (only some of these requests were directly blocked by the server, but all reflect the server slowdown) (Goose was launched using `cargo run --release --example drupal_loadtest -- -H http://apache -v -u10 --requests-file requests.log --requests-format json -t1m` and the Apache server was paused for ~12 seconds while the test was running.)

```
 === PER REQUEST METRICS ===
 Name                     |        # reqs |        # fails |    req/s |  fail/s
 GET (Anon) front page    |        14,552 |         0 (0%) |   242.53 |    0.00
 GET (Anon) node page     |         9,707 |         0 (0%) |   161.78 |    0.00
 GET (Anon) user page     |         2,914 |         0 (0%) |    48.57 |    0.00
 GET (Auth) comment form  |           257 |         0 (0%) |     4.28 |    0.00
 GET (Auth) front page    |         1,287 |         0 (0%) |    21.45 |    0.00
 GET (Auth) node page     |           857 |         0 (0%) |    14.28 |    0.00
 GET (Auth) user page     |           257 |         0 (0%) |     4.28 |    0.00
 GET static asset         |        95,029 |         0 (0%) |     1584 |    0.00
 POST (Auth) comment form |           258 |         0 (0%) |     4.30 |    0.00
 -------------------------+---------------+----------------+----------+--------
 Aggregated               |       125,118 |         0 (0%) |     2085 |    0.00
 Name                     |    Avg (ms) |        Min |        Max |      Median
 GET (Anon) front page    |        1038 |          0 |      11,660 |          0
 GET (Anon) node page     |        1076 |         63 |      11,660 |         63
 GET (Anon) user page     |        1057 |         14 |      11,645 |         14
 GET (Auth) comment form  |        1452 |         32 |      11,721 |         61
 GET (Auth) front page    |        1399 |         20 |      11,760 |         28
 GET (Auth) node page     |        1412 |         29 |      11,768 |         59
 GET (Auth) user page     |        1368 |          9 |      11,651 |         14
 GET static asset         |        1065 |          0 |      11,760 |          0
 POST (Auth) comment form |        1399 |         30 |      11,752 |         56
 -------------------------+-------------+------------+-------------+-----------
 Aggregated               |        1070 |          9 |      11,768 |          9
 Slowest page load within specified percentile of requests (in ms):
 Name                     |    50% |    75% |    98% |    99% |  99.9% | 99.99%
 GET (Anon) front page    |      0 |      0 | 10,000 | 11,000 | 11,660 | 11,660
 GET (Anon) node page     |     63 |     76 | 10,000 | 11,000 | 11,660 | 11,660
 GET (Anon) user page     |     14 |     18 | 10,000 | 11,000 | 11,645 | 11,645
 GET (Auth) comment form  |     61 |     84 | 11,721 | 11,721 | 11,721 | 11,721
 GET (Auth) front page    |     28 |     36 | 11,760 | 11,760 | 11,760 | 11,760
 GET (Auth) node page     |     59 |     77 | 11,768 | 11,768 | 11,768 | 11,768
 GET (Auth) user page     |     14 |     18 | 11,651 | 11,651 | 11,651 | 11,651
 GET static asset         |      0 |      0 | 10,000 | 11,000 | 11,760 | 11,760
 POST (Auth) comment form |     56 |     71 | 11,752 | 11,752 | 11,752 | 11,752
 -------------------------+--------+--------+--------+--------+--------+-------
 Aggregated               |      9 |      9 | 10,000 | 11,000 | 11,768 | 11,768
```

And the percentile graph:
image

bbros-dev:
Sounds great. Sorry our example hasn't landed we've been wrestling with gaggle and containers....

jeremyandrews:
I've reworked and greatly simplified the Coordinated Omission mitigation logic. Whereas before we were tracking and calculating a cadence for every single request, now we only track the cadence per-GooseUser thread. This removes all the overhead, and it still works great:

image

While working on this I found/fixed a bug in how we merge metrics from all the different GooseUser threads, which sometimes resulted in odd/illogical results. The Request metrics associated with the above graph are now very consistent and match expectations:

```
      Running `target/release/examples/drupal_loadtest -H 'http://apache' -v -u10 -t1m -r5 --requests-file requests.log --requests-format json`
09:37:03 [INFO] Output verbosity level: INFO
09:37:03 [INFO] Logfile verbosity level: WARN
09:37:03 [INFO] users = 10
09:37:03 [INFO] run_time = 60
09:37:03 [INFO] hatch_rate = 5
09:37:03 [INFO] co_mitigation = Some(Average)

 
 === PER REQUEST METRICS ===
 Name                     |        # reqs |        # fails |    req/s |  fail/s
 GET (Anon) front page    |        16,308 |         0 (0%) |   271.80 |    0.00
 GET (Anon) node page     |        10,885 |         0 (0%) |   181.42 |    0.00
 GET (Anon) user page     |         3,272 |         0 (0%) |    54.53 |    0.00
 GET (Auth) comment form  |           300 |         0 (0%) |     5.00 |    0.00
 GET (Auth) front page    |         1,505 |         0 (0%) |    25.08 |    0.00
 GET (Auth) node page     |         1,002 |         0 (0%) |    16.70 |    0.00
 GET (Auth) user page     |           300 |         0 (0%) |     5.00 |    0.00
 GET static asset         |       106,878 |         0 (0%) |     1781 |    0.00
 POST (Auth) comment form |           300 |         0 (0%) |     5.00 |    0.00
 -------------------------+---------------+----------------+----------+--------
 Aggregated               |       140,750 |         0 (0%) |     2346 |    0.00
 Name                     |    Avg (ms) |        Min |        Max |      Median
 GET (Anon) front page    |      895.09 |          1 |      13,572 |          1
 GET (Anon) node page     |      913.85 |          1 |      11,686 |         30
 GET (Anon) user page     |      899.74 |          1 |      11,686 |         14
 GET (Auth) comment form  |        1217 |         15 |      12,546 |         30
 GET (Auth) front page    |        1208 |         16 |      12,546 |         24
 GET (Auth) node page     |        1214 |         15 |      12,546 |         30
 GET (Auth) user page     |        1242 |          8 |      12,546 |         14
 GET static asset         |      914.22 |          1 |      12,546 |          1
 POST (Auth) comment form |        1271 |         25 |      12,546 |         48
 -------------------------+-------------+------------+-------------+-----------
 Aggregated               |      919.02 |          1 |      13,572 |          1
 Slowest page load within specified percentile of requests (in ms):
 Name                     |    50% |    75% |    98% |    99% |  99.9% | 99.99%
 GET (Anon) front page    |      1 |      1 | 10,000 | 11,000 | 12,000 | 12,000
 GET (Anon) node page     |     30 |     42 | 10,000 | 11,000 | 11,686 | 11,686
 GET (Anon) user page     |     14 |     18 | 10,000 | 11,000 | 11,686 | 11,686
 GET (Auth) comment form  |     30 |     41 | 11,000 | 12,000 | 12,546 | 12,546
 GET (Auth) front page    |     24 |     30 | 11,000 | 12,000 | 12,546 | 12,546
 GET (Auth) node page     |     30 |     42 | 11,000 | 12,000 | 12,546 | 12,546
 GET (Auth) user page     |     14 |     20 | 11,000 | 12,000 | 12,546 | 12,546
 GET static asset         |      1 |      1 | 10,000 | 11,000 | 12,000 | 12,000
 POST (Auth) comment form |     48 |     68 | 11,000 | 12,000 | 12,546 | 12,546
 -------------------------+--------+--------+--------+--------+--------+-------
 Aggregated               |      1 |      1 | 10,000 | 11,000 | 12,000 | 13,000
```

Still left to do:
- Capture and subtract any automatic delays between `GooseTasks`
- Clearly summarize how coordinate omission mitigation has affected the numbers (ie, display a comparison with and without and return both in the `GooseMetrics`)
- Documentation
- Apply Coordinated Omission mitigation to Task metrics also (currently we only apply them to Request metrics) (very likely will happen in a followup)
- Tests (very likely will happen in a followup)

jeremyandrews:
The following will be done in followup PRs, as this change is getting large:
- https://github.com/tag1consulting/goose/issues/296
- https://github.com/tag1consulting/goose/issues/297
- https://github.com/tag1consulting/goose/issues/298

PR #295, opened 2021-06-24
[Add][Doc] Gaggle example

bbros-dev:
What?
A self contained example of a gaggle setup.

Why?
This assists end to end testing.
It also suggests Goose functionality that is required for
end-to-end testing a gaggle configuration.
Specifically the requirement to provide IP-address with
`--master-host ...` blocks such test infrastructure.

How?
Provides a Docker or Podman or k3s container cluster.

Signed-off-by: Begley Brothers Ltd <begleybrothers@gmail.com>

### DIFF
--- a/examples/the-internet/ .dockerignore
+++ b/examples/the-internet/ .dockerignore
@@ -1,0 +1,2 @@
+/target/
+/.git/
--- a/examples/the-internet/Cargo.toml
+++ b/examples/the-internet/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "gti"
+version = "0.1.0"
+authors = ["Begley Brothers Ltd"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+goose = { version = "^0.11", features = ["gaggle"] }
+
+[[bin]]
+name = "gti"
+path = "gti.rs"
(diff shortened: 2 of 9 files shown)


jeremyandrews:
Can you provide some documentation on how this should be tested and used? Typically Cargo manages the `examples/` that ship with Goose, but this doesn't appear to work in this way?

```
% ls examples/the-internet 
 .dockerignore		docker-compose.yml	podman-pod.yaml
Cargo.toml		gti.rs			report
Dockerfile		k3s-pod.yml
README.md		log
% cargo run --example the-internet
error: no example target named `the-internet`
```

jeremyandrews:
I see that there's some explanation in the README on the steps needed: this is clearly built as a standalone application, and doesn't seem to belong in the `examples/` directory. I think it would be better to live in its own repository, and perhaps a link can be added to that repo from the Goose documentation in the appropriate place.

bbros-dev:
>  this is clearly built as a standalone application

Not really, but it looks that way because of 2 issues:

1) Docker/Podman compose assumes you use FQDN or `/etc/hostname` entries to reference each container.
2) Goose assumes you use IP addresses to refer to gaggle members.

The intention was that this serve as a basic e2e test for gaggle functionality - so it is blocking on Goose recognizing FQDN's in place of IP addresses.

Not sure if that is on your roadmap?

bbros-dev:
Closing as NA. On reflection, not being integrated with Cargo by definition means it is a standalone application.

PR #300, opened 2021-07-02
improve Coordinated Omission documentation and debug output

jeremyandrews:
- Always log the `GooseUser` thread's cadence (the time it takes to run through all `GooseTasks`) in the requests file (if enabled)
 - Generate an INFO level log message if Coordinated Omission detects an abnormally slow request, for example:
    ```
    10:10:02 [INFO] coordinated omission alert 6.957s into goose attack: "GET http://apache/node/8848" [200] took abnormally long (2932 ms), task name: "(Anon) node page"
    ```
 - update the README with improved documenation


LionsAd:
I would add how long Goose expects it to take not just the abnormally long (42 seconds) normal is 13 seconds. etc.

jeremyandrews:
It doesn’t actually know this. It’s comparing an individual request against the entire time to run ALL requests. (There’s logic in git history to track all requests, but it’s too much overhead).

jeremyandrews:
These are the requests that will trigger Coordinated Omission Mitigation, but they are not themselves generated by Coordinated Omission Mitigation.

In other words, these are actual requests that were made to the server, and report the actual time it took. You'd have to look further in the requests file to find the subsequent back-filled "requests" generated by Coordinated Omission Mitigation. (But there's not a 1:1 match, as Goose is monitoring the per GooseUser cadence, not the per request cadence.)

bbros-dev:
Okay. This sort of spelunking and cross correlating just to get data to analyze if what issue #305 aims to eliminate.

PR #321, opened 2021-07-09
Propagate `rustls` feature to `tokio-tungstenite`

medwards:
Note: even with this change I still end up with `openssl` in my `cargo tree`:
```
├── goose v0.11.2 (/home/medwards/forks/goose)
<snip>
│   ├── tokio-tungstenite v0.14.0
<snip>
│   │   ├── tokio v1.8.1 (*)
│   │   ├── tokio-rustls v0.22.0 (*)
│   │   ├── tungstenite v0.13.0
<snip>
│   │   │   ├── native-tls v0.2.7
│   │   │   │   ├── log v0.4.14 (*)
│   │   │   │   ├── openssl v0.10.32
│   │   │   │   │   ├── bitflags v1.2.1
│   │   │   │   │   ├── cfg-if v1.0.0
│   │   │   │   │   ├── foreign-types v0.3.2
│   │   │   │   │   │   └── foreign-types-shared v0.1.1
│   │   │   │   │   ├── lazy_static v1.4.0
│   │   │   │   │   ├── libc v0.2.95
│   │   │   │   │   └── openssl-sys v0.9.63
│   │   │   │   │       └── libc v0.2.95
│   │   │   │   │       [build-dependencies]
│   │   │   │   │       ├── autocfg v1.0.1
│   │   │   │   │       ├── cc v1.0.68
│   │   │   │   │       └── pkg-config v0.3.19
│   │   │   │   ├── openssl-probe v0.1.4
│   │   │   │   └── openssl-sys v0.9.63 (*)
│   │   │   ├── rand v0.8.3 (*)
│   │   │   ├── rustls v0.19.0 (*)
<snip>
│   │   │   ├── webpki v0.21.4 (*)
│   │   │   └── webpki-roots v0.21.1 (*)
│   │   ├── webpki v0.21.4 (*)
│   │   └── webpki-roots v0.21.1 (*)
│   ├── tungstenite v0.13.0 (*)
│   └── url v2.2.2 (*)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ nng = { version = "1.0", optional = true }
 [features]
 default = ["reqwest/default-tls"]
 gaggle = ["nng"]
-rustls = ["reqwest/rustls-tls"]
+rustls = ["reqwest/rustls-tls", "tokio-tungstenite/rustls-tls"]
 
 [build-dependencies]
 rustc_version = "0.3"


Cargo.toml
```diff
@@ -55,7 +55,7 @@ nng = { version = "1.0", optional = true }
 [features]
 default = ["reqwest/default-tls"]
 gaggle = ["nng"]
-rustls = ["reqwest/rustls-tls"]
+rustls = ["reqwest/rustls-tls", "tokio-tungstenite/rustls-tls]
```
jeremyandrews, line 58: You're missing an end quote here.

jeremyandrews (changes requested):
There's a typo in your PR, a missing " needs to be added at the end of your change.

> Note: even with this change I still end up with openssl in my cargo tree:

Why? Is this a problem?

medwards:
> Why? Is this a problem?

This will require `openssl-dev` in the build image, right? There are no other dependencies in our tree that require it, all the other dependencies are switched to `rustls` using feature flags.

Cargo.toml
medwards: Yeah, did this by hand in the GitHub editor to try and avoid creating a fork, but messed up :( Fixed now

medwards:
A colleague pointed out that `tungstenite` 0.13 has `nativetls` as a default feature. This is fixed in 0.14 which is used in `tokio-tungstenite` 0.15, however `goose` won't compile with this new version.

jeremyandrews:
Thanks! I also committed a followup PR #322 to update tungstenite -- note that the Goose `rustls` feature was renamed to `rustls-tls` so that Goose itself can use `rustls` when compiling examples and tests.

PR #322, opened 2021-07-10
update tungstenite so it's possible to build without openssl

jeremyandrews:
- start with #321
- update `tungstenite` to `0.14` and `tokio-tungstenite` = "0.15"
- add `rustls` as development dependency for `tests/controller.rs`
- rename Goose `rustls` feature to `rustls-tls` as feature and library can't share the same name

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.12.1-dev
+ - rename `rustls` feature to `rustls-tls` so `tests/controller.rs` can build with the `rustls` library; update `tungstenite` to `0.14` and `tokio-tungstenite` = `0.15` to allow building with `rustls`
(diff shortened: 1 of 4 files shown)


jeremyandrews:
With this update, if you enable the `rustls-tls` feature you no longer get openssl in your `cargo tree`:

```
% cargo tree | grep -i openssl
% cargo tree | grep -i tls    
│   ├── hyper-tls v0.5.0
│   │   ├── native-tls v0.2.7
│   │   └── tokio-native-tls v0.3.0
│   │       ├── native-tls v0.2.7 (*)
│   ├── native-tls v0.2.7 (*)
│   ├── tokio-native-tls v0.3.0 (*)
├── native-tls v0.2.7 (*)
├── rustls v0.19.1
```

PR #328, opened 2021-07-14
enable gzip and deflate in http client

jeremyandrews:
- enable [`gzip`](https://docs.rs/reqwest/*/reqwest/struct.ClientBuilder.html#method.gzip) support in the http client
  - can be disabled with `--no-gzip` or `GooseDefault::NoGzip`
  - documented how to manually enable [`deflate`](https://docs.rs/reqwest/*/reqwest/struct.ClientBuilder.html#method.deflate) and/or [`brotli`](https://docs.rs/reqwest/*/reqwest/struct.ClientBuilder.html#method.brotli) compression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
   - documentation cleanup; properly rename `GooseDefault::RequestFormat` and fix links
   - always configure `GooseConfiguration.manager` and `GooseConfiguration.worker`; confirm Manager is enabled when setting `--expect-workers`
   - moved `GooseConfiguration`, `GooseDefault`, and `GooseDefaultType` into new `src/config.rs` file; standardized configuration precedence through internal `GooseConfigure` trait defining `get_value()` for all supported types; general improvements to configuration documentation
+  - enable [`gzip`](https://docs.rs/reqwest/*/reqwest/struct.ClientBuilder.html#method.gzip) and [`deflate`](https://docs.rs/reqwest/*/reqwest/struct.ClientBuilder.html#method.deflate) support in the client
 
 ## 0.12.0 July 8, 2021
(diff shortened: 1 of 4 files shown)


jeremyandrews:
I ran current -dev w/o this patch, and current -dev w/ this patch, each 3 times as follows (restarting all server processes before each batch of 3):

```
cargo run --release --example drupal_loadtest -- --host http://apache -v -u12 -t10m && sleep 30 && cargo run --release --example drupal_loadtest -- --host http://apache -v -u12 -t10m && sleep 30 && cargo run --release --example drupal_loadtest -- --host http://apache -v -u12 -t10m
```

Here's the data:
https://docs.google.com/spreadsheets/d/1lsOagHY1yISQ_HcpwbIxoxXx8s2q5KxKYXvZmhgsMYs/edit#gid=0

Goose made >30% more anonymous requests, but 5.6% fewer authenticated requests with this patch applied. Without compression it averaged 6,473 requests per second. With compression it averaged 9,542 requests per second. It would be good to re-run the test with fewer GooseUsers and an evenly balanced number of Anonymous and Authenticated users: my assumption is that the reduction in authenticated traffic is simply because anonymous traffic was being returned so quickly by Varnish that it was getting scheduled far more.

The Apache/Varnish server saw no change in CPU usage (as Varnish caches compressed pages), whereas the Goose server saw an increase in CPU usage, visible here (Goose is running in a 32-core VM):
CPU usage

Though it made considerably more requests, obviously there was considerably less network traffic generated:
Network usage

LionsAd (approved):

LionsAd:
Let’s also add the request headers to the debug log and:or add a typical HTT header to the docs.

jeremyandrews:
Authenticated requests slow down as Apache is compressing each request due to how my server is configured. There's additional CPU usage showing up on the Apache server because of this.

Issue #65, opened 2020-06-08
Provide a per-client-thread mutable bucket for arbitrary data


jeremyandrews:
When working with Locust, you can arbitrarily add fields to the client object, and this is available anywhere the client is invoked.  This ultimately allows the sharing of arbitrary data between tasks within a taskset.

My general idea is to add a single field to the GooseClientState struct, which itself is either a HashMap or a BTreeMap of mutex-wrapped-generics. Then, when writing a load test, you could optionally create a custom Struct with whatever fields you want to share between tasks within a taskset and read/write from/to it as needed.

There are advantages to both structures, likely it would be best to add _two_ buckets to the load test: one that's a HashMap, and one that's a BTreeMap. If a load test doesn't use a bucket, there's not enough overhead to matter. In general the preference would be to use the HashMap for best performance, and the BTreeMap when you need to do more complex comparisons etc.

A nice overview of the two collection types, and the advantages/disadvantages of each:
https://doc.rust-lang.org/std/collections/index.html

jeremyandrews:
I believe the problem w/ my proposed solution is how we're storing client states in a lazy_static. I believe this requires us to define the object that will be in the HashMap and/or BTreeMap at compile time, which makes it far less useful.

Perhaps this could be solved through some unsafe code, it needs further thought/architecture.

jeremyandrews:
Here's the direction I was originally thinking -- this doesn't compile as we have to declare the type in the lazy_static (which is visible in the first line of the patch) and this would prevent users from declaring their own arbitrary structures as desired:

```diff
diff --git a/src/lib.rs b/src/lib.rs
index 0dd1a3a..52cd63a 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -340,7 +340,7 @@ lazy_static! {
     static ref CLIENT: RwLock<Vec<GooseClientState>> = RwLock::new(Vec::new());
 }
 
-struct GooseClientState {
+struct GooseClientState<T> {
     /// A Reqwest client, wrapped in a Mutex as a read-write copy is always needed to
     /// manage things like sessions and cookies.
     client: Mutex<Client>,
@@ -350,8 +350,9 @@ struct GooseClientState {
     /// Integer value indicating which task within the current sequenced bucket is currently
     /// running.
     weighted_bucket_position: AtomicUsize,
+    storage: HashMap<String, T>,
 }
-impl GooseClientState {
+impl<T> GooseClientState<T> {
     // Initialize one client per thread.
[...]

PR #355, opened 2021-08-30
Add GooseUser session data

finnefloyd:
The goal of this PR is to add the possibility to set session data on each GooseUser running.  
This feature is useful for multiple use cases but I'm using it for authenticated users.

Changes made:
- Remove clone trait from GooseUser because I don't want to force session data to be cloneable.
- Set the GooseUser reference as mutable in GooseTask, mandatory because you need mutable reference in order to be able to update session data
- Remove Arc/Mutex/RWLock/Atomics because they are not needed anymore with the mutable reference in GooseTask
- Add GooseSessionData trait to allow generic session data for each TaskSet if needed.


jeremyandrews:
Thanks for the contribution! I'm going to be offline for ~2 weeks so may not be able to review thorough before I return.

I believe this would close https://github.com/tag1consulting/goose/issues/65

finnefloyd:
Thank you for taking this into consideration.
Yes I believe I'll close the issues [65](https://github.com/tag1consulting/goose/issues/65)

jeremyandrews:
I ran a performance comparison to ensure no regressions with these changes, and instead measured a 4-5% speedup. Before each test I reload the database and restart all services. The load test is between two VMs on the same host. I performed the following command (running a 10 minute load test 3 times with a short pause between each):

```bash
cargo run --release --example drupal_loadtest -- --host http://apache -u10 -v -s5 -t10m && sleep 30 && cargo run --release --example drupal_loadtest -- --host http://apache -u10 -v -s5 -t10m && sleep 30 && cargo run --release --example drupal_loadtest -- --host http://apache -u10 -v -s5 -t10m
```

This PR increased Goose's overall ability to make and analyze requests with the same load test plan and configuration on average from 8,000 requests per second to 8,344 requests per second, a 4.13% increase.

The high level results can be compared here:
https://docs.google.com/spreadsheets/d/1eMC5IgUCUrvKSnraIwxx7h_riYeEHOh9Ei45XlfwPU8/edit?usp=sharing

The full Goose command line output can be viewed here:
https://docs.google.com/document/d/1Oc_IbAOVfeEaTzOAZIZGhOnGFh_3eu1FkuAGL2STXOw/edit?usp=sharing

examples/simple_with_session.rs
```diff
@@ -0,0 +1,79 @@
...
+    user.set_session_data(Session {
+        jwt_token: response.jwt_token,
+    });
+
+    Ok(())
+}
+
+/// A very simple task that simply loads the front page.
+async fn authenticated_index(user: &mut GooseUser) -> GooseTaskResult {
+    let session = user.get_session_data_uncheck::<Session>();
```
jeremyandrews, line 74: My initial testing of this example led to a panic. I feel it would be better to provide an example using something like:
```rust
    if let Some(session) = user.get_session_data::<Session>() {
```

...instead of the `_uncheck()` version, under the assumption that people will copy these examples when writing their own load tests.  In the event that `get_session_data` fails, I imagine as we would expect session data we should set a failure: https://docs.rs/goose/0.13.3/goose/goose/struct.GooseUser.html#method.set_failure

jeremyandrews: (Or another option: add a comment about why this might panic, and about instead using `get_session_data`... if the goal is just as simple of code as possible.)

src/goose.rs
```diff
@@ -913,6 +910,153 @@ impl GooseUser {
...
+    /// Leaves the session in-place, creating a new one Option with a reference
+    /// to the original session data.
+    ///
+    /// Return none if no session data has been set or that the session data is not of type `T`
+    ///
+    /// # Example
+    /// ```rust
+    /// use goose::prelude::*;
+    ///
+    /// #[derive(Debug, Clone)]
```
jeremyandrews, line 923: Thanks to your PR, this `Clone` isn't required, let's remove from the documentation example.

src/goose.rs
```diff
@@ -913,6 +910,153 @@ impl GooseUser {
...
+    ///
+    /// # Safety
+    ///
+    /// Calling this method on an GooseUser without session data or with a different type `T` is *[undefined behavior]*
+    ///
+    /// # Example
+    /// ```rust
+    /// use goose::prelude::*;
+    ///
+    /// #[derive(Debug, Clone)]
```
jeremyandrews, line 956: Again `Clone` isn't required here, let's remove.

src/goose.rs
```diff
@@ -913,6 +910,153 @@ impl GooseUser {
...
+
+    /// Returns a mutable reference to GooseUser's session data.
+    ///
+    /// Return none if no session data has been set or that the session data is not of type `T`
+    ///
+    /// # Example
+    /// ```rust
+    /// use goose::prelude::*;
+    ///
+    /// #[derive(Debug, Clone)]
```
jeremyandrews, line 985: Again `Clone` isn't required here, let's remove.

src/goose.rs
```diff
@@ -913,6 +910,153 @@ impl GooseUser {
...
+    ///
+    /// # Safety
+    ///
+    /// Calling this method on an GooseUser without session data or with a different type `T` is *[undefined behavior]*
+    ///
+    /// # Example
+    /// ```rust
+    /// use goose::prelude::*;
+    ///
+    /// #[derive(Debug, Clone)]
```
jeremyandrews, line 1016: Again `Clone` isn't required here, let's remove.

src/goose.rs
```diff
@@ -1513,62 +1657,67 @@ impl GooseUser {
...
             // Grab the current timestamp to calculate the difference since the last
             // time through the loop.
             let now = std::time::Instant::now();
 
             // Swap out the `slept` counter, which is the total time the GooseUser slept
             // between tasks, a potentially randomly changing value. Reset to 0 for the
             // next loop through all GooseTasks.
-            request_cadence.delays_since_last_time = self.slept.swap(0, Ordering::SeqCst);
+            self.request_cadence.delays_since_last_time = self.slept;
+            self.slept = 0;
```
jeremyandrews, line 1675: Why not continue using swap here?

jeremyandrews:
A lot of nice cleanups and simplifications here, and a handy new feature -- thanks! Once we address a few little nits, I plan to merge -- this will bump the Goose version to 0.14 as well as it's an API change.

If you are comfortable adding something to CHANGELOG.txt documenting at a high level what's changed, please do. Otherwise I'll do so in a followup.

src/goose.rs
finnefloyd: Oh yes, my bad.

src/goose.rs
finnefloyd: I change `slept` from `AtomicU64` to `u64`, it's not possible to use swap anymore. 
Will `self.request_cadence.delays_since_last_time` always be `0` when I reach this part of the code? If yes I could use `std::mem::swap`.
Am I missing something?

examples/simple_with_session.rs
finnefloyd: I added a comment because `set_failure`  require a `GooseRequestMetric` which is not available at that moment. 

Maybe we could enhance `GooseTaskError` to handle custom error.

finnefloyd:
If you don't mind waiting a bit longer for the release, there is a few other PR I'd like to submit to you.

src/goose.rs
jeremyandrews: Perhaps I'm missing something. `self.request_cadence.delays_since_last_time` is always set to `self.slept` (which wouldn't be `0`), and then `self.slept` is set to `0`.

jeremyandrews:
> If you don't mind waiting a bit longer for the release, there is a few other PR I'd like to submit to you.

Definitely! I'm eager to see whatever else you contribute! And as we're going to be bumping the API version I'll want to review if there are other changes we can/should make at this time.

src/goose.rs
finnefloyd: Oh oki, I think we should only use `std::men::swap` if we really want to swap the value. In this case we reset one of the value to 0 right after the swap. It would make the code a bit unclear IMHO

finnefloyd:
> Definitely! I'm eager to see whatever else you contribute! And as we're going to be bumping the API version I'll want to review if there are other changes we can/should make at this time.

Great, I'll open them once this one is merged otherwise I'll have conflicts with myself.

jeremyandrews:
I'm on the road today, I hope to do a final review and merge tomorrow morning. Thanks again!

src/goose.rs
```diff
@@ -811,21 +811,22 @@ impl GooseRequestCadence {
     }
 }
 
+pub trait GooseUserData: Downcast + Send + Sync + 'static {}
```
jeremyandrews, line 814: This needs documentation (can be in followup).

src/goose.rs
```diff
@@ -811,21 +811,22 @@ impl GooseRequestCadence {
...
+impl_downcast!(GooseUserData);
+impl<T: Send + Sync + 'static> GooseUserData for T {}
+
 /// An individual user state, repeatedly running all [`GooseTask`](./struct.GooseTask.html)s
 /// in a specific [`GooseTaskSet`](./struct.GooseTaskSet.html).
-#[derive(Debug, Clone)]
```
jeremyandrews, line 816: In your comments you mentioned removing `Clone`, but here you're also removing `Debug`. Either add it back here, or remove `Debug` from the various `GooseUserData` examples for consistency.

It's not strictly required here, but it can be quite useful so I'd lean toward adding it back unless there's an important reason to remove it.

src/goose.rs
```diff
@@ -811,21 +811,22 @@ impl GooseRequestCadence {
...
 pub struct GooseUser {
     /// The Instant when this `GooseUser` client started.
     pub started: Instant,
     /// An index into the internal [`GooseAttack`](../struct.GooseAttack.html)`.task_sets`
     /// vector, indicating which [`GooseTaskSet`](./struct.GooseTaskSet.html) is running.
     pub task_sets_index: usize,
     /// Client used to make requests, managing sessions and cookies.
-    pub client: Arc<Mutex<Client>>,
-    /// Integer value tracking the current task the user is running.
-    pub position: Arc<AtomicUsize>,
```
jeremyandrews, line 826: I'm very happy to see `position` removed! 👍

src/goose.rs
```diff
@@ -913,6 +910,153 @@ impl GooseUser {
...
+    ///
+    /// /// A very simple task that makes a GET request.
+    /// async fn get_session_data_uncheck_function(user: &mut GooseUser) -> GooseTaskResult {
+    ///     let foo = user.get_session_data_uncheck::<Foo>();
+    ///     println!("Session data: {}", foo.0);
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn get_session_data_uncheck<T: GooseUserData>(&self) -> &T {
```
jeremyandrews, line 969: It seems the standard naming for helpers that don't do any checking is `_unchecked()` not `_uncheck()`:
https://doc.rust-lang.org/std/?search=uncheck

For consistency with Rust's std lib please rename this to `get_session_data_unchecked`.

src/goose.rs
```diff
@@ -913,6 +910,153 @@ impl GooseUser {
...
+    /// let mut task = task!(get_mut_session_data_uncheck_function);
+    ///
+    /// /// A very simple task that makes a GET request.
+    /// async fn get_mut_session_data_uncheck_function(user: &mut GooseUser) -> GooseTaskResult {
+    ///     let foo = user.get_mut_session_data_uncheck::<Foo>();
+    ///     foo.0 = "Bar".to_owned();
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn get_mut_session_data_uncheck<T: GooseUserData>(&mut self) -> &mut T {
```
jeremyandrews, line 1028: It seems the standard naming for helpers that don't do any checking and return a mutable is `_unchecked_mut()` not `mut_ .. _uncheck()`:
https://doc.rust-lang.org/std/pin/struct.Pin.html#method.get_unchecked_mut

For consistency with Rust's std lib please rename this to `get_session_data_unchecked_mut`.

src/goose.rs
```diff
@@ -913,6 +910,153 @@ impl GooseUser {
...
+    /// let mut task = task!(get_mut_session_data_function);
+    ///
+    /// /// A very simple task that makes a GET request.
+    /// async fn get_mut_session_data_function(user: &mut GooseUser) -> GooseTaskResult {
+    ///     let foo = user.get_mut_session_data::<Foo>().expect("Missing session data!");
+    ///     foo.0 = "Bar".to_owned();
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn get_mut_session_data<T: GooseUserData>(&mut self) -> Option<&mut T> {
```
jeremyandrews, line 997: For consistency w/ the Rust std lib, and the changes recommended below, please rename this to `get_session_data_mut`.

src/goose.rs
```diff
@@ -2444,6 +2582,7 @@ impl Hash for GooseTask {
 }
 
 #[cfg(test)]
+#[allow(clippy::blacklisted_name)]
```
jeremyandrews, line 2585: Instead of allowing blacklisted variable names, let's choose more descriptive variables below.

src/goose.rs
```diff
@@ -2832,4 +2967,83 @@ mod tests {
         assert_eq!(goose.request.status_code, 200);
         comment.assert_hits(1);
     }
+
+    #[test]
+    fn test_set_session_data() {
+        #[derive(Debug, PartialEq, Eq, Clone)]
+        struct Foo {
```
jeremyandrews, line 2974: Perhaps rename this to `CustomUserData`.

src/goose.rs
```diff
@@ -2832,4 +2967,83 @@ mod tests {
...
+            bar: String,
```
jeremyandrews, line 2975: And this to `custom_data`.

src/goose.rs
```diff
@@ -2832,4 +2967,83 @@ mod tests {
         assert_eq!(goose.request.status_code, 200);
         comment.assert_hits(1);
     }
+
+    #[test]
+    fn test_set_session_data() {
+        #[derive(Debug, PartialEq, Eq, Clone)]
```
jeremyandrews, line 2973: `Eq` is not needed here.

src/goose.rs
```diff
@@ -2832,4 +2967,83 @@ mod tests {
...
+        assert!(session.is_some());
+        assert_eq!(session.unwrap(), &foo);
+
+        let session = user.get_session_data_uncheck::<Foo>();
+        assert_eq!(session, &foo);
+    }
+
+    #[test]
+    fn test_get_mut_session_data() {
+        #[derive(Debug, PartialEq, Eq, Clone)]
```
jeremyandrews, line 2998: These derives are not needed here, remove.

src/goose.rs
```diff
@@ -2832,4 +2967,83 @@ mod tests {
...
+
+        let session = user.get_mut_session_data_uncheck::<Foo>();
+        session.bar = "bar".to_owned();
+        let session = user.get_session_data_uncheck::<Foo>();
+        assert_eq!(session.bar, "bar".to_string());
+    }
+
+    #[test]
+    fn test_set_session_data_override() {
+        #[derive(Debug, PartialEq, Eq, Clone)]
```
jeremyandrews, line 3028: Only `Clone` needs to be derived here, remove the rest.

tests/common.rs
```diff
@@ -64,16 +64,16 @@ pub fn build_configuration(server: &MockServer, custom: Vec<&str>) -> GooseConfi
...
 /// Launch each Worker in its own thread, and return a vector of Worker handles.
 #[allow(dead_code)]
-pub fn launch_gaggle_workers(
-    // A goose attack object which is cloned for each Worker.
-    goose_attack: GooseAttack,
+pub fn launch_gaggle_workers<F: Fn() -> GooseAttack>(
     // The number of Workers to launch.
     expect_workers: usize,
+    // A goose attack object which is cloned for each Worker.
+    goose_attack_provider: F,
```
jeremyandrews, line 71: Nice cleanup! 👍

src/goose.rs
jeremyandrews: (As it's a pub trait, documentation is important and will show up at https://docs.rs/goose/)

src/goose.rs
```diff
@@ -845,19 +846,17 @@ pub struct GooseUser {
     /// An index into the internal [`GooseAttack`](../struct.GooseAttack.html)`.weighted_users`
     /// vector, indicating which weighted `GooseUser` is running.
     pub weighted_users_index: usize,
-    /// A weighted list of all tasks that run when the user first starts.
-    pub weighted_on_start_tasks: WeightedGooseTasks,
-    /// A weighted list of all tasks that this user runs once started.
-    pub weighted_tasks: WeightedGooseTasks,
-    /// A weighted list of all tasks that run when the user stops.
-    pub weighted_on_stop_tasks: WeightedGooseTasks,
```
jeremyandrews, line 853: Nice de-duplication, moving this from `GooseUser` to `GooseTaskSet` 👍

jeremyandrews:
I was finally able to sit down and do a full read-through of your PR -- I'm still very happy with it, thanks again! There's some nice cleanup in here, in addition to adding session support. 

If you can please rename a few functions for general consistency per my comments, and address a few other final nits, then once addressed this is ready to merge.

src/goose.rs
finnefloyd: Debug is needed by `assert!` and assert_eq!` macro

src/goose.rs
finnefloyd: removed

src/goose.rs
jeremyandrews: I removed locally and re-ran tests. It seems it wasn’t necessary?

finnefloyd:
Thank you for the review. 

I have addressed all your remarks.

jeremyandrews:
Fantastic contribution, thanks!

PR #356, opened 2021-09-01
Make GooseAttack.execute async and remove Tokio runtime initialization

finnefloyd:
In complex use case, a user may need to run some code alongside the GooseAttack which is currently complicated be Goose create a tokio Runtime which prevent user to use #[tokio::main] and run other futures.

This change remove the runtime initialization from Goose and make the function `execute` async.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
  - update `GooseTaskSet::set_wait_time()` to accept `std::time::Duration` instead of `usize` allowing more granularity (API change)
  - use request name when displaying errors to avoid having a large volume of distinct error for the same endpoint when using path parameters
  - updated `tungstenite` dependency to [`0.15`](https://github.com/snapview/tungstenite-rs/blob/master/CHANGELOG.md)
+ - Make `GooseAttack.execute` async
 
 ## 0.13.3 August 25, 2021
  - document GooseConfiguration fields that were only documented as gumpdrop parameters (in order to generate new lines in the help output) so now they're also documented in the code
--- a/README.md
+++ b/README.md
@@ -82,7 +82,8 @@ The function is declared `async` so that we don't block a CPU-core while loading
 We have to tell Goose about our new task function. Edit the `main()` function, setting a return type and replacing the hello world text as follows:
 
 ```rust
-fn main() -> Result<(), GooseError> {
+#[tokio::main]
+async fn main() -> Result<(), GooseError> {
     GooseAttack::initialize()?
         .register_taskset(taskset!("LoadtestTasks")
             .register_task(task!(loadtest_index))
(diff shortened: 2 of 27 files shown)


jeremyandrews:
It seems the tests aren’t working.

finnefloyd:
I ran them locally and they passed. 

it look's like there is an issues with the runner
`2021-09-01T09:21:49.0070419Z Found online and idle hosted runner in the current repository's organization account that matches the required labels: 'ubuntu-latest'
2021-09-01T09:21:49.0764957Z Waiting for a Hosted runner in the 'organization' to pick this job...
2021-09-01T09:21:49.3159568Z Job is waiting for a hosted runner to come online.
2021-09-01T09:21:53.4901978Z Job is about to start running on the hosted runner: GitHub Actions 4 (hosted)`

Are you able to restart it? I'm not.

jeremyandrews:
I’ve restarted it

jeremyandrews:
I re-ran the tests and they failed again - perhaps we need additional configuration in order for the tests to run properly in GitHub actions.

I tried running the tests locally, and ran into a different problem with `tests/logs.rs` where 11 tests failed all with the following error (on Mac OS X):
```
thread '<unnamed>' panicked at 'Cannot build local tokio runtime: Os { code: 24, kind: Other, message: "Too many open files" }', /Users/jeremyandrews/.cargo/registry/src/github.com-1ecc6299db9ec823/httpmock-0.6.2/src/api/server.rs:285:14
thread 'test_tasks_logs_raw_gaggle' panicked at 'Cannot get server address: RecvError(())', /Users/jeremyandrews/.cargo/registry/src/github.com-1ecc6299db9ec823/httpmock-0.6.2/src/api/server.rs:290:37
```

jeremyandrews:
If locally I remove `flavor = "multi_thread"` I see the same behavior as GitHub actions: I wonder if it's not respecting this setting, or if it has different defaults.

finnefloyd:
I finally manage to fix the CI. 

The issues was that Tokio must create a run with the number of threads equal to the number of core of the machine it's running on which in our case must be one. But it's not enough because `nng` is not async in a Rust point of view so to block the thread used in the tokio runtime. The solution is to set a minimum of 8 threads for all the gaggle tests.

jeremyandrews (changes requested):
I very much like removing the calls to `Runtime::new()` to create a runtime in Goose and instead make `GooseAttack.execute()` itself asynchronous as you've done in this PR. All that's needed is updating the inline documentation consistently to demonstrate this change everywhere affected, as well as the README.

There seems to be some unrelated changes in this PR that would also ideally live in a separate PR to give cleaner git history (sometimes helpful in tracking down regressions etc).

finnefloyd:
Thanks for the review. I addressed all your comments.

jeremyandrews:
There are still quite a few calls to `.execute()?` instead of `.execute().await?` in the README which should be updated to avoid confusing people using the examples.

jeremyandrews (approved):

PR #357, opened 2021-09-01
Allow shared reqwest client configuration

finnefloyd:
From Reqwest official documentation:

> The Client holds a connection pool internally, so it is advised that you create one and reuse it.
    You do not have to wrap the Client it in an Rc or Arc to reuse it, because it already uses an Arc internally.

This PR, use a shared reqwest client among all GooseUser and allow the user the configure the global client.

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -286,9 +286,11 @@
 //! limitations under the License.
 
 use downcast_rs::{impl_downcast, Downcast};
+use http::header::{Entry, IntoHeaderName};
 use http::method::Method;
+use http::HeaderValue;
 use reqwest::cookie::{CookieStore, Jar};
-use reqwest::{cookie, header, Client, ClientBuilder, RequestBuilder, Response};
+use reqwest::{cookie, header, Client, RequestBuilder, Response};
 use serde::{Deserialize, Serialize};
 use std::hash::{Hash, Hasher};
 use std::sync::Arc;
@@ -853,6 +855,8 @@ pub struct GooseUser {
     session_data: Option<Box<dyn GooseUserData>>,
 
     cookie_store: Jar,
+
+    headers: header::HeaderMap,
 }
 impl GooseUser {
     /// Create a new user state.
@@ -883,6 +887,7 @@ impl GooseUser {
             task_name: None,
             session_data: None,
             cookie_store: cookie::Jar::default(),
+            headers: header::HeaderMap::default(),
         })
     }
 
@@ -1581,7 +1586,11 @@ impl GooseUser {
             self.weighted_users_index,
         );
 
+        // Set cookies
         self.add_cookie_header(&mut request);
+
+        // Set default headers
+        self.add_default_headers(&mut request);
 
         // Make the actual request.
         let response = self.client.execute(request).await;
@@ -1643,10 +1652,22 @@ impl GooseUser {
         Ok(GooseResponse::new(request_metric, response))
     }
 
+    // Add cookies from the cookie store.
     fn add_cookie_header(&self, request: &mut reqwest::Request) {
         if request.headers().get(header::COOKIE).is_none() {
             if let Some(header) = self.cookie_store.cookies(request.url()) {
                 request.headers_mut().insert(header::COOKIE, header);
+            }
+        }
+    }
+
+    // insert default headers in the request headers
+    // without overwriting already appended headers.
+    fn add_default_headers(&self, request: &mut reqwest::Request) {
+        let headers = request.headers_mut();
+        for (key, value) in &self.headers {
+            if let Entry::Vacant(entry) = headers.entry(key) {
+                entry.insert(value.clone());
             }
         }
     }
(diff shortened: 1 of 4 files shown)


jeremyandrews:
Merge conflict needs to be resolved

jeremyandrews:
It's unclear to me if this will result in cookies and headers being shared between all `GooseUser`s, which would be undesirable.

finnefloyd:
> It's unclear to me if this will result in cookies and headers being shared between all GooseUsers, which would be undesirable.

Goose point, I did not think about it and I'm not using cookies in my current load tests, I didn't detect it. I moved the cookies handling to the `GooseUser` so we can shared the connection pool

jeremyandrews:
This addresses cookies, but there’s still a problem from shared headers. For example, if using http header authentication, with this PR once one GooseUser logs in all would be.

jeremyandrews:
For example, as documented here it's currently possible to set different headers with different GooseUsers, and this is required functionality for many load tests:
https://docs.rs/goose/*/goose/goose/struct.GooseUser.html#example-18

jeremyandrews:
This example will also need to be updated to show how to set custom cookies:
https://docs.rs/goose/*/goose/goose/struct.GooseUser.html#example-19

finnefloyd:
I added support for default headers and custom cookies to the GooseUser.

src/goose.rs
```diff
@@ -2041,152 +2071,55 @@ impl GooseUser {
...
+    ///     );
     ///
     ///     Ok(())
     /// }
     /// ```
+    pub fn add_cookie_str(&mut self, cookie: &str) {
+        self.cookie_store.add_cookie_str(cookie, &self.base_url);
+    }
+
+    /// Default headers can be manually set to each [`GooseUser`](./goose/struct.GooseUser.html).
```
jeremyandrews, line 2099: `set on` instead of `set to`

src/goose.rs
```diff
@@ -2041,152 +2071,55 @@ impl GooseUser {
...
+    /// The headers will automatically been added to each request made by the [`GooseUser`](./goose/struct.GooseUser.html).
```
jeremyandrews, line 2100: `be added` instead of `been added`

src/goose.rs
```diff
@@ -2041,152 +2071,55 @@ impl GooseUser {
...
-    /// [`reqwest::Client::builder()`], following the documentation above.
-    ///
-    /// # Custom Cookies
-    /// Custom cookies can also be manually set when building a custom [`reqwest::Client`]. This requires
-    /// loading the [`GooseUser::base_url`] being load tested in order to properly build the cookie. Then
-    /// a custom [`reqwest::cookie::Jar`] is created and the custom cookie is added with
-    /// [`reqwest::cookie::Jar::add_cookie_str`]. Finally, the new cookie jar must be specified as the
-    /// [`reqwest::ClientBuilder::cookie_provider`] for the custom client.
-    ///
+    /// Default headers doesn't overwrite already appended headers.
```
jeremyandrews, line 2102: `do not` (or `don't`) instead of `doesn't`

src/lib.rs
```diff
@@ -842,9 +841,55 @@ impl GooseAttack {
...
+    ///
+    ///     GooseAttack::initialize()?
+    ///         .set_scheduler(GooseScheduler::Random)
+    ///         .set_client_builder(client_builder)?;
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn set_client_builder(mut self, client_builder: ClientBuilder) -> Result<Self, GooseError> {
+        self.client = client_builder.cookie_store(false).build()?;
```
jeremyandrews, line 889: This seems to make it impossible to enable the `cookie_store` -- if that's the intent it needs to be documented.

jeremyandrews (changes requested):
What is the intent of this change? Why are you trying to share a single Client pool?

Goose intentionally maintains a unique client-per-`GooseUser`, with the intent of better simulating independent users -- sharing a Client pool adds complexity and I worry this could result in unintended bugs. It also further ties Goose to the inner-workings of Reqwest, whereas the goal is to enhance Goose to support other client libraries and be less-tied to specific Reqwest implementation details.

I imagine there could be reduced memory requirements through this change, and possibly better startup-performance? (I am on the road an unable to test/validate this, which seems to be the only potential advantage of this added complexity?)

At this time I do not see sufficient value in merging this PR.

LionsAd:
@finnefloyd Thanks for all your contributions so far!

I share Jeremy's sentiment that I would like to understand the motivation of the change a little bit more, before considering this PR.

netom:
> @finnefloyd Thanks for all your contributions so far!
> 
> I share Jeremy's sentiment that I would like to understand the motivation of the change a little bit more, before considering this PR.

I used Goose recently, and I needed to create thoudands of users. Starting Goose took about thirty seconds. The tests themselves were not terribly long, so the startup time was significant.

I solved this by monkey-patching goose to use a single reqwest client. (Cookies and headers weren't an issue for me.) This way the statup time disappeared.

I know this is a corner case, but IMHO worth implementing client sharing.

The general argument would be "because it's just more resource efficient".

PR #358, opened 2021-09-01
Allow to set duration instead of seconds for task wait configuration

finnefloyd:
I some cases, you want to have a high number of request from a small amount of users which is not possible with second as the time units. This allow you to set a waiting time in Duration.

### DIFF
--- a/src/user.rs
+++ b/src/user.rs
@@ -1,5 +1,7 @@
+use futures::future::Fuse;
+use futures::{pin_mut, select, FutureExt};
 use rand::Rng;
-use std::time;
+use std::time::{self, Duration, Instant};
 
 use crate::get_worker_id;
 use crate::goose::{GooseTaskFunction, GooseTaskSet, GooseUser, GooseUserCommand};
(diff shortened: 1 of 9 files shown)


src/user.rs
```diff
@@ -50,73 +52,66 @@ pub(crate) async fn user_main(
...
-                            GooseUserCommand::Exit => {
-                                break 'launch_tasks;
-                            }
-                            command => {
-                                debug!("ignoring unexpected GooseUserCommand: {:?}", command);
-                            }
-                        }
-                        message = thread_receiver.try_recv();
+        let mut task_iter = thread_task_set.weighted_tasks.iter().cycle();
+        let next_task = Fuse::terminated();
```
jeremyandrews, line 56: Please name this something that better documents what this variable is for, such as `time_to_next_task`. (I initially assumed it held a GooseTask from the name.)

src/user.rs
```diff
@@ -50,73 +52,66 @@ pub(crate) async fn user_main(
...
-                        );
-                        tokio::time::sleep(sleep_duration).await;
-                        slept += 1;
-                        if slept > wait_time {
-                            in_sleep_loop = false;
-                        }
+
+                    // Determine which task we're going to run next.
+                    let task = &thread_task_set.tasks[*thread_task_index];
+                    let function = &task.function;
```
jeremyandrews, line 80: This comment doesn't really make sense anymore, perhaps combine these two lines of code and rework the comment as follows:

```rust
                    // Get a reference to the task function we're going to invoke next.
                    let function = &thread_task_set.tasks[*thread_task_index].function;
```

jeremyandrews:
Another very nice PR, thanks again! I'll merge once you address the minor feedback above to make the code a little more readable.

jeremyandrews:
All it needs is a line added to CHANGELOG, and like the other API changes note that this is too.

PR #359, opened 2021-09-02
Replace path by request name in error message

finnefloyd:
Replace path by request name in error message to avoid having a large volume of distinct error for the same endpoint when using path parameters


jeremyandrews:
I did some local testing and agree this is preferred -- and if the request name isn't set it will display the paths anyways.

PR #368, opened 2021-10-10
optimize fast path with Duration-based wait time

### LINKED ISSUE #367: Throughput performance drop
We are experiencing significant throughput regression after commit ef711076f16cf2f4e4d492c68a087dc64250eff3 which was introduced with #358.
Before this commit our tests were able to run at **~18000** RPS and it was possible to detect the bottleneck of our service, but after ef711076f16cf2f4e4d492c68a087dc64250eff3 we only getting around **~6000** RPS with the same tests and now Goose is bottleneck.

jeremyandrews:
- reverted https://github.com/tag1consulting/goose/pull/358
 - re-implemented with minimal changes to the underlying logic
 - optimized the fast path avoiding unnecessary Duration->integer conversions
 - fixes #367 -- the issue was the amount of time it take to convert Duration->integer, while small, adds up in the fast path where this can be happening tens of thousands of times a second: the code is optimized to not perform any conversions if there is no wait time set

**0.13.3**
- 9,668
- 10,385
- 10,031

**0.14.1-dev (before this PR)**
- 7,526
- 7,739
- 7,669

**0.14.1-dev (after this PR)**
- 9,919
- 10,206
- 10,082

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -55,17 +55,17 @@
 //!
 //! ### Task Set Wait Time
 //!
-//! Wait time is specified as a low-high integer range. Each time a task completes in
-//! the task set, the user will pause for a random number of seconds inclusively between
+//! Wait time is specified as a low-high Duration range. Each time a task completes in the
+//! task set, the user will pause for a random number of milliseconds inclusively between
 //! the low and high wait times. In the following example, users loading `foo` tasks will
-//! sleep 0 to 3 seconds after each task completes, and users loading `bar` tasks will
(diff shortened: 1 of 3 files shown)


jeremyandrews:
I've run a comparison of 0.13.3, 0.14.1-dev without this PR, and 0.14.1-dev with this PR, confirming that this seems to fix the reported performance regression:
https://docs.google.com/spreadsheets/d/1E7LwL_kJmRgRJ_BVW3QPTIoe7hqmODGhOpiGWjppuNY/edit#gid=0

Issue #370, opened 2021-10-18
errors updated to success with set_success() show up in error summary


jeremyandrews:
Create a task that intentionally loads a 404 page, for example:

```rust
// Load a random non-existent and uncached page.
async fn random_404_page(user: &mut GooseUser) -> GooseTaskResult {
    let mut goose = user
        .get(&format!(
            "{}/{}/{}",
            goose_eggs::text::random_word(6),
            goose_eggs::text::random_word(4),
            goose_eggs::text::random_word(3)
        ))
        .await?;

    // Goose will automatically flag a 404 as an error, but in this case a 404 is
    // expected so update it as a success in the metrics.
    if let Ok(response) = &goose.response {
        if response.status() == 404 {
            user.set_success(&mut goose.request)?;
        }

    Ok(())
}
```

The metrics will show 0 errors when running this task, but confusingly errors still show up in the error summary at the end of the metrics, for example:

```
 === ERRORS ===
 ------------------------------------------------------------------------------
 Count       | Error
 ------------------------------------------------------------------------------
 16            GET 404 page: 404 Not Found: 404 page
 ------------------------------------------------------------------------------
```

When running `set_success()` the expectation is for the error to not show up in the error summary.

jeremyandrews:
Fixed in 0.15

PR #373, opened 2021-10-29
introduce GooseRequest and builder pattern

jeremyandrews:
- **API change**: introduce `GooseRequest` and `GooseRequestBuilder` for more flexibility when making requests
    o remove `GooseUser::post_named`, `GooseUser::head_named`, `GooseUser::delete_named`, `GooseUser::goose_get`, `GooseUser::goose_put`, `GooseUser::goose_head`, `GooseUser::goose_put`, `GooseUser::goose_patch`, `GooseUser::goose_delete`, and `GooseUser::goose_send`
    o adds or modifies helpers `GooseUser::get`, `GooseUser::get_named`, `GooseUser::post`, `GooseUser::post_form`, `GooseUser::post_json`, `GooseUser::head`, and `GooseUser::delete`
    o replaces `GooseUser::goose_send` with `GooseUser::request` which accepts a `GooseRequest` object
    o fixes [#370] (see `GooseRequestBuilder::expect_status_code`)

### DIFF
--- a/examples/simple_with_session.rs
+++ b/examples/simple_with_session.rs
@@ -77,7 +77,7 @@ async fn authenticated_index(user: &mut GooseUser) -> GooseTaskResult {
     let session = user.get_session_data_unchecked::<Session>();
     let goose_request = GooseRequest::builder()
         .request_builder(
-            user.request_builder(&GooseMethod::Get, "/")?
+            user.get_request_builder(&GooseMethod::Get, "/")?
                 .bearer_auth(&session.jwt_token),
         )
         .build();
--- a/examples/simple_with_session.rs
+++ b/examples/simple_with_session.rs
@@ -55,11 +55,9 @@ async fn main() -> Result<(), GooseError> {
 /// on_start task when registering it above. This means it only runs one time
 /// per user, when the user thread first starts.
 async fn website_signup(user: &mut GooseUser) -> GooseTaskResult {
-    let request_builder = user.goose_post("/signup")?;
-    // https://docs.rs/reqwest/*/reqwest/blocking/struct.RequestBuilder.html#method.form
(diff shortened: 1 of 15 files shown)


examples/simple_with_session.rs
```diff
@@ -77,8 +75,14 @@ async fn authenticated_index(user: &mut GooseUser) -> GooseTaskResult {
     // This will panic if the session is missing or if the session is not of the right type
     // use `get_session_data` to handle missing session
     let session = user.get_session_data_unchecked::<Session>();
-    let request = user.goose_get("/")?.bearer_auth(&session.jwt_token);
-    let _goose = user.goose_send(request, None).await?;
+    let goose_request = GooseRequest::builder()
+        .request_builder(
+            user.get_request_builder(&GooseMethod::Get, "/")?
+                .bearer_auth(&session.jwt_token),
```
LionsAd, line 81: I would just use a helper var, so that it looks like this:

```
let reqwest_request_builder =  user.get_request_builder(&GooseMethod::Get, "/")?
                .bearer_auth(&session.jwt_token),

let goose_request = GooseRequest::builder()
        .set_request_builder(reqwest_request_builder)
        .build();
```

However for this common example I would just add bearer_auth as a function to the GooseRequest builder and pass it through to Reqwest.

Easy.

jeremyandrews: As both are built at different times for maximum flexibility, there's no clear way to accept it in the `GooseRequestBuilder` and pass it to the `RequestBuilder` -- in some workflows that could work, but not in all.

jeremyandrews: I've split it out as you suggested, which I agree makes it easier to understand.

LionsAd (approved):
RTBM

PR #381, opened 2021-11-08
Add requests per second graph to the HTML report.

### LINKED ISSUE #246: add charts to HTML report
The `--report-file` option already generates an HTML report with the same information found on the CLI when running Goose. It should be enhanced to also include useful charts to visualize the load test.

A continuation of #232

slashrsm:
Closes #246.

## TODO

- ~Graph color scheme~
- ~Only display starting/stopping phase when `--no-reset-metrics` is used~
- Investigate spike on transition between starting and started phase
- Confirm that metrics are correct (when throttling is enabled RPS numbers do not match exactly)
- Automated tests
- Add additional graphs. 

### With `--no-reset-metrics`
Screenshot 2021-11-18 at 20 53 10

### Without `--no-reset-metrics`
Screenshot 2021-11-18 at 20 49 50

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -68,8 +68,6 @@ const DEFAULT_PORT: &str = "5115";
 /// --debug-format FORMAT      Sets debug log format (csv, json, raw, pretty)
 /// --no-debug-body            Do not include the response body in the debug log
 /// --status-codes             Tracks additional status code metrics
-/// --requests-per-second      Tracks additional requests per second metrics
-/// --rps-bucket               Size (in seconds) of the time-based requests per second bucket (default: 10)
 ///
 /// Advanced:
 /// --no-telnet                Doesn't enable telnet Controller
@@ -190,18 +188,8 @@ pub struct GooseConfiguration {
     #[options(no_short)]
     pub no_debug_body: bool,
     /// Tracks additional status code metrics
-    #[options(no_short, help = "Tracks additional status code metrics")]
+    #[options(no_short, help = "Tracks additional status code metrics\n\nAdvanced:")]
     pub status_codes: bool,
-    /// Tracks additional request per second metrics
-    #[options(no_short, help = "Tracks additional request per second metrics")]
-    pub requests_per_second: bool,
-    // Size (in seconds) of the time-based requests per second bucket (default: 10)
-    // Add a blank line and then an Advanced: header after this option
-    #[options(
-        no_short,
-        help = "Size of the time bucket in time-based requests per second calculations\n\nAdvanced:"
-    )]
-    pub rps_bucket: usize,
     /// Doesn't enable telnet Controller
     #[options(no_short)]
     pub no_telnet: bool,
@@ -335,10 +323,6 @@ pub(crate) struct GooseDefaults {
     pub co_mitigation: Option<GooseCoordinatedOmissionMitigation>,
     /// An optional default to track additional status code metrics.
     pub status_codes: Option<bool>,
-    /// An optional default to track requests per second metrics.
-    pub requests_per_second: Option<bool>,
-    // An optional default size of the requests per second bucket size.
-    pub rps_bucket: Option<usize>,
     /// An optional default maximum requests per second.
     pub throttle_requests: Option<usize>,
     /// An optional default to follows base_url redirect with subsequent request.
@@ -1539,42 +1523,6 @@ impl GooseConfiguration {
                 },
             ])
             .unwrap_or(false);
-
-        // Configure `requests_per_second`.
-        self.requests_per_second = self
-            .get_value(vec![
-                // Use --requests-per-second if set.
-                GooseValue {
-                    value: Some(self.requests_per_second),
-                    filter: !self.requests_per_second,
-                    message: "requests_per_second",
-                },
-                // Otherwise use GooseDefault if set.
-                GooseValue {
-                    value: defaults.requests_per_second,
-                    filter: defaults.requests_per_second.is_none() || self.worker,
-                    message: "requests_per_second",
-                },
-            ])
-            .unwrap_or(false);
-
-        // Configure `rps_bucket`.
-        self.rps_bucket = self
-            .get_value(vec![
-                // Use --requests-per-second-bucket if set.
-                GooseValue {
-                    value: Some(self.rps_bucket),
-                    filter: self.rps_bucket == 0,
-                    message: "rps_bucket",
-                },
-                // Otherwise use GooseDefault if set.
-                GooseValue {
-                    value: defaults.rps_bucket,
-                    filter: defaults.rps_bucket.is_none() || self.worker,
-                    message: "rps_bucket",
-                },
-            ])
-            .unwrap_or(10);
 
         // Configure `no_telnet`.
         self.no_telnet = self
--- a/src/config.rs
+++ b/src/config.rs
@@ -188,8 +188,10 @@ pub struct GooseConfiguration {
     #[options(no_short)]
     pub no_debug_body: bool,
     /// Tracks additional status code metrics
+    // Add a blank line and then an Advanced: header after this option
     #[options(no_short, help = "Tracks additional status code metrics\n\nAdvanced:")]
     pub status_codes: bool,
+
     /// Doesn't enable telnet Controller
     #[options(no_short)]
     pub no_telnet: bool,
--- a/src/config.rs
+++ b/src/config.rs
@@ -68,6 +68,8 @@ const DEFAULT_PORT: &str = "5115";
 /// --debug-format FORMAT      Sets debug log format (csv, json, raw, pretty)
 /// --no-debug-body            Do not include the response body in the debug log
 /// --status-codes             Tracks additional status code metrics
+/// --requests-per-second      Tracks additional requests per second metrics
+/// --rps-bucket               Size (in seconds) of the time-based requests per second bucket (default: 10)
 ///
 /// Advanced:
 /// --no-telnet                Doesn't enable telnet Controller
@@ -188,10 +190,18 @@ pub struct GooseConfiguration {
     #[options(no_short)]
     pub no_debug_body: bool,
     /// Tracks additional status code metrics
-    // Add a blank line and then an Advanced: header after this option
-    #[options(no_short, help = "Tracks additional status code metrics\n\nAdvanced:")]
+    #[options(no_short, help = "Tracks additional status code metrics")]
     pub status_codes: bool,
-
+    /// Tracks additional request per second metrics
+    #[options(no_short, help = "Tracks additional request per second metrics")]
+    pub requests_per_second: bool,
+    // Size (in seconds) of the time-based requests per second bucket (default: 10)
+    // Add a blank line and then an Advanced: header after this option
+    #[options(
+        no_short,
+        help = "Size of the time bucket in time-based requests per second calculations\n\nAdvanced:"
+    )]
+    pub rps_bucket: usize,
     /// Doesn't enable telnet Controller
     #[options(no_short)]
     pub no_telnet: bool,
@@ -325,6 +335,10 @@ pub(crate) struct GooseDefaults {
     pub co_mitigation: Option<GooseCoordinatedOmissionMitigation>,
     /// An optional default to track additional status code metrics.
     pub status_codes: Option<bool>,
+    /// An optional default to track requests per second metrics.
+    pub requests_per_second: Option<bool>,
+    // An optional default size of the requests per second bucket size.
+    pub rps_bucket: Option<usize>,
     /// An optional default maximum requests per second.
     pub throttle_requests: Option<usize>,
     /// An optional default to follows base_url redirect with subsequent request.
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -68,6 +68,8 @@ const DEFAULT_PORT: &str = "5115";
 /// --debug-format FORMAT      Sets debug log format (csv, json, raw, pretty)
 /// --no-debug-body            Do not include the response body in the debug log
 /// --status-codes             Tracks additional status code metrics
+/// --requests-per-second      Tracks additional requests per second metrics
+/// --rps-bucket               Size (in seconds) of the time-based requests per second bucket (default: 10)
```
slashrsm, line 72: Any ideas for better names?

src/metrics.rs
```diff
@@ -451,6 +458,26 @@ impl GooseRequestMetricAggregate {
         self.status_code_counts.insert(status_code, counter);
         debug!("incremented {} counter: {}", status_code, counter);
     }
+
+    /// Record request in requests per second metric.
+    pub(crate) fn record_requests_per_second(&mut self, bucket_size: usize) {
+        let time_bucket = (Utc::now().timestamp() - self.start_time) as usize / bucket_size;
```
slashrsm, line 464: Instead of the actual time of the request we are using the time when the request metric is reported. 

It would be better to collect the time when the request happened, but I think that this data currently isn't reported by the worker.

src/report.rs
```diff
@@ -404,6 +406,46 @@ pub fn error_row(error: &metrics::GooseErrorMetricAggregate) -> String {
...
+            map(|bucket| format!("{}s - {}s", (bucket * bucket_size), ((bucket + 1) * bucket_size)))
+            .collect::<Vec<_>>()
+    );
+
+    format!(
+        r#"<div class="graph-rps">
+        <h2>Requests per second</h2>
+            <div id="main" style="width: 1000px; height:660px; background: white;"></div>
+
+            <script src="https://cdn.jsdelivr.net/npm/echarts@5.2.2/dist/echarts.min.js"></script>
```
slashrsm, line 422: Is it OK to load echarts from the CDN?Is it OK to load echarts from the CDN?

src/metrics.rs
```diff
@@ -2951,6 +2987,33 @@ impl GooseAttack {
                 errors_template = "".to_string();
             }
 
+            let graph_rps_template: String;
+            if self.configuration.requests_per_second {
+                let mut aggregated_count: HashMap<usize, usize> = HashMap::new();
```
jeremyandrews, line 2992: I wonder if instead we could use the logger thread to efficiently write time-data to log files (I assume we'd need multiple: one for each type of chart we add to the reports). So instead of buckets, in the main parent loop we'd track the total # of requests, total # of tasks, total # of users seen each second and push that to the logger thread. Then, when we write the reports we could perhaps read in the time-data from these files instead of trying to hold them in memory? In this way we always have access to per-second granularity, but we can decide how often to plot a point (every second, every 10 seconds, every minute, etc) depending on how long the load test actually ran, something we can't know at start-time and run-time), and we can even add controls allow us to zoom in/out on the data.

jeremyandrews (changes requested):
It's fantastic seeing this feature come together! 

As we only use the requests-per-second (and tasks-per-second and users-per-second etc) data when writing the html-report, perhaps we simply tie it to the existing `--report-file` option.

Specifying a bucket size doesn't seem to scale well, we need a mechanism that works regardless of if the load test runs for 5 minutes or 5 hours.

The code is clean and when it works it looks great -- very nice addition to the load test report!

src/report.rs
LionsAd: That's fine to start - we could make this configurable in the future - if you wanted to host your own.

src/metrics.rs
```diff
@@ -9,6 +9,7 @@
 //! [`GooseErrorMetrics`] are displayed in tables.
 
 use chrono::prelude::*;
+use chrono::Utc;
```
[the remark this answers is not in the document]
slashrsm: Removed.

src/config.rs
```diff
@@ -319,6 +329,10 @@ pub(crate) struct GooseDefaults {
     pub co_mitigation: Option<GooseCoordinatedOmissionMitigation>,
     /// An optional default to track additional status code metrics.
     pub status_codes: Option<bool>,
+    /// An optional default to track requests per second metrics.
+    pub requests_per_second: Option<bool>,
```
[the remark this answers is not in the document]
slashrsm: This is a great point. I removed other options and tied the graph to the report being enabled. 

We can always add more granular control at a later point if it turns out that we need it.

src/config.rs
slashrsm: These are now gone and so is the concept of "bucket".

src/report.rs
slashrsm: Great points. Mentioned the library in the docs.

src/metrics.rs
```diff
@@ -396,6 +397,10 @@ pub struct GooseRequestMetricAggregate {
     /// The hash is primarily used when running a distributed Gaggle, allowing the Manager to confirm
     /// that all Workers are running the same load test plan.
     pub load_test_hash: u64,
+    // Timestamp when the metric collection was started.
+    start_time: i64,
```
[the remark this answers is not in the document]
slashrsm: This is now gone. I switched to `GooseMetrics.duration`.

src/config.rs
```diff
@@ -1495,6 +1509,42 @@ impl GooseConfiguration {
...
+                    message: "rps_bucket",
+                },
+                // Otherwise use GooseDefault if set.
+                GooseValue {
+                    value: defaults.rps_bucket,
+                    filter: defaults.rps_bucket.is_none() || self.worker,
+                    message: "rps_bucket",
+                },
+            ])
+            .unwrap_or(10);
```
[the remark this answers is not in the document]
slashrsm: This is now gone.

src/metrics.rs
slashrsm: I changed the approach and now it stores # of requests for each second as you described here. 

I calculated how much data that approach uses in memory and I got about ~350kB per day (4 bytes/second * 86400 = 345600 bytes). Considering how little memory this uses I am not sure if it is worth adding the complexity with storing data to files. Thoughts?

slashrsm:
I removed the options that were added in the initial version of this pull request and now we rely on `--report-file` to enable/disable this.

I also changed how the data is stored. Now we are storing the number of requests for each second and the (flawed) concept of "buckets" is now gone. Let me know what you think. 

Also, what do you think about using something like this for the chart style: https://echarts.apache.org/examples/en/editor.html?c=area-simple. I imagine that it could be quite useful specially for long-running load tests (ability to zoom in, etc.).

src/config.rs
```diff
@@ -187,10 +187,8 @@ pub struct GooseConfiguration {
     #[options(no_short)]
     pub no_debug_body: bool,
     /// Tracks additional status code metrics
-    // Add a blank line and then an Advanced: header after this option
```
jeremyandrews, line 190: Remove this and the next unnecessary change in this file.

src/metrics.rs
```diff
@@ -2415,6 +2434,9 @@ impl GooseAttack {
             if self.configuration.status_codes {
                 merge_request.set_status_code(request_metric.status_code);
             }
+            if !self.configuration.report_file.is_empty() {
+                merge_request.record_requests_per_second(self.metrics.duration);
```
jeremyandrews, line 2438: We already have `self`, so no need to pass `self.metrics.duration`

jeremyandrews: The graph should include ramp-up time and ramp-down time -- in the following test I ran with the following parameters:

```bash
cargo run --release --example drupal_memcache -- --host http://apache/ -v -u5 -t10 --report-file report.html --throttle-requests 11
```

It's not including the ramp-up time, and it needs to. Also, I think we're missing the first second, as we should see 10 seconds of requests for ~11 rps:

image

jeremyandrews (changes requested):
This is much cleaner, thanks for reworking! 

The graph should include ramp-up, running, and ramp-down (and ideally should visually indicate each somehow). I also like the idea of a different graph type that allows you to zoom in.

src/metrics.rs
jeremyandrews: Completely agreed, there's no reason to add this sort of complexity for such small data sizes -- even if we add many charts.

jeremyandrews:
> I also changed how the data is stored. Now we are storing the number of requests for each second and the (flawed) concept of "buckets" is now gone. Let me know what you think.

This is much simpler, I like it!

Note that the data is still getting cut off: I ran a load test for 1 hour (3600 seconds) but the data was cut off -- I manually added the missing bit of the report to see the data that was recorded. Likely this is a limitation of the current `--report-file` implementation, and a buffer or something will need to be increase to support larger file sizes.

image

> Also, what do you think about using something like this for the chart style: https://echarts.apache.org/examples/en/editor.html?c=area-simple. I imagine that it could be quite useful specially for long-running load tests (ability to zoom in, etc.).

I like that idea very much, it would add richness to the charts being able to zoom in.

Perhaps the X axis should be actual timestamps though, instead of seconds since the test started? I wonder which is more useful? (The latter would be easier to match up against data in New Relic etc)

slashrsm:
I am still trying to figure out the cutting off problem (I am able to reproduce it). I fell that there is something wrong with JSON serialization or maybe `format!` when the data series grows over certain point. Looking into it as we speak, but is quite time consuming, since I can only reproduce it if I run a load test for a while.

src/report.rs
```diff
@@ -404,6 +406,48 @@ pub fn error_row(error: &metrics::GooseErrorMetricAggregate) -> String {
...
+                    xAxis: {{
+                        type: 'category',
+                        data: {buckets}
+                    }},
+                    yAxis: {{
+                        type: 'value'
+                    }},
+                    series: [
+                        {{
+                            data: {values},
```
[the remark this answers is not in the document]
slashrsm: This was caused by using `write` instead of `write_all`. My runs are producing valid HTML now.

src/config.rs
slashrsm: Done.

src/metrics.rs
slashrsm: > We already have self, so no need to pass self.metrics.duration

`record_requests_per_second()` is a function on a different struct so `self` represents  something else there. And `duration` is not available there, so we need to send it.

slashrsm:
I improved the appearance of the graph:

1. Improved styling of the line
2. Added zoom support
3. Added tooltip on mouse hover
4. Marked starting and stopping period (I can't make the latter to appear as it completes so quickly)
5. Added "Save as image" function.

I noticed that there is always a spike of requests exactly at the point when the load test was started up and `--no-reset-metrics`. Is it possible that we are recording some requests twice in this case?

## Without `--no-reset-metrics`
Screenshot 2021-11-11 at 20-32-39 Goose Attack Report

## With `--no-reset-metrics`
Screenshot 2021-11-11 at 20-38-36 Goose Attack Report

jeremyandrews:
I think your graphs in the PR are labeled backwards -- the first one looks like `--no-reset-metrics` IS enabled, the second looks like `--no-reset-metrics` is NOT enabled.

jeremyandrews:
Any idea what this spike is at the start of the load test? It seems to show up in the graphs semi-regularly (in this case I'm using the following options: `-v --report-file report.html -u10 -s10 -t10 --no-reset-metrics`:

image

jeremyandrews:
Any idea what this spike is at the start of the load test? It seems to show up in the graphs semi-regularly (in this case I'm using the following options: `-v --report-file report.html -u10 -s10 -t10 --no-reset-metrics`:

image

src/metrics.rs
```diff
@@ -3004,6 +3030,44 @@ impl GooseAttack {
...
             }
 
+            let graph_rps_template: String;
+            if Some(starting) == self.metrics.starting
+                && Some(started) == self.metrics.started
+                && Some(stopping) == self.metrics.stopping
+                && Some(stopped) == self.metrics.stopped
+            {
+                let mut count = 0;
+                for (_path, path_metric) in self.metrics.requests.iter() {
```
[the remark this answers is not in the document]
jeremyandrews: If `--report-file` is enabled and `--no-reset-metrics` is not enabled, we should store a copy of the `metrics.requests` "somewhere else" before we reset the metrics so we can generate a complete chart. Alternatively, in this configuration perhaps we should simply not display the "Starting" portion of the chart as it will always be blank.

jeremyandrews (changes requested):
This is looking very nice! I like the new charts much better.
 - When we are displaying starting/stopping, perhaps we could use "gray" instead of a red shade, as the latter draws more attention to it than it likely should
 - It would be ideal to create an automated test that confirms the data in the report is truly accurate
 - We either need to preserve a copy of requests made during Startup when running with "--no-reset-metrics" so we can show all data in the chart, or not show that section of time, as it's confusing/wrong to show it blank
 - I was able to run an hour long load test and got full data, the report file was only 117k

src/metrics.rs
```diff
@@ -2951,6 +2973,25 @@ impl GooseAttack {
                 errors_template = "".to_string();
             }
 
+            let graph_rps_template: String;
+            if !self.configuration.report_file.is_empty() {
+                let mut count = 0;
+                for (_path, path_metric) in self.metrics.requests.iter() {
```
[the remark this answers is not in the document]
slashrsm: I learned something new!

src/metrics.rs
```diff
@@ -451,6 +454,22 @@ impl GooseRequestMetricAggregate {
         self.status_code_counts.insert(status_code, counter);
         debug!("incremented {} counter: {}", status_code, counter);
     }
+
+    /// Record request in requests per second metric.
+    pub(crate) fn record_requests_per_second(&mut self, second: usize) {
+        if self.requests_per_second.len() <= second {
```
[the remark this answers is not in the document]
slashrsm: Done.

slashrsm:
I played with the color scheme and aspect ratio and I think that graph looks better now. Starting phase is only displayed when `--no-reset-metrics` is used and stopping phase is only displayed if it lasts for long enough to actually be visible on the graph. 

## Screenshots

### With `--no-reset-metrics`
Screenshot 2021-11-18 at 20 53 10


### Without `--no-reset-metrics`
Screenshot 2021-11-18 at 20 49 50

jeremyandrews:
This is a very nice enhancement, and only affects the HTML report generated when the load test ends. Committing as is: the rest of your TODO will make great followups -- thanks!

PR #382, opened 2021-11-11
set default request timeout of 60 seconds, configurable

jeremyandrews:
- [#382](https://github.com/tag1consulting/goose/pull/382) set client timeout to 60 seconds by default, used for all requests made; introduce `--timeout VALUE` where VALUE is seconds as integer or a float; timeout can be configured programatically using `GooseDefault::Timeout`
 - introduces `utils::get_float_from_string()`, now used by `utils::get_hatch_rate()`
 - updates `set_client_builder()` documentation to reflect how this changes the default Reqwest Client configuration

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2084,9 +2084,12 @@ impl GooseConfiguration {
             }
         }
 
-        // If set, timeout must be non-zero.
+        // If set, timeout must be greater than zero.
         if let Some(timeout) = self.timeout.as_ref() {
-            if timeout == "0" {
+            if crate::util::get_float_from_string(self.timeout.clone())
+                .expect("failed to re-convert string to float")
+                <= 0.0
+            {
                 return Err(GooseError::InvalidOption {
                     option: "`configuration.timeout`".to_string(),
                     value: timeout.to_string(),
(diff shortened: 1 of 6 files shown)


src/util.rs
```diff
@@ -355,18 +355,44 @@ pub fn ms_timer_expired(started: time::Instant, elapsed: usize) -> bool {
...
+/// assert_eq!(util::get_float_from_string(Some(".1".to_string())), Some(0.1));
+///
+/// // Valid float string returns a proper float.
+/// assert_eq!(util::get_float_from_string(Some("1.1".to_string())), Some(1.1));
+///
+/// // Invalid number with too many decimals returns None.
+/// assert_eq!(util::get_float_from_string(Some("1.1.1".to_string())), None);
+///
+/// // No number returns None.
+/// assert_eq!(util::get_float_from_string(None), None);
```
slashrsm, line 384: I love doc tests!

slashrsm (commented):
Looks good to my (n00b) me.

src/config.rs
```diff
@@ -2046,6 +2084,17 @@ impl GooseConfiguration {
             }
         }
 
+        // If set, timeout must be non-zero.
+        if let Some(timeout) = self.timeout.as_ref() {
+            if timeout == "0" {
```
LionsAd, line 2089: Should this not already do a test conversion to check if eg “abc” is passed?

jeremyandrews: It’s validate le already to be an unsigned integer.  Now we validate it’s not 0.

jeremyandrews: actually: I'm wrong -- we were validating in an earlier version but it's currently broken -- I'll fix.

jeremyandrews: fixed so it's properly validating the timeout is > 0.0

LionsAd (approved):
RTBM

PR #392, opened 2021-11-21
Add automated tests for requests per second graph in HTML report

slashrsm:
This is a follow-up of #381 and is part of #246.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -462,7 +462,7 @@ impl GooseRequestMetricAggregate {
         // test will at the beginning we need to push new elements (second
         // counters) as the test is running.
         if self.requests_per_second.len() <= second {
-            for _ in 0..=(second - self.requests_per_second.len() + 1) {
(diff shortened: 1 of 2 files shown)


slashrsm:
In order to reliably test [this hunk](https://github.com/tag1consulting/goose/pull/381/files#diff-9f90e93b5696724fd8e4407c9db0bfd381d5468bcfbf2f09c17ac8510a85e104R2441) I would need to control the return value of `Utc::now()`. There is a [pull request that allows that](https://github.com/chronotope/chrono/pull/580), but it is still in progress. 

Until that is in we could add something on `GooseAttack` that would wrap `Utc::now()` and let us override it in tests. Not ideal since we would be polluting code with test-only logic. @jeremyandrews @LionsAd Any ideas/thoughts?

jeremyandrews:
@slashrsm There's plenty of places in the code not currently tested. Perhaps we add a comment w/ a @TODO linking to the PR in question as a reminder to add a test when it's available?

Issue #365, opened 2021-09-16
Specify the traffic profile/shape?


PsypherPunk:
I wanted to check: is there a way to define the "profile" of a load test?

To clarify, we're currently using Gatling, wherein you can do something like:

```scala
  setUp(scn.inject(constantUsersPerSec(500) during (1 seconds))).throttle(
    reachRps(500) in (1 seconds),
    holdFor(4),
    reachRps(10) in (1 seconds),
  ).protocols(httpConf)
```

…to have a specific traffic pattern, specifying the rate of change, for how long to hold at that rate, etc. Specifically, we need to simulate varying rates with sudden changes.

It looks like, for instance, Locust have implemented this as ["custom shapes"](https://docs.locust.io/en/stable/custom-load-shape.html).

Is there a way to achieve this in Goose?

Many thanks.

jeremyandrews:
@PsypherPunk Unfortunately this is not yet possible with Goose. Tagging it as feature request.

PR #394, opened 2021-11-24
Add additional graphs to the HTML report

### LINKED ISSUE #246: add charts to HTML report

slashrsm:
Closes #246.

## Graphs

- Errors per second
- Average response time per second
- Active users per second
- Active tasks per second

### Errors

Screenshot 2021-11-24 at 16-07-12 Goose Attack Report

### Response time

Screenshot 2021-11-24 at 19 01 35

### Active users

Screenshot 2021-11-24 at 19 25 52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.15.2-dev
  - [#391](https://github.com/tag1consulting/goose/pull/391) properly sleep for configured `set_wait_time()` walking regularly to exit quickly if the load test ends
+ - [#394](https://github.com/tag1consulting/goose/pull/394) add additional graphs to the HTML report: errors per second,
+ average response time, active users, active tasks
 
 ## 0.15.1 November 19, 2021
  - [#374](https://github.com/tag1consulting/goose/pull/374) renamed `simple-with-session.rs` to `session.rs` and `simple-closure.rs` to `closure.rs` to avoid confusion with the `simple.rs` example as they all do different things
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## 0.15.2-dev
  - [#391](https://github.com/tag1consulting/goose/pull/391) properly sleep for configured `set_wait_time()` walking regularly to exit quickly if the load test ends
- - [#394](https://github.com/tag1consulting/goose/pull/394) add additional graphs to the HTML report: errors per second,
- average response time, active users, active tasks
+ - [#394](https://github.com/tag1consulting/goose/pull/394) add additional graphs to the HTML report: errors per second, average response time, active users, active tasks
 
 ## 0.15.1 November 19, 2021
  - [#374](https://github.com/tag1consulting/goose/pull/374) renamed `simple-with-session.rs` to `session.rs` and `simple-closure.rs` to `closure.rs` to avoid confusion with the `simple.rs` example as they all do different things
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -287,6 +287,7 @@
 //! See the License for the specific language governing permissions and
 //! limitations under the License.
 
+use chrono::Utc;
 use downcast_rs::{impl_downcast, Downcast};
 use http::method::Method;
 use reqwest::{header, Client, ClientBuilder, RequestBuilder, Response};
@@ -1510,6 +1511,7 @@ impl GooseUser {
 
         // Once past the throttle, the request is officially started.
         let started = Instant::now();
+        let request_time = Utc::now();
 
         // Create a Reqwest Request object from the RequestBuilder.
         let built_request = request_builder.build()?;
@@ -1558,6 +1560,7 @@ impl GooseUser {
             request_name,
             self.started.elapsed().as_millis(),
             self.weighted_users_index,
+            request_time,
         );
 
         // Make the actual request.
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -287,7 +287,6 @@
 //! See the License for the specific language governing permissions and
 //! limitations under the License.
 
-use chrono::Utc;
 use downcast_rs::{impl_downcast, Downcast};
 use http::method::Method;
 use reqwest::{header, Client, ClientBuilder, RequestBuilder, Response};
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -1511,7 +1511,6 @@ impl GooseUser {
 
         // Once past the throttle, the request is officially started.
         let started = Instant::now();
-        let request_time = Utc::now();
 
         // Create a Reqwest Request object from the RequestBuilder.
         let built_request = request_builder.build()?;
@@ -1560,7 +1559,6 @@ impl GooseUser {
             request_name,
             self.started.elapsed().as_millis(),
             self.weighted_users_index,
-            request_time,
         );
 
         // Make the actual request.
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -511,7 +511,7 @@ impl GooseRequestMetricAggregate {
 /// better to do it as we go to save memory.
 fn expand_per_second_metric_array<T: Clone>(data: &mut Vec<T>, second: usize, initial: T) {
     // Each element in per second metric vectors (self.requests_per_second,
-    // self.errors_per_second, ...) is count for a given second since the start
+    // self.errors_per_second, ...) is counted for a given second since the start
     // of the test. Since we don't know how long the test will at the beginning
     // we need to push new elements (second counters) as the test is running.
     if data.len() <= second {
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -25,9 +25,7 @@ use crate::config::GooseDefaults;
 use crate::goose::{get_base_url, GooseMethod, GooseTaskSet};
 use crate::logger::GooseLog;
 use crate::report;
-use crate::util::{
-    median, ms_timer_expired, standard_deviation, timer_expired, truncate_string, MovingAverage,
-};
+use crate::util;
 #[cfg(feature = "gaggle")]
 use crate::worker::{self, GaggleMetrics};
 use crate::{AttackMode, GooseAttack, GooseAttackRunState, GooseConfiguration, GooseError};
@@ -403,7 +401,7 @@ pub struct GooseRequestMetricAggregate {
     /// Counts errors per second. Each element of the vector represents one second.
     pub errors_per_second: Vec<u32>,
     /// Maintains average response time per second. Each element of the vector represents one second.
-    pub average_response_time_per_second: Vec<MovingAverage>,
+    pub average_response_time_per_second: Vec<util::MovingAverage>,
 }
 impl GooseRequestMetricAggregate {
     /// Create a new GooseRequestMetricAggregate object.
@@ -494,7 +492,7 @@ impl GooseRequestMetricAggregate {
         expand_per_second_metric_array(
             &mut self.average_response_time_per_second,
             second,
-            MovingAverage::new(),
+            util::MovingAverage::new(),
         );
         self.average_response_time_per_second[second].add_item(response_time as f32);
 
@@ -1122,7 +1120,7 @@ impl GooseMetrics {
                 writeln!(
                     fmt,
                     " {:<24} | {:>13} | {:>14} | {:>8.reqs_p$} | {:>7.fails_p$}",
-                    truncate_string(request_key, 24),
+                    util::truncate_string(request_key, 24),
                     total_count.to_formatted_string(&Locale::en),
                     format!(
                         "{} ({}%)",
@@ -1138,7 +1136,7 @@ impl GooseMetrics {
                 writeln!(
                     fmt,
                     " {:<24} | {:>13} | {:>14} | {:>8.reqs_p$} | {:>7.fails_p$}",
-                    truncate_string(request_key, 24),
+                    util::truncate_string(request_key, 24),
                     total_count.to_formatted_string(&Locale::en),
                     format!(
                         "{} ({:.1}%)",
@@ -1254,7 +1252,7 @@ impl GooseMetrics {
                     writeln!(
                         fmt,
                         " {:24 } |",
-                        truncate_string(
+                        util::truncate_string(
                             &format!("{}: {}", task.taskset_index + 1, &task.taskset_name),
                             60
                         ),
@@ -1266,7 +1264,7 @@ impl GooseMetrics {
                     writeln!(
                         fmt,
                         " {:<24} | {:>13} | {:>14} | {:>8.runs_p$} | {:>7.fails_p$}",
-                        truncate_string(
+                        util::truncate_string(
                             &format!("  {}: {}", task.task_index + 1, task.task_name),
                             24
                         ),
@@ -1285,7 +1283,7 @@ impl GooseMetrics {
                     writeln!(
                         fmt,
                         " {:<24} | {:>13} | {:>14} | {:>8.runs_p$} | {:>7.fails_p$}",
-                        truncate_string(
+                        util::truncate_string(
                             &format!("  {}: {}", task.task_index + 1, task.task_name),
                             24
                         ),
@@ -1397,7 +1395,7 @@ impl GooseMetrics {
                     writeln!(
                         fmt,
                         " {:24 } |",
-                        truncate_string(
+                        util::truncate_string(
                             &format!("{}: {}", task.taskset_index + 1, &task.taskset_name),
                             60
                         ),
@@ -1429,14 +1427,14 @@ impl GooseMetrics {
                 writeln!(
                     fmt,
                     " {:<24} | {:>11.avg_precision$} | {:>10} | {:>11} | {:>10}",
-                    truncate_string(
+                    util::truncate_string(
                         &format!("  {}: {}", task.task_index + 1, task.task_name),
                         24
                     ),
                     average,
                     format_number(task.min_time),
                     format_number(task.max_time),
-                    format_number(median(
+                    format_number(util::median(
                         &task.times,
                         task.counter,
                         task.min_time,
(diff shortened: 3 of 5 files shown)


jeremyandrews (changes requested):
This is wonderful to see! I'm looking forward to getting this merged once my comments are addressed. There also seems to be a bug in the generation of the graph during startup, when startup is >=20 seconds (more or less):
image

I also would prefer to see the graphs mixed in with the relevant data, and graph first data next. So, show the requests per second graph, then show the requests break down, etc.

src/metrics.rs
```diff
@@ -3090,8 +3139,102 @@ impl GooseAttack {
...
+            for path_metric in self.metrics.requests.values() {
+                count = max(count, path_metric.requests_per_second.len());
+            }
+            for path_metric in self.metrics.requests.values() {
+                count = max(count, path_metric.errors_per_second.len());
+            }
+            for path_metric in self.metrics.requests.values() {
+                count = max(count, path_metric.average_response_time_per_second.len());
+            }
+            count = max(count, self.metrics.users_per_second.len());
```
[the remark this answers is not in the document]
slashrsm: Renamed it to `total_graph_seconds`. Will that work?

src/metrics.rs
```diff
@@ -398,6 +398,10 @@ pub struct GooseRequestMetricAggregate {
     pub load_test_hash: u64,
     // Counts requests per second. Each element of the vector represents one second.
     pub requests_per_second: Vec<u32>,
+    // Counts errors per second. Each element of the vector represents one second.
+    pub errors_per_second: Vec<u32>,
+    // Maintains average response time per second. Each element of the vector represents one second.
+    pub average_response_time_per_second: Vec<(u32, f32)>,
```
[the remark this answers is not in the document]
slashrsm: `response_time` is u64 and we want a floating point number here since we are calculating an average. We could use f64, but I don't think that we need as much precision and we would still need to cast from u64 to f64?

src/metrics.rs
```diff
@@ -867,6 +915,9 @@ pub struct GooseMetrics {
     /// This value may be smaller than what was configured at start time if the test
     /// didn't run long enough for all configured users to start.
     pub users: usize,
+    /// Number of users at the end of each second of the test. Each element of the vector
+    /// represents one second.
+    pub users_per_second: Vec<u32>,
```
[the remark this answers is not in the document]
slashrsm: Done. Well... the reason was that some of the functions expected `u32` and `Vec<u32>` and I was trying to avoid generic types. Now you forced me to bite into that apple :)

src/metrics.rs
```diff
@@ -2440,9 +2506,18 @@ impl GooseAttack {
...
+                    let second_since_start =
+                        (Utc::now().timestamp() - starting.timestamp()) as usize;
+
+                    merge_request.record_requests_per_second(second_since_start);
+                    merge_request.record_average_response_time_per_second(
+                        second_since_start,
+                        request_metric.response_time,
                     );
+
+                    if !request_metric.success {
```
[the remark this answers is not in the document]
slashrsm: > Is there ever a time success is false but the error text is empty?

I have no idea :) I am not  familiar enough with Goose to be able to answer that. 

With that said - I do think that using a bool instead of a string in a condition makes more sense.

src/metrics.rs
```diff
@@ -2467,6 +2542,12 @@ impl GooseAttack {
         let mut received_message = false;
         let mut message = goose_attack_run_state.metrics_rx.try_recv();
 
+        let mut tasks = 0;
+        for set in self.task_sets.iter() {
+            tasks += set.tasks.len();
+        }
+        self.metrics.record_users_tasks_per_second(tasks as u32);
```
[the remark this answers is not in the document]
slashrsm: > Why do we calculate this before merging in the received statistics? Shouldn't we merge statistics and then count?

I don't think this would change anything. This is simply counting tasks and users at this point and I don't think other stats would influence that. 

> Also, why does this happen here, instead of in the same place as where we record the other metrics for the graphs?

Because number of tasks do not need to be recorded for every request that we are collecting stats for; we need to do it just once.

src/metrics.rs
```diff
@@ -455,26 +461,67 @@ impl GooseRequestMetricAggregate {
...
+            / self.average_response_time_per_second[second].0 as f32;
+
+        debug!(
+            "updated second {} for average response time per second: {}",
+            second, self.average_response_time_per_second[second].1
+        );
+    }
+}
+
+fn expand_per_second_metric_array<T: Clone>(data: &mut Vec<T>, second: usize, initial: T) {
```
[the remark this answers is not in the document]
slashrsm: Done.

src/metrics.rs
```diff
@@ -455,26 +461,67 @@ impl GooseRequestMetricAggregate {
...
+    // we need to push new elements (second counters) as the test is running.
+    if data.len() <= second {
+        for _ in 0..(second - data.len() + 1) {
+            data.push(initial.clone());
+        }
+    };
 }
+
+/// Implement equality for GooseRequestMetricAggregate. We can't simply derive since
+/// we have floats in the struct.
```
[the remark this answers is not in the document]
slashrsm: Done.

src/metrics.rs
```diff
@@ -2159,6 +2213,18 @@ impl GooseMetrics {
 
         Ok(())
     }
+
+    pub(crate) fn record_users_tasks_per_second(&mut self, tasks: u32) {
```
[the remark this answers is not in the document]
slashrsm: Done.

src/metrics.rs
```diff
@@ -455,26 +461,67 @@ impl GooseRequestMetricAggregate {
...
+        &mut self,
+        second: usize,
+        response_time: u64,
+    ) {
+        expand_per_second_metric_array(
+            &mut self.average_response_time_per_second,
+            second,
+            (0, 0.0),
+        );
+        self.average_response_time_per_second[second].0 += 1;
```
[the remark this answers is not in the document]
slashrsm: Done. Or should I create a struct for this data to make it more self-descriptive?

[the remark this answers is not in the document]
jeremyandrews: A struct would likely be even better / more readable.

src/metrics.rs
```diff
@@ -455,26 +464,81 @@ impl GooseRequestMetricAggregate {
...
+            data.push(initial.clone());
+        }
+    };
 }
+
+/// Implement equality for GooseRequestMetricAggregate. We can't simply derive since
+/// we have floats in the struct.
+///
+/// `Eq` trait has no functions on it - it is there only to inform the compiler
+/// that this is and equivalence rather than a partial equivalence.
```
jeremyandrews, line 537: typo: `this is an equivalence` not `this is and...`

src/metrics.rs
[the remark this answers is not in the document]
jeremyandrews: Reviewing the code in question, I see it's done that way as we store the actual text of the error for a final report: but that's only in the latter part of that function, so likely we should start by checking the boolean as you've done. In other words: what you're doing is correct, thanks.

src/metrics.rs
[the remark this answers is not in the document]
jeremyandrews: This still feels like an illogical place to perform this function, especially if we don't actually care whether or not we receive the metrics (the purpose of this function).

src/metrics.rs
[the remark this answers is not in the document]
slashrsm: 👍 I updated that condition too.

src/metrics.rs
slashrsm: Done.

src/metrics.rs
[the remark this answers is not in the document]
slashrsm: Agreed. Moved it to `GooseAttack::sync_metrics()`. Will that work?

src/metrics.rs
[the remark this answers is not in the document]
slashrsm: Added `util::MovingAverage`

src/metrics.rs
```diff
@@ -879,6 +930,9 @@ pub struct GooseMetrics {
     /// [GooseDefault::NoTaskMetrics](../config/enum.GooseDefault.html#variant.NoTaskMetrics) or
     /// [GooseDefault::NoMetrics](../config/enum.GooseDefault.html#variant.NoMetrics).
     pub tasks: GooseTaskMetrics,
+    /// Number of tasks at the end of each second of the test. Each element of the vector
+    /// represents one second.
+    pub tasks_per_second: Vec<u32>,
```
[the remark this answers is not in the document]
slashrsm: Changed to `usize`

src/metrics.rs
```diff
@@ -2578,7 +2578,7 @@ impl GooseAttack {
             match message.unwrap() {
                 GooseMetric::Request(request_metric) => {
                     // If there was an error, store it.
-                    if !request_metric.error.is_empty() {
+                    if !request_metric.success {
```
jeremyandrews, line 2577: This isn't enough -- if we remove this check here, we'll need to add it before we log the error text (as `success` can be `false` but `error` also empty and with this change you've made we could get "empty" errors, a regression to this functionality):
https://github.com/tag1consulting/goose/blob/7121f4617c8a07ebf7183a7cf67b6105bf0f7402/src/metrics.rs#L2562

slashrsm: Will this work?

https://github.com/tag1consulting/goose/pull/394/files#diff-9f90e93b5696724fd8e4407c9db0bfd381d5468bcfbf2f09c17ac8510a85e104L2479-L2480

src/metrics.rs
```diff
@@ -2530,7 +2547,7 @@ impl GooseAttack {
             if !self.configuration.report_file.is_empty() {
                 if let Some(starting) = self.metrics.starting {
                     let second_since_start =
-                        (Utc::now().timestamp() - starting.timestamp()) as usize;
+                        (request_metric.request_time - starting.timestamp() as u64) as usize;
```
slashrsm: Now that we have the actual time of the request on the `GooseRequestMetric` we can use that to calculate the second on the graph that the request belongs to. This gives us more accurate data and as a side-effect it will prevent those gaps.

src/metrics.rs
```diff
@@ -2232,10 +2241,18 @@ impl GooseMetrics {
...
+            };
+            expand_per_second_metric_array(&mut self.users_per_second, second, last_user_count);
             self.users_per_second[second] = self.users;
 
-            expand_per_second_metric_array(&mut self.tasks_per_second, second, 0);
+            let last_task_count = match self.tasks_per_second.last() {
+                Some(last) => *last,
+                None => 0,
+            };
+            expand_per_second_metric_array(&mut self.tasks_per_second, second, last_task_count);
```
slashrsm: Users and tasks metric are not tied to a specific request so we still need to rely on `Utc:now()`. But this is OK since we will hit this immediately after a new user was spawned so the data should be accurate. The only thing we need to make sure is to carry the last count over the seconds we missed while the thread was sleeping to avoid gaps on the graph.

src/metrics.rs
```diff
@@ -293,6 +293,8 @@ impl GooseRawRequest {
 /// knows which request is being updated.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct GooseRequestMetric {
+    // Time when the request was started.
+    pub request_time: u64,
```
slashrsm, line 297: This helps us avoid gaps on the graph, but also (more importantly) gives us more accurate data. 

The only downside I can see is that we could get strange results if the time on various servers in the load test cluster would be out of sync.

jeremyandrews: We can make a note in the Gaggle section of the book, "It's important to keep reliable time between your servers if you want accurate metrics." I don't see this as a downside or a concern, and it would already affect existing logging. NTP was created around 1980 ...

src/metrics.rs
jeremyandrews: Tasks metrics has `thread_user.started.elapsed().as_millis(),` which is what is in the logs and is already accurate, why can't we use that?

jeremyandrews: And for users, why can't we also use `thread_user.started`?

src/goose.rs
```diff
@@ -1510,6 +1511,7 @@ impl GooseUser {
 
         // Once past the throttle, the request is officially started.
         let started = Instant::now();
+        let request_time = Utc::now();
```
jeremyandrews, line 1514: I don't understand what `Utc::now()` is giving us that `Instant::now()` isn't? It feels like we should use one or the other, but not both?

src/metrics.rs
jeremyandrews: IMO we shouldn't be making this change/fix in this PR as it's unrelated. Let's tackle this as a separate PR -- and in that case, I'd do what you've done here first, and then within the invoked function I'd add the second check only before where we care about if `error` isn't empty.

src/goose.rs
slashrsm: Reworked as we discussed so this is now gone.

src/metrics.rs
slashrsm: This was removed.

src/metrics.rs
slashrsm: Agreed. Removed.

src/metrics.rs
slashrsm: I used `GooseTaskMetric.elapsed`, which also required me to move this code into `receive_metric()` again. But this time it is in the place where task metrics are received, so I think that it makes more sense.

src/metrics.rs
```diff
@@ -2159,6 +2235,23 @@ impl GooseMetrics {
 
         Ok(())
     }
+
+    /// Records number of users for a current second.
+    ///
+    /// This is called from [`GooseAttack::sync_metrics()`] and the data
+    /// collected is used to display users graph on the HTML report.
+    pub(crate) fn record_users_per_second(&mut self) {
+        if let Some(starting) = self.starting {
```
jeremyandrews, line 2244: Just FYI: there are two upcoming changes that will impact this: 1) it will be possible to adjust the # of users of a running load test through the controller, so this value will be able to change outside of starting, and 2) in #365 it will become possible to shape the traffic to increase and decrease so again the value will be able to change outside of starting

src/report.rs
```diff
@@ -407,12 +417,112 @@ pub fn error_row(error: &metrics::GooseErrorMetricAggregate) -> String {
...
+        starting,
+        started,
+        stopping,
+        stopped,
+    )
+}
+
+/// Helper function to build HTML charts powered by the
+/// [ECharts](https://echarts.apache.org) library.
+#[allow(clippy::too_many_arguments)]
```
jeremyandrews, line 517: I always aim to restructure how I call the function instead of selectively ignoring Clippy. Perhaps we can create an object to pass in, for example?

src/metrics.rs
```diff
@@ -25,7 +25,9 @@ use crate::config::GooseDefaults;
 use crate::goose::{get_base_url, GooseMethod, GooseTaskSet};
 use crate::logger::GooseLog;
 use crate::report;
-use crate::util;
+use crate::util::{
+    median, ms_timer_expired, standard_deviation, timer_expired, truncate_string, MovingAverage,
+};
```
jeremyandrews, line 30: This change adds a lot of noise to the diff, and is an unnecessary style change. I'd much prefer to discuss and if we decide to rework the coding style then do it consistently across the entire codebase and not just in this file, which is now inconsistent with the rest of the codebase.

jeremyandrews (changes requested):
This is fantastic! A couple small nits (in particular a style change that creates a lot of noise) then it's ready to merge. Very useful addition, thanks again!

src/metrics.rs
```diff
@@ -1046,7 +1122,7 @@ impl GooseMetrics {
                 writeln!(
                     fmt,
                     " {:<24} | {:>13} | {:>14} | {:>8.reqs_p$} | {:>7.fails_p$}",
-                    util::truncate_string(request_key, 24),
+                    truncate_string(request_key, 24),
```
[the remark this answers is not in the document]
slashrsm: I am not even sure why I did that. Changed back.

src/metrics.rs
slashrsm: Done.

src/metrics.rs
slashrsm: I *think* that this should still work, but we will test it.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 0.15.2-dev
  - [#391](https://github.com/tag1consulting/goose/pull/391) properly sleep for configured `set_wait_time()` walking regularly to exit quickly if the load test ends
+ - [#394](https://github.com/tag1consulting/goose/pull/394) add additional graphs to the HTML report: errors per second,
+ average response time, active users, active tasks
```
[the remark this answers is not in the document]
slashrsm: Done.

src/metrics.rs
```diff
@@ -455,26 +463,75 @@ impl GooseRequestMetricAggregate {
...
+
+/// Expands vectors that collect per-second data for HTML report graphs with a
+/// default value.
+///
+/// We need to do that since we don't know for how long the load test will run
+/// and we can't initialize these vectors at the beginning. It is also
+/// better to do it as we go to save memory.
+fn expand_per_second_metric_array<T: Clone>(data: &mut Vec<T>, second: usize, initial: T) {
+    // Each element in per second metric vectors (self.requests_per_second,
+    // self.errors_per_second, ...) is count for a given second since the start
```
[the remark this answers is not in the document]
slashrsm: Done.

src/report.rs
slashrsm: Makes sense. Done. 

I ended up moving the function that generates markup onto the struct as well. This looks much nicer to me now.

src/metrics.rs
```diff
@@ -2308,6 +2399,9 @@ impl GooseAttack {
                 }
             }
 
+            // Record current users for users per second graph in HTML report.
+            self.metrics.record_users_per_second();
```
jeremyandrews, line 2403: I still think this should be moved down to _after_ we receive metrics, as otherwise we may not record a user until later than we should? It can be addressed in a follow, or let's talk through why this isn't relevant.

jeremyandrews (approved):
One small issue I'd like to talk through and address in a followup if it makes sense after we talk it through. Yay, merging!

PR #406, opened 2021-12-14
#400: Make sure that graphs are built if the load test is interrupted during startup.

### LINKED ISSUE #400: no graphs created if --no-reset-metrics not set and cancel during startup
If you start a load test without specifying `--no-reset-metrics` and cancel while the load test is still Starting, the html report doesn't have any data in the graphs.

For example:
```
cargo run --release --example drupal_memcache -- --host http://apache -u100 -v --report-file report.html --error-log errors.log --debug-log debug.log -s1h -t1m
```

If I cancel the load test during the hour long startup time, no data is displayed in the html report graphs.

slashrsm:
Closes #400.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.15.3-dev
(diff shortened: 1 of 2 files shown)


jeremyandrews (approved):
Looks great! I'll merge once tests pass.

PR #407, opened 2021-12-14
#405: Move GooseMetrics::record_users_per_second() call to the main loop.

### LINKED ISSUE #405: Check if users stats for graphs are recorded in the right place
This is a followup of #394. 

We need to make sure that the user stats are recorded after the update of the counter. Not sure if this is the case at the moment so we need to make sure it is or fix it.

slashrsm:
Closes #405.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1705,6 +1705,10 @@ impl GooseAttack {
                 // By reaching the Shutdown phase, break out of the GooseAttack loop.
                 AttackPhase::Shutdown => break,
             }
+
+            // Record current users for users per second graph in HTML report.
+            self.metrics.record_users_per_second();
+
             // Regularly synchronize metrics.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.15.3-dev
+ - [#407](https://github.com/tag1consulting/goose/pull/407) move `GooseMetrics::record_users_per_second()` call to the main loop
```
jeremyandrews, line 4: I don't think this is worthy of the Changelog, it shouldn't have any impact on the end user.

jeremyandrews (changes requested):

CHANGELOG.md
slashrsm: Removed.

jeremyandrews (approved):
Thanks!

PR #411, opened 2021-12-19
Move all graph functionality into an own struct/file and always enable startup data on graphs

### LINKED ISSUE #409: Always include startup phase data when rendering report graphs
At the moment we remove metrics for startup phase unless `--no-reset-metrics` is used. Ideally we would want to keep that data for graphs in any case.

slashrsm:
Closes #409.


src/graph.rs
```diff
@@ -0,0 +1,924 @@
...
+        )
+    }
+
+    /// Adds timestamps to the graph data series to ensure correct time display on x axis.
+    ///
+    /// Will take a vector of (generally numerical) values and convert them into tuples where
+    /// the second element will be the data point and the first element will be formatted time
+    /// it belongs to.
+    fn add_timestamp_to_html_graph_data<T: Copy>(&self, data: &[T]) -> Vec<(String, T)> {
+        // TODO properly handle the date Option timestamps.
```
jeremyandrews, line 215: I'm not understanding what's not handled properly here? What's the implication?

src/graph.rs
```diff
@@ -0,0 +1,924 @@
...
+        for _ in 0..(second - data.len() + 1) {
+            data.push(initial.clone());
+        }
+    };
+}
+
+/// Defines the HTML graph data.
+#[derive(Debug)]
+// TODO why does this need to be pub (instead pub(crate)) in order for stuff on
+// report.rs to not complain?
```
jeremyandrews, line 260: Typically this would be because the structure it's stored in is `pub`, not `pub(crate)`, and therefore requires everything it contains to also be `pub`.

src/graph.rs
slashrsm: Yeah, me neither. I was thinking that I should handle the case where those dates are not set yet. But then I realized that this function is never called if that is the case.

src/graph.rs
slashrsm: Thanks, makes sense! I changed the functions in `report.rs` to be `pub(crate)`.

jeremyandrews:
We should have a line in the Changelog flagging all the previously public APIs that are being removed with this commit, so anyone affected by this can quickly spot it.

src/lib.rs
```diff
@@ -396,14 +400,16 @@ impl GooseAttack {
...
+        let configuration = GooseConfiguration::parse_args_default_or_exit();
         Ok(GooseAttack {
             test_start_task: None,
             test_stop_task: None,
             task_sets: Vec::new(),
             weighted_users: Vec::new(),
             weighted_gaggle_users: Vec::new(),
             defaults: GooseDefaults::default(),
-            configuration: GooseConfiguration::parse_args_default_or_exit(),
+            graph_data: GraphData::new(configuration.no_reset_metrics),
```
jeremyandrews, line 411: Why disable startup metrics in the charts? It seems we collect the startup statistics either way, so IMO we should always show them. Or is there some reason they're not reliable?

.github/workflows/CI.yml
```diff
@@ -33,3 +33,5 @@ jobs:
       run: cargo rustdoc --lib --all-features --examples
     - name: Run tests
       run: cargo test --verbose --all-features
+    - name: Run timezone-sensitive tests in non UTC (GMT-5)
+      run: TZ="std +5" cargo test --verbose --all-features graph
```
jeremyandrews, line 37: So this test runs twice, once "normally" and once with a custom TZ?

jeremyandrews (changes requested):
Let's document the API breakage.

Also, I'd prefer to ALWAYS see startup metrics if we're collecting them (we could optionally add a new flag to disable them I suppose, but IMO not needed)

.github/workflows/CI.yml
slashrsm: Correct. I think that CI runs them in UTC since they passed when I am pretty sure they should fail with non UTC timezone. So I added this. It is a very short and quick test.

src/lib.rs
slashrsm: Done.

src/metrics.rs
```diff
@@ -396,12 +396,6 @@ pub struct GooseRequestMetricAggregate {
     /// The hash is primarily used when running a distributed Gaggle, allowing the Manager to confirm
     /// that all Workers are running the same load test plan.
     pub load_test_hash: u64,
-    /// Counts requests per second. Each element of the vector represents one second.
-    pub requests_per_second: Vec<u32>,
-    /// Counts errors per second. Each element of the vector represents one second.
-    pub errors_per_second: Vec<u32>,
-    /// Maintains average response time per second. Each element of the vector represents one second.
-    pub average_response_time_per_second: Vec<util::MovingAverage>,
```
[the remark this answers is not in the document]
slashrsm: Added a change record.

src/metrics.rs
```diff
@@ -396,12 +396,6 @@ pub struct GooseRequestMetricAggregate {
     /// The hash is primarily used when running a distributed Gaggle, allowing the Manager to confirm
     /// that all Workers are running the same load test plan.
     pub load_test_hash: u64,
-    /// Counts requests per second. Each element of the vector represents one second.
-    pub requests_per_second: Vec<u32>,
-    /// Counts errors per second. Each element of the vector represents one second.
-    pub errors_per_second: Vec<u32>,
```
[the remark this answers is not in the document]
slashrsm: Added a change record.

src/metrics.rs
```diff
@@ -396,12 +396,6 @@ pub struct GooseRequestMetricAggregate {
     /// The hash is primarily used when running a distributed Gaggle, allowing the Manager to confirm
     /// that all Workers are running the same load test plan.
     pub load_test_hash: u64,
-    /// Counts requests per second. Each element of the vector represents one second.
-    pub requests_per_second: Vec<u32>,
```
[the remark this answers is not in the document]
slashrsm: Added a change record.

src/graph.rs
```diff
@@ -0,0 +1,1279 @@
...
+        )
+    }
+}
+
+/// Data structure to maintain moving averages.
+///
+/// It will maintain the current average and the number of data items that
+/// were used to compute it.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+struct MovingAverage {
```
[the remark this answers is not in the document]
slashrsm: Added a change record.

src/util.rs
```diff
@@ -445,41 +444,6 @@ pub(crate) fn setup_ctrlc_handler(canceled: &Arc<AtomicBool>) {
     }
 }
 
-/// Data structure to maintain moving averages.
-///
-/// It will maintain the current average and the number of data items that
-/// were used to compute it.
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub struct MovingAverage {
```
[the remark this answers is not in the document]
slashrsm: Added a change record.

jeremyandrews (approved):
Thank you!

PR #414, opened 2022-01-04
Make structs and functions in report.rs visible only to the crate.

slashrsm:
This should not be a breaking change, since we are not exporting report as public in `lib.rs`:

```
mod report;
```


jeremyandrews:
Thanks!

PR #416, opened 2022-02-01
fix error-log documentation

nicompte:
Hello, from what I understand, `--error-file` has been renamed to `--error-log`, but that change was missing in the documentation.


jeremyandrews:
Thanks for the documentation fix! 

Normally I wouldn't merge with failures, but these are unrelated to your change and seem to be from a clippy update. I'll follow up to fix those.

PR #418, opened 2022-02-02
fix startup-time documentation

nicompte:
Another small documentation update ;)


jeremyandrews:
I greatly appreciate the documentation fixes, keep them coming! 👍 

As before, I'll go ahead and merge with the tests failing, as I'm fixing that in #417

PR #420, opened 2022-02-03
[WIP] Update to tokio 1.16

### LINKED ISSUE #419: update to tokio 1.16
When updating to tokio 1.16, the controller test fails. It seems that the tokio testing functionality may have changed, and that it's no longer properly running multi-threaded. The controller thread spins up a telnet (and/or websocket) state machine in its own thread which then makes requests against the Goose controller: with 1.16+ it seems only one of the threads runs.

jeremyandrews:
- update Tokio to 1.16.1+ (fixes #419)
 - timeout Goose if controller tests fail to shut it down after 60 seconds

@TODO: Solve the issue where tests appear to not be running in threads

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ serde = { version = "1.0", features = [
 serde_cbor = "0.11"
 serde_json = "1.0"
 simplelog = "0.10"
-tokio = { version = "~1.15.0", features = [
+tokio = { version = "1.16.1", features = [
     "fs",
     "io-util",
     "macros",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,8 +47,8 @@ tokio = { version = "1.16.1", features = [
     "time",
     "sync",
 ] }
-tokio-tungstenite = "0.15"
-tungstenite = "0.15"
+tokio-tungstenite = "0.16.1"
+tungstenite = "0.16.0"
 url = "2"
 
 # optional dependencies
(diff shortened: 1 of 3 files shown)


ahmedriza:
@jeremyandrews were you able to make the Gaggle tests to work.  Also running into the locked in `tokio` version `1.15` dependency issue.  Be nice to move onto the newer `tokio` versions.

jeremyandrews:
@ahmedriza Sorry, I missed your question somehow. I need to pick this back up again. Unfortunately it's a very non-trivial update, as I need to remove all non-async code (which is heavily used in Gaggles).

lamalex:
I'm confused what the problem is.
The tests already hang for me on `main` without updating to a recent Tokio.

So is upgrading Tokio an actual regression?
I'd be curious to know how many of the 50 forks are just Tokio version bumps.
Goose is really great - but this old pinned tokio is an achilles heel. It basically makes it impossible to integrate with an existing project.

```
$ cargo test --features gaggle

     Running tests/controller.rs (target/debug/deps/controller-8042a651a5d002a0)

running 2 tests
test test_telnet_controller has been running for over 60 seconds
test test_websocket_controller has been running for over 60 seconds
````

jeremyandrews:
All tests currently run successfully every time a PR is committed, so the tests are not currently broken.

If you don't need Gaggles, you'd be able to update tokio locally without problem. The underlying issue is explained in https://github.com/tag1consulting/goose/pull/420#issuecomment-1034914835 -- I have a WIP replacement that I've not had time to complete. Hopefully around new years ...

jeremyandrews:
Possibly? In any case, I'm planning to roll a release w/o Gaggle support shortly, which should allow us to upgrade all libraries. (And then Gaggle support will be added back in a future release, when it's ready.)

PR #422, opened 2022-03-20
Introduce `--test-plan` for more complex test plans

jeremyandrews:
- introduce `--test-plan` and `GooseDefault::TestPlan` for configuring test plans
 - closes https://github.com/tag1consulting/goose/issues/365
 - use [`FromStr`] to auto convert `--test-plan "{users},{timespan};{users},{timespan}"`, where `{users}` must be an integer, ie "100", and `{timespan}` can be integer seconds or "30s", "20m", "3h", "1h30m", etc, to internal `Vec<(usize, usize)>` representation
 - don't allow `--test-plan` together with `--users`, `--startup-time`, `--hatch-rate` or `--run-time`
 - add page to https://book.goose.rs/ explaining how to configure test plans

In the current implementation, a test plan can only Increase and Maintain, the load test will shut down any time it defines a Decrease. This will be resolved in a followup.

An example run:
```bash
% cargo run --release --example drupal_memcache -- --host http://apache.fosciana/ --test-plan "10,5;20,5;30,5;30,5;60,10;60,2;0,0"
```

Includes the follow overview in the metrics:
```
 === OVERVIEW ===
 Action       Started             Stopped           Elapsed    Users
 Increasing:  22-04-18 07:07:06 - 22-04-18 07:07:11 (00:00:05, 0 -> 10)
 Increasing:  22-04-18 07:07:11 - 22-04-18 07:07:17 (00:00:06, 10 -> 20)
 Increasing:  22-04-18 07:07:17 - 22-04-18 07:07:22 (00:00:05, 20 -> 30)
 Maintaining: 22-04-18 07:07:22 - 22-04-18 07:07:27 (00:00:05, 30)
 Increasing:  22-04-18 07:07:27 - 22-04-18 07:07:40 (00:00:13, 30 -> 60)
 Maintaining: 22-04-18 07:07:40 - 22-04-18 07:07:42 (00:00:02, 60)
 Decreasing:  22-04-18 07:07:42 - 22-04-18 07:07:42 (00:00:00, 0 <- 60)

 Target host: http://apache.fosciana/
 goose v0.16.0-dev
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
    o **note**: `-v` now sets Goose to `DEBUG` level verbosity which when enabled will negatively impact load test performance; set `-q` to restore previous level of verbosity
  - [#379](https://github.com/tag1consulting/goose/pull/379) **API change**: remove `.print()` which is no longer required to display metrics after a load test, disable with `--no-print-metrics` or `GooseDefault::NoPrintMetrics`
  - @TODO: introduce `--test-plan` and `GooseDefault::TestPlan`
+    o internally represent all load tests as `Vec<(usize, usize)>`l test plan
     
 ## 0.15.2 December 13, 2021
  - [#391](https://github.com/tag1consulting/goose/pull/391) properly sleep for configured `set_wait_time()` walking regularly to exit quickly if the load test ends
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,11 @@
  - [#379](https://github.com/tag1consulting/goose/pull/379) **API change**: default to `INFO` level verbosity, introduce `-q` to reduce Goose verbosity
    o **note**: `-v` now sets Goose to `DEBUG` level verbosity which when enabled will negatively impact load test performance; set `-q` to restore previous level of verbosity
  - [#379](https://github.com/tag1consulting/goose/pull/379) **API change**: remove `.print()` which is no longer required to display metrics after a load test, disable with `--no-print-metrics` or `GooseDefault::NoPrintMetrics`
- - @TODO: introduce `--test-plan` and `GooseDefault::TestPlan`
+ - [#422](https://github.com/tag1consulting/goose/pull/422) **API change**: introduce `--test-plan` and `GooseDefault::TestPlan`
     o internally represent all load tests as `Vec<(usize, usize)>`l test plan
+    o use [FromStr] to auto convert --test-plan "{users},{timespan};{users},{timespan}", where {users} must be an integer, ie "100", and {timespan} can be integer seconds or "30s", "20m", "3h", "1h30m", etc, to internal Vec<(usize, usize)> representation
+    o don't allow `--test-plan` together with `--users`, `--startup-time`, `--hatch-rate`, `--run-time`, `--no-reset-metrics`, `--manager` and `--worker`
+    o internal `AttackPhase`s renamed: `Starting` -> `Increase`, `Running` -> `Maintain`, `Stopping` -> `Decrease`
     
 ## 0.15.2 December 13, 2021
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
  - [#379](https://github.com/tag1consulting/goose/pull/379) **API change**: default to `INFO` level verbosity, introduce `-q` to reduce Goose verbosity
    o **note**: `-v` now sets Goose to `DEBUG` level verbosity which when enabled will negatively impact load test performance; set `-q` to restore previous level of verbosity
  - [#379](https://github.com/tag1consulting/goose/pull/379) **API change**: remove `.print()` which is no longer required to display metrics after a load test, disable with `--no-print-metrics` or `GooseDefault::NoPrintMetrics`
+ - @TODO: introduce `--test-plan` and `GooseDefault::TestPlan`
```
slashrsm, line 16: PR # missing

src/metrics.rs
```diff
@@ -2032,73 +2059,102 @@ impl GooseMetrics {
...
+        writeln!(
+            fmt,
+            "\n === OVERVIEW ===\n ------------------------------------------------------------------------------"
+        )?;
 
+        writeln!(
+            fmt,
+            " {:<12} {:<19} {:<17} {:<10} Users",
+            "Action", "Started", "Stopped", "Elapsed",
+        )?;
```
slashrsm, line 2075: I am wondering if this output really belongs in this file. Are these metrics or something else? 

`metrics.rs` is already a huge file and we might want to consider splitting it at some point? Maybe this is a good time to start this discussion (actually doing it is out of scope for this PR of course).

src/config.rs
```diff
@@ -367,6 +375,126 @@ pub(crate) struct GooseDefaults {
...
+/// Internal data structure representing a test plan.
+//#[derive(Options, Debug, Clone, Serialize, Deserialize)]
+#[derive(Options, Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct TestPlan {
+    // A test plan is a vector of tuples each indicating a # of users and milliseconds.
+    pub(crate) steps: Vec<(usize, usize)>,
+    // Which step of the test_plan is currently running.
+    pub(crate) current: usize,
+}
+
```
slashrsm, line 387: I would put test plan and its logic in a separate file. I would also include `TestPlanStepAction` and `TestPlanHistory` that are currently in `metrics.rs`.

slashrsm: Besides that I'd also add some test coverage.

slashrsm (commented):
I tested and I can confirm that it works as expected (with the current limitations). Great job! This is not a small task!

It would be awesome to have some more automated tests for this as we're dealing with quite significant change. 

I would also propose to create a list of follow-ups that are required and mark those that are release blocker. This way we'll avoid forgetting to fix something crucial.

src/config.rs
jeremyandrews: There's very little logic here, but I suppose it's likely going to grow -- splitting it out works for me. (Especially as when I was writing this I could never remember which file it lived in).

And yes, it DEFINITELY needs tests! :)

CHANGELOG.md
jeremyandrews: Thanks, fixed!

src/metrics.rs
jeremyandrews: This is the ASCII display of the metrics. I suppose it could be moved into report.rs if that's a better fit?

src/config.rs
jeremyandrews: I've split it into its own file now -- definitely the right call, thanks.

slashrsm:
Awesome work! 🎉

LionsAd:
RTBM - retro — great work 👍

PR #427, opened 2022-03-29
#426: Allow --report-file on managers.

### LINKED ISSUE #426: HTML report file when using Gaggle
How would I generate an HTML report when using a manager node and multiple workers? Is this even possible?
* The manager doesn't accept the `--report-file` flag
* The worker does, but after a loadtest finishes (either Ctrl-C or via `--run-time`, it throws this error:
```
thread 'main' panicked at 'internal error: entered unreachable code', C:\Users\nbellack\.cargo\registry\src\github.com-1ecc6299db9ec823\goose-0.15.2\src\metrics.rs:2189:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\debug\contitrade-loadtesting.exe --worker` (exit code: 101)
```

slashrsm:
Closes #426.

I think that it makes sense to allow `--report-file` on managers and disallow them on workers.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1905,13 +1905,6 @@ impl GooseConfiguration {
                     detail: "`configuration.no_autostart` can not be set on the Manager."
                         .to_string(),
                 });
-            } else if !self.report_file.is_empty() {
-                return Err(GooseError::InvalidOption {
-                    option: "`configuration.report_file`".to_string(),
-                    value: self.report_file.to_string(),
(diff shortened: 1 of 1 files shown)


jeremyandrews (approved):
Looks good!

jeremyandrews:
@slashrsm please rebase

slashrsm:
We should rework how this works instead. See: https://github.com/tag1consulting/goose/issues/426#issuecomment-1081826319

Issue #434, opened 2022-04-09
taskset => scenario, task => step


kazimir-malevich:
I know Goose is based on Locust and this would be a massive refactor. But I think vocabulary is important and testers would see **tasksets** as **scenarios** and **tasks** as **steps**.  Is a rename worth it?

jeremyandrews:
I think whatever names are chosen, someone will be confused if they're used to different names from a previous load testing tool. I'd prefer to see improved documentation that explains "if you have a jMeter background, you can consider a TaskSet similar to a Scenario" etc.

kazimir-malevich:
See PR **glossary**

jeremyandrews:
Doing a little more research on other load testing tools, I'm leaning more and more toward renaming `GooseTaskSet` to `Scenario`. Ideally the goal will to allow either name for a major version or two with a deprecation warning.

kazimir-malevich:
**Scenario** has been used since manual testing days. I also thinking removing **tasks** and having the API more granular. We make requests **using HTTP VERBS**, we validate a 200 or 300. We make additional assertions using against the HTML body. I just don't know enough as yet, but things come across as confusing.

jeremyandrews:
Whatever they are named, the abstraction is fairly simple: Individual Requests (which indeed map to HTTP verbs) are grouped together in Tasks, which can be further grouped together in TaskSets. 

I recommend taking a look at this example for a better understanding of this works, and the added flexibility it offers beyond what you're describing: https://github.com/tag1consulting/goose-eggs/tree/main/examples/umami

kazimir-malevich:
I was thinking last night, the design is correct. You'd want to measure timings of transactions/tasks to determine bottlenecks in specific areas of the AUT, which no other load test I am aware of does. But I would suggest **scenarios** and **transactions** for naming *tasksets** and **tasks**. I need to learn rust now, which will take a few days, but feel free to assign me simple tasks to begin with. I am not intentionally criticising, I just believe this has the potential to be the best and most comprehensive load test tool around..

jeremyandrews:
> But I would suggest scenarios and transactions for naming tasksets and tasks.

[...]

PR #435, opened 2022-04-10
Glossary

kazimir-malevich:
Glossary start

### DIFF
--- a/src/docs/goose-book/src/glossary.md
+++ b/src/docs/goose-book/src/glossary.md
@@ -6,5 +6,9 @@ case** or a **scenario** in other test tools. It is a class of actions or steps
 that the user would undertake to achieve a specific user journey.
 
 ## Task
-A **task** is a single step within a **taskset**, like GET a homepage or POST a
-form.
+A **task** is what we may normally define as a transaction. A collection of one
+or more requests that may include logging into a website or adding to a
+shopping cart, and could include the validation of expectations.
+
+## Request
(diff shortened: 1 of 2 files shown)


src/docs/goose-book/src/glossary.md
```diff
@@ -0,0 +1,10 @@
+# Glossary
+
+## Taskset
+A **taskset** in Goose is akin to what we would normally define as a **test
+case** or a **scenario** in other test tools. It is a class of actions or steps
+that the user would undertake to achieve a specific user journey.
+
+## Task
+A **task** is a single step within a **taskset**, like GET a homepage or POST a
+form.
```
jeremyandrews, line 10: Tasks are more arbitrary than this. A [`GooseRequest`](https://docs.rs/goose/latest/goose/goose/struct.GooseRequest.html) maps to an individual request. Whereas a [`GooseTask`](https://docs.rs/goose/latest/goose/goose/struct.GooseTask.html) is a collection of one or more requests that typically would simulate a single workflow (such as logging into a website: which may require loading the page, scraping the form, posting the username, login and form elements, and finally loading the resulting profile page and validating the expected elements exist).

jeremyandrews:
I'm holding off on merging this, as we're planning to rename things and so it likely won't be necessary, or will then need to be updated.

slashrsm:
#434 (#449) is done so we can merge this too. Thanks!

PR #438, opened 2022-04-14
Add error_on_fail to return error if request fails

d00z3l:
This adds a `error_on_fail()` option to GooseRequest which will return an error if the request fails

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -1628,6 +1628,19 @@ impl GooseUser {
             self.send_request_metric_to_parent(request_metric.clone())?;
         }
 
+        if request.error_on_fail && !request_metric.success {
+            if let Ok(r) = response {
+                if let Ok(body) = r.text().await {
+                    if !body.is_empty() {
+                        error!("{:?} {}, {}", &path, &request_metric.error, &body);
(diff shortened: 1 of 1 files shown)


src/goose.rs
```diff
@@ -1628,6 +1628,19 @@ impl GooseUser {
             self.send_request_metric_to_parent(request_metric.clone())?;
         }
 
+        if request.error_on_fail && !request_metric.success {
+            if let Ok(r) = response {
+                if let Ok(body) = r.text().await {
+                    if !body.is_empty() {
+                        error!("{:?} {}, {}", &path, &request_metric.error, &body);
```
jeremyandrews, line 1635: This information is already available if you enable `--debug-log`. I'd much prefer to keep it there, as this can lead to huge errors on the console that are difficult to review. This will simplify this block of code.

jeremyandrews (changes requested):
Please run `cargo fmt` which should fix the formatting error, and simplify the `request_on_fail` logic, relying instead on `--debug-log` to obtain the full response body when there are errors.

jeremyandrews (approved):

PR #443, opened 2022-04-19
#440: Re-enable HTML report after introduction of test plans.

### LINKED ISSUE #440: restore HTML report and Graphs
PR https://github.com/tag1consulting/goose/pull/422 temporarily disabled `--report-file` which needs to be updated to properly support the new Test Plan functionality. The HTML report should be able to generally duplicate `fmt_overview`. The graphs will need to be reworked as `GoosePhase::Starting`, `::Running` and `::Stopping` have been replaced with `::Increase`, `::Maintain` and `::Decrease`, each of which can happen multiple times during a given test plan.

slashrsm:
Closes #440. 

- Re-enables report file
- Instead of starting, started, stopping, stopped summary displays a table of plan steps overview
- Removes "Starting" and "Stopping" section from graphs
- Add increasing (green) and decreasing (red) sections to the graph

Screenshot 2022-04-21 at 12-24-03 Goose Attack Report

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -254,7 +254,7 @@ impl<'a, T: Clone + TimeSeriesValue<T, U>, U: Serialize + Copy + PartialEq + Par
                         r#"[
                             {{
                                 xAxis: '{started}',
-                                itemStyle: {{ borderColor: 'rgba(44, 102, 79, 1)', borderWidth: 1 }},
+                                itemStyle: {{ borderColor: 'rgba(44, 102, 79, 0.25)', borderWidth: 1 }},
                             }},
                             {{
                                 xAxis: '{started}'
@@ -263,7 +263,7 @@ impl<'a, T: Clone + TimeSeriesValue<T, U>, U: Serialize + Copy + PartialEq + Par
                         [
                             {{
                                 xAxis: '{started}',
-                                itemStyle: {{ color: 'rgba(44, 102, 79, 0.25)' }},
+                                itemStyle: {{ color: 'rgba(44, 102, 79, 0.05)' }},
                             }},
                             {{
                                 xAxis: '{stopped}'
@@ -272,7 +272,7 @@ impl<'a, T: Clone + TimeSeriesValue<T, U>, U: Serialize + Copy + PartialEq + Par
                         [
                             {{
                                 xAxis: '{stopped}',
-                                itemStyle: {{ borderColor: 'rgba(44, 102, 79, 1)', borderWidth: 1 }},
+                                itemStyle: {{ borderColor: 'rgba(44, 102, 79, 0.25)', borderWidth: 1 }},
                             }},
                             {{
                                 xAxis: '{stopped}'
@@ -288,7 +288,7 @@ impl<'a, T: Clone + TimeSeriesValue<T, U>, U: Serialize + Copy + PartialEq + Par
                         r#"[
                             {{
                                 xAxis: '{started}',
-                                itemStyle: {{ borderColor: 'rgba(179, 65, 65, 1)', borderWidth: 1 }},
+                                itemStyle: {{ borderColor: 'rgba(179, 65, 65, 0.25)', borderWidth: 1 }},
                             }},
                             {{
                                 xAxis: '{started}'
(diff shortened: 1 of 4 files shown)


src/metrics.rs
```diff
@@ -2607,50 +2602,61 @@ impl GooseAttack {
...
-                report_range.push_str(&format!(
-                    "<p>Running: <span>{} - {} (Duration: {:02}:{:02}:{:02})</span></p>",
-                    started.format("%Y-%m-%d %H:%M:%S"),
-                    stopping.format("%Y-%m-%d %H:%M:%S"),
-                    running_hours,
-                    running_minutes,
-                    running_seconds,
-                ));
+            let mut steps_overview = String::new();
+            for step in self.metrics.history.windows(2) {
```
slashrsm, line 2606: This loop is basically a copy-paste of the same thing in console printout part. Maybe we should move it to a helper and try to share it?

src/lib.rs
```diff
@@ -1733,15 +1735,15 @@ impl GooseAttack {
...
                     self.sync_metrics(&mut goose_attack_run_state, true).await?;
                     // Record last users for users per second graph in HTML report.
-                    //self.graph_data.record_users_per_second(self.metrics.users, Utc::now());
+                    self.graph_data
+                        .record_users_per_second(self.metrics.users, Utc::now());
                     // The load test is fully stopped at this point.
                     self.metrics
                         .history
                         .push(TestPlanHistory::step(TestPlanStepAction::Finished, 0));
                     //self.graph_data.set_stopped(Utc::now());
```
jeremyandrews, line 1744: This is no longer needed, delete.

src/graph.rs
```diff
@@ -16,14 +17,7 @@ use std::marker::PhantomData;
 /// Used to collect graph data during a load test.
 pub(crate) struct GraphData {
     /// Tracks when the load test first started with an optional system timestamp.
-    starting: Option<DateTime<Utc>>,
-    /// Tracks when all [`GooseUser`](../goose/struct.GooseUser.html) threads fully
-    /// started with an optional system timestamp.
     started: Option<DateTime<Utc>>,
```
jeremyandrews, line 20: Why is this needed? Is this missing in `metrics.history`?

src/graph.rs
```diff
@@ -53,26 +44,11 @@ impl GraphData {
         }
     }
 
-    /// Sets starting time.
-    pub(crate) fn set_starting(&mut self, starting: DateTime<Utc>) {
-        self.starting = Some(starting);
-    }
-
     /// Sets started time.
     pub(crate) fn set_started(&mut self, started: DateTime<Utc>) {
```
jeremyandrews, line 48: I think this is unnecessary now?

jeremyandrews (changes requested):
I don't think we need to track `started` a second time, but can instead get it from `metrics.history`?

src/lib.rs
jeremyandrews, line 1744: `//self.graph_data.set_stopped(Utc::now());`

It seems this isn't used anymore, so can be removed?

slashrsm: Done.

src/graph.rs
slashrsm: Removed

src/graph.rs
slashrsm: Indeed.

src/lib.rs
```diff
@@ -78,12 +78,13 @@ use tokio::fs::File;
...
 use crate::controller::{GooseControllerProtocol, GooseControllerRequest};
 use crate::goose::{GaggleUser, GooseTask, GooseTaskSet, GooseUser, GooseUserCommand};
-//use crate::graph::GraphData;
+use crate::graph::GraphData;
 use crate::logger::{GooseLoggerJoinHandle, GooseLoggerTx};
 use crate::metrics::{GooseMetric, GooseMetrics};
 use crate::test_plan::{TestPlan, TestPlanHistory, TestPlanStepAction};
 #[cfg(feature = "gaggle")]
 use crate::worker::{register_shutdown_pipe_handler, GaggleMetrics};
+use chrono::prelude::*;
```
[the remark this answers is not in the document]
slashrsm: This is gone now.

src/metrics.rs
```diff
@@ -2604,61 +2599,73 @@ impl GooseAttack {
...
 
-    /* @FIXME:
     // Write an HTML-formatted report, if enabled.
     pub(crate) async fn write_html_report(
         &mut self,
         goose_attack_run_state: &mut GooseAttackRunState,
     ) -> Result<(), GooseError> {
         // Only write the report if enabled.
         if let Some(report_file) = goose_attack_run_state.report_file.as_mut() {
+            let test_start_time = self.metrics.history.first().unwrap().timestamp;
```
jeremyandrews, line 2609: I try to use `.expect()` instead of `.unwrap()` to more explicitly document our assumptions -- but not a blocker as it will panic either way.

jeremyandrews (approved):
Great work!

PR #446, opened 2022-04-20
test test-plan

jeremyandrews:
- test configuration of test plan with `--test-plan` (which includes multiple Increase and Maintain steps)
 - test configuration of test plan with `GooseDefault::TestPlan` (which includes multiple Increase and Maintain steps)
 - internally configure `users` before `test_plan`, to avoid automatically setting the former when building the latter from `GooseDefault::TestPlan`
 - when calculating the `hatch_rate` from the test plan, subtract the first GooseUser thread of each step as it launches immediately without delay
 - determine when all users for a test plan are launched based on if running stand-alone or in a gaggle configuration, but not based on compiled-in features, as gaggles can be compiled in but not used
 - when tracking duration, round partial seconds (so 1.2 seconds rounds down to 1 second, and 1.6 seconds rounds up to 2 seconds -- previously 1.999 was still considered 1 second)
 - add longer delay in sequence test to account for improved accuracy on hatch_rate and duration tracking


src/lib.rs
```diff
@@ -1320,8 +1320,9 @@ impl GooseAttack {
         // in the current step.
         debug_assert!(self.test_plan.steps[self.test_plan.current].0 > previous_users);
 
-        // Divide the number of new users to launch by the time configured to launch them.
-        let hatch_rate: f32 = (self.test_plan.steps[self.test_plan.current].0 - previous_users)
+        // Divide the number of new users to launch (minus one, as the first launches immediately without any delay)
```
slashrsm, line 1323: Out of curiosity - why is that the case?

jeremyandrews: It's something we could revisit, but originally it's because when Starting a load test the first GooseUser thread starts immediately, versus waiting whatever hatch-rate milliseconds before starting the first user, as otherwise the load test is idle for some time.

Another idea would be to just add a delay after the last user starts but before we advance to the next step. But this gets a little fuzzy if starting lots of users quickly because users may not start at the exact cadence configured (ie, if the server simply can't keep up with it).

PR #450, opened 2022-04-25
Support multiple and variable speed Decrease attack phases

jeremyandrews:
- Support multiple and variable speed decrease attack phases

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -353,6 +353,8 @@ struct GooseAttackRunState {
     /// Thread-safe boolean flag indicating if the [`GooseAttack`](./struct.GooseAttack.html)
     /// has been canceled.
     canceled: Arc<AtomicBool>,
+    /// Whether or not the load test is currently canceling.
+    canceling: bool,
     /// Optional socket used to coordinate a distributed Gaggle.
     socket: Option<Socket>,
 }
@@ -1299,6 +1301,7 @@ impl GooseAttack {
             all_users_spawned: false,
             shutdown_after_stop: !self.configuration.no_autostart,
             canceled: Arc::new(AtomicBool::new(false)),
+            canceling: false,
             socket,
         };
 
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -1732,45 +1808,48 @@ impl GooseAttack {
...
-                    break;
-                }
+                AttackPhase::Shutdown => break,
             }
 
             // Record current users for users per second graph in HTML report.
             if let Some(started) = self.started {
                 self.graph_data.record_users_per_second(
-                    self.metrics.users,
+                    goose_attack_run_state.active_users,
```
slashrsm, line 1852: Other graphs look OK, but this one seems odd. Is this variable always storing the number of users at the time or the target number for the step?

slashrsm (commented):
Tested and I can confirm that it works as expected. I noticed two things:

1. If I interrupted the load test with Ctrl+C the decrease after that moment isn't presented as a single step but as a lot of really short steps:

Screenshot 2022-04-26 at 17 49 29
Screenshot 2022-04-26 at 17 49 20

The overview in the CLI looks similar. 

2. Initially I made a mistake with the plan string formatting (swapped user and time in pairs). Test took a very long time and when I interrupted it I got a panic: 

```
thread 'main' panicked at 'index out of bounds: the len is 3 but the index is 3', 
.../goose/src/lib.rs:1545:52
```

Ideally we would want to catch such a mistake and warn the user. With the current format it would be hard to do that I suppose, but what if we tweak the formatting a bit? JSON maybe (e.g. `[{users: 5, time: 30}, {users: 10, time: 30}, {users: 0, time: 10}]`)?

If that is not possible we should at least try to exit without panic when the test is interrupted so that user can understand the mistake from the overview that is printed at the end.

src/lib.rs
slashrsm: In my case it happened after the first `Increase` step when the `Maintain` step started. After that is seems OK. 

The load test was started with `--test-plan "5,10;5,20;20,10;20,30;100,10;20,10;20,30;0,20"`

jeremyandrews:
I've addressed the bugs you ran into, introducing a new `TestPlanStepAction::Canceling` so that we also properly report when a load test was canceled versus stopping automatically.

I would not want to require someone to manually type out JSON as I find this complex (something I've done when testing APIs with Curl), but we could consider optionally supporting this if necessary. I do agree it's confusing however and easy to get backwards (I've done the same) -- I've fixed it so it doesn't panic anymore. I'd prefer to address this in a followup (we could support both the current style, and the json style. Or we could make it something like "u#,t#" where u is the prefix on users and t is the prefix on time, for example?)

I also created https://github.com/tag1consulting/goose/issues/453 another followup to add testing -- it's rather non-trivial so I'd prefer to not hold up this landing.

slashrsm (approved):
I agree re JSON. And the issue is mitigated by the fact that user now sees the overview and can understand the mistake. When we will create the UI and we'll potentially allow to configure the test plan there the issues will be even less relevant.

I pushed one small change that marks the `Cancelling` phase as red on graphs.

Issue #439, opened 2022-04-15
Default config


kazimir-malevich:
Do you think print() and StatusCodes should be made default? We'd definitely want to see percentiles at the end of the test run and HTTP status codes are useful as well. I'm thinking more towards new users.

jeremyandrews:
We already committed a a PR to print() by default which will be in `0.16` when released, see:
As of https://github.com/tag1consulting/goose/pull/379

I'm not opposed to changing `--status-codes` to `--no-status-codes` so we collect and display the Status Code table by default, PR welcome (documentation will also need to be updated). The current default is because there is a small amount of performance and memory overhead to enabling this option, but I generally always enable it when I'm running load tests anyway so it's a logical default.

kazimir-malevich:
Where's the Goose binary?
```
$:~/projects/goose$ cargo build --release
Finished release [optimized] target(s) in 0.12s
$:~/projects/goose$ cd target/release/
$~/projects/goose/target/release$ ls
build  deps  examples  incremental  libgoose.d  libgoose.rlib
```

jeremyandrews:
Goose itself is a library, so there is no binary. You can build an example to create a binary, or create a load test as documented in the book to create a binary with it.

PR #452, opened 2022-04-26
enable status code summary by default

jeremyandrews:
- rename `--status-codes` to `--no-status-codes` (and `GooseDefault::StatusCodes` to `GooseDefault::NoStatusCodes`)
 - collect and summarize status codes by default
 - update the documentation (to review locally, run `mdbook build src/docs/goose-book --open`)
 - while updating the metrics book page (which originally included `--status-codes`) I also added the html report
 - closes https://github.com/tag1consulting/goose/issues/439


src/config.rs
```diff
@@ -2378,11 +2378,11 @@ impl GooseConfiguration {
         // Validate `no_metrics`.
         if self.no_metrics {
             // Status codes are not collected if metrics are disabled.
-            if self.status_codes {
+            if self.no_status_codes {
```
slashrsm, line 2381: Isn't this the other way around - status codes can't be enabled if we don't have metrics but can be disabled in any case?

jeremyandrews: Thanks for catching that! This check is no longer relevant -- if `--no-metrics` is enabled then status codes are disabled automatically (whether or not `--no-status-codes` are enabled).

src/config.rs
```diff
@@ -2157,11 +2157,11 @@ impl GooseConfiguration {
                         .to_string(),
                 });
             // Can't set `status_codes` on Worker.
```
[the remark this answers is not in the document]
jeremyandrews: Thanks! I had missed one other as well, also fixed.

slashrsm:
Awesome work!

PR #461, opened 2022-05-02
Prefer explicit time notation in the documentation.

slashrsm:


src/docs/goose-book/src/getting-started/runtime-options.md
```diff
@@ -45,7 +45,7 @@ Metrics:
   --no-status-codes           Do not track status code metrics
 
 Advanced:
-  --test-plan "TESTPLAN"      Defines a more complex test plan ("10,60;0,30")
+  --test-plan "TESTPLAN"      Defines a more complex test plan ("10,60s;0,30s")
```
jeremyandrews, line 48: You'll also need to update config.rs at https://github.com/tag1consulting/goose/blob/main/src/config.rs#L210, and then also at https://github.com/tag1consulting/goose/blob/main/src/config.rs#L78

(Run `cargo run --release --example simple -- -h` to be sure the help is properly updated.)

jeremyandrews (changes requested):
This is very helpful, thanks! However, you also need to update `src/config.rs` so the actual built-in help is correct and matches the documentation.

src/docs/goose-book/src/getting-started/runtime-options.md
slashrsm: Done.

jeremyandrews (approved):
Thank you for this clarification! It definitely helps make the order of the tuple-style configuration much more clear.

Issue #454, opened 2022-04-28
support dynamically changing users and hatch rate during load test with controller


jeremyandrews:
Now that Goose is always using a Test Plan internally, it's possible to add support for dynamically changing the # of active users and/or the rate users are increasing or decreasing while the load test is running. When using a predefined test plan, Goose pre-allocates all users for optimal performance and to strictly enforce the configured scheduler.

In the case of adjusting the number of users through the controller there's no way to pre-allocate the users as we don't know how many will be needed, so we'll need to add support for dynamically allocating new users when needed. Internally, Goose can treat a re-configuration the same it treats receiving SIGINT (ie, rebuild the internal testplan and record the actual numbers in the metric history).

PR #469, opened 2022-05-05
allow dynamically configuring users during running load test

jeremyandrews:
- accept the `users` controller command during a running load test
 - allocate new users on-the-fly if the number of users is increased during a running load test
 - test dynamically increasing and decreasing users during a running load test
 - users are increased/decreased per configured `hatch_rate` (or at a default of 1 per second)
 - add book page with example graph (use `mdbook build src/docs/goose-book --open` to preview)
 - builds on https://github.com/tag1consulting/goose/pull/464 which must be merged first
 - closes https://github.com/tag1consulting/goose/issues/454


src/controller.rs
```diff
@@ -1216,31 +1221,101 @@ impl GooseAttack {
...
+                                        let total_time =
+                                            (ms_hatch_rate * user_difference as f32) as usize;
+
+                                        // Reset the test_plan to adjust to the newly specified users.
+                                        self.test_plan.steps = vec![
+                                            // Record how many active users there are currently.
+                                            (goose_attack_run_state.active_users, elapsed),
+                                            // Configure the new user count.
+                                            (new_users, total_time),
+                                        ];
```
slashrsm, line 1287: Won't this remove history if there were multiple steps before this change?

jeremyandrews: No, history is stored in `self.metrics.history` -- we do this same trick in [cancel_attack()](https://github.com/tag1consulting/goose/blob/main/src/lib.rs#L1728).

jeremyandrews: (That's part of why I showed the Overview table in the book page: that's the actual history shown after executing this controller command multiple times.)

Issue #470, opened 2022-05-05
Encapsulate all data (and logic) for controller commands


slashrsm:
Currently we carry definition of controller commands in a few places. It would be nice if we could encapsulate all their data (parsing regular expression, status messages, error messages, ...) and their logic (validation, implementation, ...) in a single place and let wrapping logic be completely generic. This would make it easier to add new commands, to test them and make code easier to read, since we would have everything that belongs to a single command in one place. 

Maybe we could define a trait (or a set of traits) that all commands must implement. Or something along those lines? 

More info: https://github.com/tag1consulting/goose/pull/464#discussion_r865644181

PR #471, opened 2022-05-08
combine controller command logic into central place

jeremyandrews:
- introduce `ControllerCommandDetails` which includes `help`, `regex`, `process_response` all defined together
 - drop `Goose` prefix from Controller
 - improve `help` formatting (logically grouping and separating command types)
 - move anything that has to be modified when adding new commands to the top of the source file
 - closes https://github.com/tag1consulting/goose/issues/470


slashrsm (approved):
Looks good. I also tested both Telnet and WebSocket controllers and they worked as expected.

PR #487, opened 2022-05-21
confirm Goose shuts down correctly when canceled

jeremyandrews:
- move CANCELED flag into a global lazy_static so it can be reset and re-used on multiple tests
 - move Gaggle-shutdown handling to new SHUTDOWN_GAGGLE flag (also a global lazy_static) so that a Gaggle test shutting down at the same time as a non-Gaggle test is running won't cause both to be shut down
 - test coverage with `--run-time` configured
 - test coverage without `--run-time` configured (ie, run forever until canceled)
 - test coverage with `--iterations` configured
 - test coverage with `--test-plan` including canceling while the test is Increasing, Maintaining, and Decreasing
 - adds dev-dependency on [`nix`](https://docs.rs/nix) in order to access `kill` for sending `SIGINT`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 0.16.3-dev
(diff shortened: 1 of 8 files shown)


slashrsm (approved):
Looks good to me. Awesome work!

Issue #499, opened 2022-06-13
Machine-readable results


samoylovfp:
I would like to request a feature of writing the resulting data into a machine-readable format.
Something like json, so results from different runs could be combined together.

Will you accept a merge request with this functionality?
I would replace all the calls to `self.write_html_report().await?;` with `self.write_reports().await?;`
and have a serde-serializeable structure to hold the results, which is then used to generate both the json file and the html report.

samoylovfp:
Still working on this, I made a little bit of a refactoring to avoid copying a lot of error handling regarding writing things to file, looking to finalize this probably next week

PR #504, opened 2022-06-21
Machine-readable reports

samoylovfp:
Solves https://github.com/tag1consulting/goose/issues/499 (once finished)

The idea is to have reports written into a json file,
so further processing and merging can be made out of the process

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,7 @@ use lazy_static::lazy_static;
 use nng::Socket;
 use rand::seq::SliceRandom;
 use rand::thread_rng;
+use report::ReportFile;
 use std::collections::{hash_map::DefaultHasher, BTreeMap, HashSet};
 use std::hash::{Hash, Hasher};
 use std::sync::{
@@ -71,7 +72,6 @@ use std::sync::{
 };
 use std::time::{self, Duration};
 use std::{fmt, io};
-use tokio::fs::File;
 
 use crate::config::{GooseConfiguration, GooseDefaults};
 use crate::controller::{ControllerProtocol, ControllerRequest};
@@ -878,10 +878,13 @@ impl GooseAttack {
     }
 
     // If enabled, returns the path of the report_file, otherwise returns None.
-    fn get_report_file_path(&mut self) -> Option<String> {
+    fn get_report_file(&mut self) -> Option<ReportFile> {
         // Return if enabled.
         if !self.configuration.report_file.is_empty() {
-            Some(self.configuration.report_file.to_string())
+            Some(ReportFile {
+                path: self.configuration.report_file.to_string(),
+                command_line_option: "--report-file".to_string(),
+            })
         // Otherwise there is no report file.
         } else {
             None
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -124,6 +125,44 @@ pub fn get_worker_id() -> usize {
...
+        let cmd = self.command_line_option.clone();
+
+        file.write_all(report)
+            .await
+            .map_err(|e| GooseError::InvalidOption {
+                option: cmd,
+                value: path,
+                detail: format!("Failed to write to report file: {}", e),
+            })?;
+        // FIXME: should we add context here as well?
```
samoylovfp, line 160: @jeremyandrews I would appreciate feedback on this

src/lib.rs
```diff
@@ -124,6 +125,44 @@ pub fn get_worker_id() -> usize {
...
+                option: self.command_line_option.clone(),
+                value: self.path.clone(),
+                detail: format!("Failed to create report file: {}", e),
+            })?;
+        self.file = Some(file);
+        Ok(())
+    }
+
+    pub(crate) async fn write_all_and_flush(&mut self, report: &[u8]) -> Result<(), GooseError> {
+        let file = self.file.as_mut().unwrap();
```
jeremyandrews, line 149: I'm leery of having an `.unwrap()` here. At minimum, please add a comment explaining why this is safe.

src/lib.rs
jeremyandrews: It's definitely helpful to add context anywhere we can fail. I imagine flush can fail for example if a filesystem fills up, and that's definitely something that can happen during a large load test.

src/lib.rs
```diff
@@ -124,6 +125,44 @@ pub fn get_worker_id() -> usize {
 /// Socket used for coordinating a Gaggle distributed load test.
 pub(crate) struct Socket {}
 
+/// Wrapper over File object that contains context for errors
```
jeremyandrews, line 128: Perhaps instead of moving this all into `lib.rs`, it would be more appropriate for it to live in `report.rs`?

jeremyandrews (changes requested):
So far it's looking good. Adding more context to errors is always helpful. A few little comments, I'm looking forward to seeing the machine-readable aspects added.

samoylovfp:
Didn't have time to get to it this week and I am on holiday next 3 weeks. I expect I'll return to this some time in the next 40 days

src/lib.rs
samoylovfp: Added

src/lib.rs
samoylovfp: Removed the filed, the files is now created here

src/lib.rs
samoylovfp: Good idea. Moved there

samoylovfp:
We found out we can extract the need information from requests log.

mTsBucy1:
@samoylovfp
May I ask how you further processed the data? Did you parse the json requests log via code? Or did you use a Spreadsheet application?

Because I am currently in the process of using the csv requests log (which lead to #515)  and importing in Excel, where I can do more customized processing, and I was wondering what other paths there might be.

samoylovfp:
We wrote our own small app using egui, that loads different benchmarking results and shows request time histograms on a log scale. Our app has several caching layers and performance of these layers show up as several "humps" on the histograms, this helped us track changes to the overall performance as well as each cache layer

Issue #479, opened 2022-05-17
How to parametrize?


dpc:
I want to write a load tester that will test very synthetic loads, and I need it to happen in different configurations. Eg. request A + request B, then request B + request C, then A, B, C combined. Stuff like that. Different combinations, different weights.

I might have missed something but it looks like by default goose will run all the defined scenarios mixed together (possibly by weight, etc.) executed by N users. So I don't really see a way to run workload X first, then workload Y, then Z.

I would be OK with running my binary multiple times, each time with some different arguments. However I'm not sure how to best glue it with command line argument parsing already handled by goose. Ideally I'd like to add `--workload X` command line argument or something like that.

I guess I could use environment variable, and goose will just ignore it, but it's not a best UX: discoverability will be rather poor.

jeremyandrews:
Would the request in https://github.com/tag1consulting/goose/issues/472 solve what you're asking for? I believe it solves the same ask. (It's not yet implemented, but coming soon.)

dpc:
```
fn get_opts(args: &[String]) -> color_eyre::Result<(Opts, GooseConfiguration)> {
    let opts = opts::Opts::parse_args_default(&args[1..])?;
    let goose_opts = GooseConfiguration::parse_args(&opts.free, ParsingStyle::default())?;

    if !opts.help_requested() && !opts.help_goose && !(opts.mainnet ^ opts.testnet) {
        bail!("{}: Must set one of: `--mainnet` or `--testnet`", args[0]);
    }

    Ok((opts, goose_opts))
}

#[tokio::main]
async fn main() -> color_eyre::Result<()> {
    let args = std::env::args().collect::<Vec<_>>();
    let (opts, goose_opts) = match get_opts(&args) {
        Err(e) => {
            eprintln!("{}: {}", args[0], e);
            std::process::exit(2);
        }
        Ok(opts) => opts,
    };

    if opts.help_requested() {
        eprintln!("Usage: {} [OPTIONS]", args[0]);
        eprintln!();
        eprintln!("{}", opts.self_usage());
        return Ok(());
    }

    if opts.help_goose {
        eprintln!("Usage: {} [OPTIONS]", args[0]);
        eprintln!();
        eprintln!("{}", goose_opts.self_usage());
        return Ok(());
    }
...
```

That's the pattern I used so far. Let me check the #472

dpc:
[...]

Issue #472, opened 2022-05-09
Provide a method of running only a subset of scenarios


kkovaacs:
We have a use-case where a load-test contains a lot of scenarios. Occasionally we would like to run only _some_ (one or more) of those scenarios and it would be great if this could be accomplished without commenting out the rest and recompiling the load test program.

One can already list the available scenarios/transactions with the `-l` command line argument. Could `goose` maybe provide a way to run only a subset of the scenarios?
For example, it could accept an optional list of scenario names on the command line and just run those if specified?

jeremyandrews:
I would also find this interesting/useful. It would be important that the ASCII and HTML reports properly reflect if a subset of Scenarios were enabled for a given load test.

So to enable a subset of the Scenarios from the Umami example (for example) perhaps you'd simply do something like:
```bash
---scenarios "Anonymous Spanish user, Admin user "
```

...which would effectively disable the "Anonymous English user" Scenario. Is this the sort of functionality you're looking for?

It would still respect weighting etc of any scenarios enabled, but in this way you could select a subset of scenarios.

jeremyandrews:
Referring to the `-l` option you mentioned, here's the current output for the [umami example](https://book.goose.rs/example/umami.html):

```bash
% cargo run --release --example umami -- -l 
    Finished release [optimized] target(s) in 0.11s
     Running `target/release/examples/umami -l`
Available transactions:
 - Anonymous English user (weight: 40)
    o anon / (weight: 2)
    o anon /en/basicpage (weight: 1)
    o anon /en/articles/ (weight: 1)
    o anon /en/articles/% (weight: 2)
    o anon /en/recipes/ (weight: 1)
    o anon /en/recipes/% (weight: 4)
    o anon /node/%nid (weight: 1)
    o anon /en term (weight: 2)
    o anon /en/search (weight: 1)
    o anon /en/contact (weight: 1)
 - Anonymous Spanish user (weight: 9)
    o anon /es/ (weight: 2)
    o anon /es/basicpage (weight: 1)
    o anon /es/articles/ (weight: 1)
    o anon /es/articles/% (weight: 2)
    o anon /es/recipes/ (weight: 1)
    o anon /es/recipes/% (weight: 4)
    o anon /es term (weight: 2)
    o anon /es/search (weight: 1)
    o anon /es/contact (weight: 1)
 - Admin user (weight: 1)
    o auth /en/user/login (weight: 1)
    o auth / (weight: 2)
[...]

PR #505, opened 2022-06-24
run-time option to configure which scenarios run

jeremyandrews:
- NOTE: depends on #506 being merged fixed (to address clippy errors) -- I've rebased on top of this PR

 - assign a unique machine name for each scenario
 - introduce `--scenarios` to optionally specify which scenarios run during a load test
 - introduce `--scenarios-list` to list scenarios by machine name
 - introduce `GooseDefault::Scenarios`
 - test coverage
 - update `test/common.rs::build_load_test()` to support multiple scenarios (necessary to test this feature)
 - add book page titled "Limiting Which Scenarios Run" that provides complete details on this new feature
 - closes https://github.com/tag1consulting/goose/issues/479
 - closes https://github.com/tag1consulting/goose/issues/472

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -671,14 +671,11 @@ impl GooseAttack {
             return true;
         }
 
-        if scenario
+        // Returns true or false depending on if the machine name is included in the
+        // configured `--scenarios`.
+        scenario
             .machine_name
             .contains(&self.configuration.scenarios)
-        {
-            true
-        } else {
-            false
-        }
     }
 
     /// Use configured GooseScheduler to build out a properly weighted list of
@@ -690,7 +687,7 @@ impl GooseAttack {
         let mut u: usize = 0;
         let mut v: usize;
         for scenario in &self.scenarios {
-            if self.scenario_is_active(&scenario) {
+            if self.scenario_is_active(scenario) {
                 if u == 0 {
                     u = scenario.weight;
                 } else {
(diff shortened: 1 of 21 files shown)


src/lib.rs
```diff
@@ -575,6 +577,22 @@ impl GooseAttack {
...
+        // already seen.
+        while !self.scenario_machine_names.insert(machine_name) {
+            // For each conflict increase the counter and try again.
+            conflicts += 1;
+            machine_name = format!("{}_{}", scenario.machine_name, conflicts);
+        }
+        // If there was a conflict, also update the scenario itself.
+        if conflicts > 0 {
+            scenario.machine_name = format!("{}_{}", scenario.machine_name, conflicts);
+        }
```
slashrsm, line 594: How does the ordering work here? Is it possible that the order of scenarios would change with some minor code change? This would then shuffle them and effectively change machine names. Which is not the end of the world, but could cause confusion.

src/test_plan.rs
```diff
@@ -15,7 +15,6 @@ use crate::util;
 use crate::{AttackPhase, GooseAttack, GooseAttackRunState, GooseError};
 
 /// Internal data structure representing a test plan.
-//#[derive(Options, Debug, Clone, Serialize, Deserialize)]
```
slashrsm, line 18: 😄 👏

slashrsm (commented):
Code looks great and I also confirmed that it works as expected.

slashrsm (approved):

PR #509, opened 2022-07-17
[WIP] re-implement Gaggle removing compile-time feature

jeremyandrews:
- move Manager into a thread to provide coordination of Workers
 - allow configurations and control of Manager thread via Controller

TODO:
 - move Worker into a thread
 - allow both Manager and Worker thread on same goose instance
 - implement actual Gaggle commands through Controller to allow configuring/starting/stopping Gaggle, and for collection of Metrics

PR #510, opened 2022-08-03
Cargo.toml: upgrade dependencies

kkovaacs:
In particular, tokio isn't pinned to 1.15 now. Pinning tokio was making it hard for us to use `goose` in projects which use other dependencies that require a newer version of tokio.


jeremyandrews:
Please see https://github.com/tag1consulting/goose/pull/420 -- there's a known issue preventing this update, as tests fail. The work in  https://github.com/tag1consulting/goose/pull/509 will solve this once it lands.

PR #512, opened 2022-08-17
log proper method

### LINKED ISSUE #511: HTML report doesn't include correct method and path when creating request using builder
First all of thank you for making this tool. :pray: 

version: `0.16.4-dev`

I need to include `header` in POST request so I'm following standard practice of creating a request using builder. The mechanism is working fine but the final report doesn't capture it correctly. Here is the screenshot and code snippet,
```rust
let req_builder = user
    .get_request_builder(&GooseMethod::Post, "/payments")?
    .header("API-KEY", "test")
    .json(&payload);

let request = GooseRequest::builder().set_request_builder(req_builder).build();
```
Screenshot from 2022-08-15 16-01-11

I looked into the code and figured that `GooseRequestBuilder` sets `GET` to method by default. And since I'm using `request_builder` to set header, path and method information, it doesn't reflect in `GooseRequest` value. So after setting a request builder, I'm setting method and path (redundant) as following,
```rust
let req_builder = user
    .get_request_builder(&GooseMethod::Post, "/payments")?
    .header("API-KEY", "test")
    .json(&payload);

let request = GooseRequest::builder()
    .set_request_builder(req_builder)
    .path("/payments")
    .method(GooseMethod::Post)
    .build();
```
Screenshot from 2022-08-15 16-10-16

I _think_ there should be no need to set those values again once we specified in builder, and while calling `set_request_builder` respective state should get updated as per provided argument. As of now, I couldn't find a way to get those information out of request builder. Please guide me if I'm doing anything wrong, or let me know if I can help you to fix this. Thanks!

jeremyandrews:
- use Reqwest's [`method`](https://docs.rs/reqwest/latest/reqwest/struct.Request.html#method.method) function to log proper method, closes #511
- Add test coverage to be confirm that metrics (and therefore logging) are the same whether or not `GooseRequest::builder()` is used

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -1245,6 +1245,7 @@ impl GooseUser {
         // POST request.
         let goose_request = GooseRequest::builder()
             .method(GooseMethod::Post)
+            .path(path)
             .set_request_builder(reqwest_request_builder.body(body))
             .build();
 
@@ -1290,6 +1291,7 @@ impl GooseUser {
         // POST form request.
         let goose_request = GooseRequest::builder()
             .method(GooseMethod::Post)
+            .path(path)
             .set_request_builder(reqwest_request_builder.form(&form))
             .build();
 
@@ -1338,6 +1340,7 @@ impl GooseUser {
         // POST json request.
         let goose_request = GooseRequest::builder()
             .method(GooseMethod::Post)
+            .path(path)
             .set_request_builder(reqwest_request_builder.json(&json))
             .build();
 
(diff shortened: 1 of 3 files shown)


src/goose.rs
```diff
@@ -1245,6 +1245,7 @@ impl GooseUser {
         // POST request.
         let goose_request = GooseRequest::builder()
             .method(GooseMethod::Post)
+            .path(path)
```
jeremyandrews, line 1248: We have to call `.path(path)` for consistent naming in the metrics, for this to work the same if you're using the builder directly.

slashrsm:
Looks good to me.

slashrsm (approved):

PR #514, opened 2022-09-01
Fix panic when a empty wait time interval is set

mTsBucy1:
Previously, despite what the [documentation](https://docs.rs/goose/0.16.3/goose/goose/index.html#scenario-wait-time) says ("inclusively between the low and high"), the wait time was generated with the upper bound exclusive. This lead to the problem that a constant wait time (the same upper and lower limit) would panic because there is no valid value for rand to return.
`thread 'tokio-runtime-worker' panicked at 'cannot sample empty range', [left out for privacy]\rand-0.8.5\src\rng.rs:134:9`

This PR fixes this by this by using a closed range to generate the wait time. Additionally, a example in the documentation now reflects this usecase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.16.4-dev
  - [#512](https://github.com/tag1consulting/goose/pull/512) include proper HTTP method and path in logs and html report when using `GooseRequest::builder()`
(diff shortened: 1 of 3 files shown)


jeremyandrews (approved):
Thanks for catching and fixing this!

I've added a record to the CHANGELOG.

PR #516, opened 2022-09-05
Refactor CSV Logging

### LINKED ISSUE #515: Requests-log with csv format gives malformatted files
The CSV request logs generated by goose include a RAW column that does not properly escape quotes causing many csv readers to confuse column headers.

In particluar this is the first two rows of the CSV requests log unittest.
(Running `cargo test --package goose --test logs -- test_requests_logs_csv --exact --nocapture` after commenting out the cleanup of files https://github.com/tag1consulting/goose/blob/main/tests/logs.rs#L316:L322).
```
elapsed,raw,name,final_url,redirected,response_time,status_code,success,update,user,error,coordinated_omission_elapsed,user_cadence
1,"GooseRawRequest { method: Get, url: "http://127.0.0.1:65508/", headers: [], body: "" }","/","http://127.0.0.1:65508/",false,5,200,true,false,0,,0,0
7,"GooseRawRequest { method: Get, url: "http://127.0.0.1:65508/error", headers: [], body: "" }","/error","http://127.0.0.1:65508/error",false,0,503,false,false,0,503 Service Unavailable: /error,0,0
```
Which both Google Spreadsheet and Microsoft Excel interpret as (shortened for clarity)
```
+──────────+───────────────────────────────────────────+──────────────+──────────────+─────────────+
| elapsed  | raw                                       | name         | final_url    | redirected  |
+──────────+───────────────────────────────────────────+──────────────+──────────────+─────────────+
| 1        | GooseRawRequest { method: Get, url: ..."  | headers: []  | body: "" }"  | /           |
| 7        | GooseRawRequest { method: Get, url: ..."  | headers: []  | body: "" }"  | /error      |
+──────────+───────────────────────────────────────────+──────────────+──────────────+─────────────+
```
Which clearly is not desireable, since now the rest of the csv heads are shifted over by two columns.

The solution is to propertly escape quotes ([see this stackoverflow answer](https://stackoverflow.com/a/4617967/19331219)) resulting in this csv output (notice the doubled up quotes in the raw column). This gets property interpreted by both Google Docs and Excel.
```
elapsed,raw,name,final_url,redirected,response_time,status_code,success,update,user,error,coordinated_omission_elapsed,user_cadence
1,"GooseRawRequest { method: Get, url: ""http://127.0.0.1:65508/"", headers: [], body: """" }","/","http://127.0.0.1:65508/",false,5,200,true,false,0,,0,0
7,"GooseRawRequest { method: Get, url: ""http://127.0.0.1:65508/error"", headers: [], body: """" }","/error","http://127.0.0.1:65508/error",false,0,503,false,false,0,503 Service Unavailable: /error,0,0
```

mTsBucy1:
To fix #515 I added a macro that properly escapes the values.
I like this approach because it is very lightweight and doesn't add any csv reader functionality as many external dependencies would. On the other hand, this still struggles with nested data and Option<> as well as hard coded headers.

However, currently in order to use the macro in the tests, I use macro_export, but then this leaks into the docs. Does anyone know how to prevent this in Edition 2018? Or is this a case for docs::hidden?

I additionally propose to flatten the GooseRequestMetric.raw into its fields method, url, headers, body for the request log and error log. For GooseDebug, which includes a whole GooseRequestMetric, a similar argument applies.
The Option<> values should further be represented by an empty string when None.
Since those changes potentially break external metric analysis via CSV import, I am reluctant to make these changes without further input.
What do you think? Do you import goose metrics in another program? And via csv or other?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.16.4-dev
  - [#512](https://github.com/tag1consulting/goose/pull/512) include proper HTTP method and path in logs and html report when using `GooseRequest::builder()`
  - [#514](https://github.com/tag1consulting/goose/pull/514) fix panic when an empty wait time interval is set
+ - [#516](https://github.com/tag1consulting/goose/pull/516) fix unescaped inner quotes in csv logs
 
 ## 0.16.3 July 17, 2022
  - [#498](https://github.com/tag1consulting/goose/issues/498) ignore `GooseDefault::Host` if set to an empty string
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -160,6 +160,31 @@ pub(crate) type GooseLoggerJoinHandle =
 /// Optional unbounded sender from all GooseUsers to logger thread, if enabled.
 pub(crate) type GooseLoggerTx = Option<flume::Sender<Option<GooseLog>>>;
 
+/// Formats comma separated arguments into a csv row according to RFC 4180. Every argument has to be `Display`.
+///
+/// Specifically, this encloses all values with double quotes `"` which contain a comma, a quote or a new line.
+/// Inner quotes are doubled according to RFC 4180 2.7.
+/// The fields are joined by commas `,`, but *not* terminated with a line ending.
+#[macro_export]
+macro_rules! format_csv_row {
+    ($( $field:expr ),+ $(,)?) => {{
+        [$( $field.to_string() ),*]
+            .iter()
+            .map(|s| {
+                if s.contains('"') || s.contains(',') || s.contains('\n') {
+                    // Enclose in quotes and escape inner quotes
+                    format!("\"{}\"", s.replace('"', "\"\""))
+                } else {
+                    // Because into_iter is not available in edition 2018
+                    s.clone()
+                }
+            })
+            .collect::<Vec<String>>()
+            .join(",")
+    }};
+}
+pub use format_csv_row;
+
 /// If enabled, the logger thread can accept any of the following types of messages, and will
 /// write them to the correct log file.
 #[derive(Debug, Deserialize, Serialize)]
@@ -214,15 +239,12 @@ impl FromStr for GooseLogFormat {
 
 // @TODO this should be automatically derived from the structure.
 fn debug_csv_header() -> String {
-    // No quotes needed in header.
-    format!("{},{},{},{}", "tag", "request", "header", "body")
+    format_csv_row!("tag", "request", "header", "body")
 }
 
 // @TODO this should be automatically derived from the structure.
 fn error_csv_header() -> String {
-    // No quotes needed in header.
-    format!(
-        "{},{},{},{},{},{},{},{},{}",
+    format_csv_row!(
         "elapsed",
         "raw",
         "name",
@@ -237,9 +259,7 @@ fn error_csv_header() -> String {
 
 // @TODO this should be automatically derived from the structure.
 fn requests_csv_header() -> String {
-    // No quotes needed in header.
-    format!(
-        "{},{},{},{},{},{},{},{},{},{},{},{},{}",
+    format_csv_row!(
         "elapsed",
         "raw",
         "name",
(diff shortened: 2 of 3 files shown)


jeremyandrews (changes requested):
>  I like this approach because it is very lightweight and doesn't add any csv reader functionality as many external dependencies would. On the other hand, this still struggles with nested data and Option<> as well as hard coded headers.

In my local testing, it fixes the bug you reported and I don't see any regressions. Where are you running into problems with nested data and Option<>?

> However, currently in order to use the macro in the tests, I use macro_export, but then this leaks into the docs. Does anyone know how to prevent this in Edition 2018? Or is this a case for docs::hidden?

Doing some searches, I think this is indeed the correct solution (and confirmed locally it works as expected):
https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#hidden

> I additionally propose to flatten the GooseRequestMetric.raw into its fields method, url, headers, body for the request log and error log. For GooseDebug, which includes a whole GooseRequestMetric, a similar argument applies.

I'm not sure I follow what you're suggesting. Are you proposing replacing the "raw" column with a column for each of the 4 contained fields? I don't see any fundamental problem with doing that.

> Since those changes potentially break external metric analysis via CSV import, I am reluctant to make these changes without further input.

I'd prefer to get this fix in first, before making further changes. Changing the format of the CSV export would require a version bump (ie going to 0.17).

Pleae do add an entry to the Changelog.

> What do you think? Do you import goose metrics in another program? And via csv or other?

I use the json format, and do processing of fields on the command line with [jq](https://stedolan.github.io/jq/) (see [this example](https://book.goose.rs/getting-started/tips.html#debugging-html-responses) for example).

src/logger.rs
```diff
@@ -160,6 +160,31 @@ pub(crate) type GooseLoggerJoinHandle =
 /// Optional unbounded sender from all GooseUsers to logger thread, if enabled.
 pub(crate) type GooseLoggerTx = Option<flume::Sender<Option<GooseLog>>>;
 
+/// Formats comma separated arguments into a csv row according to RFC 4180. Every argument has to be `Display`.
+///
+/// Specifically, this encloses all values with double quotes `"` which contain a comma, a quote or a new line.
+/// Inner quotes are doubled according to RFC 4180 2.7.
+/// The fields are joined by commas `,`, but *not* terminated with a line ending.
```
jeremyandrews, line 167: Thanks for catching that this ends up in the docs. Please hide:

`#[doc(hidden)]`

tests/logs.rs
```diff
@@ -741,3 +741,12 @@ async fn test_all_logs_pretty() {
 async fn test_all_logs_pretty_gaggle() {
     run_gaggle_test(TestType::All, "pretty").await;
 }
+
+#[test]
```
jeremyandrews, line 745: 👍 Thanks for adding test coverage, too!

mTsBucy1:
> > I like this approach because it is very lightweight and doesn't add any csv reader functionality as many external dependencies would. On the other hand, this still struggles with nested data and Option<> as well as hard coded headers.
> 
> In my local testing, it fixes the bug you reported and I don't see any regressions. Where are you running into problems with nested data and Option<>?

There is no problem per se, but the existing [TODO](https://github.com/tag1consulting/goose/pull/516/files#diff-6658812715089c8241f20814aaf257dbd94a5448e01cf0010a0a2da54406f844L309) hasn't been addressed. Specifically, the csv still contains "Some(value)" and "None", instead of the more CSV-style empty string and "value".

> > I additionally propose to flatten the GooseRequestMetric.raw into its fields method, url, headers, body for the request log and error log. For GooseDebug, which includes a whole GooseRequestMetric, a similar argument applies.
> 
> I'm not sure I follow what you're suggesting. Are you proposing replacing the "raw" column with a column for each of the 4 contained fields? I don't see any fundamental problem with doing that.

Yes, that is exactly what I meant. It feels wrong to have json-like fields in a csv file. But that is material for a future PR.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.16.4-dev
  - [#512](https://github.com/tag1consulting/goose/pull/512) include proper HTTP method and path in logs and html report when using `GooseRequest::builder()`
  - [#514](https://github.com/tag1consulting/goose/pull/514) fix panic when an empty wait time interval is set
+ - [#514](https://github.com/tag1consulting/goose/pull/515) fix unescaped inner quotes in csv logs
```
jeremyandrews, line 6: this should be 516 in both places

jeremyandrews (approved):

jeremyandrews:
Thanks, this is a much appreciated improvement: confirmed that CSV import works correctly now. Waiting for tests to pass then will merge.

The suggested followups would be very welcome if you wish to submit additional PRs, both expanding `GooseRawRequest` to proper columns, and properly formatting `Option<>`. ///

PR #519, opened 2022-09-14
Impl `Default` for `GooseConfiguration`

### LINKED ISSUE #518: Impl `Default` for `GooseConfiguration`
Hello,

I'm trying to use Goose in a binary that itself has some command line options. The issue I ran into is that when I do `GooseAttack::initialize` it will read the command line arguments of my binary and they are not recognized, causing the program to exit. I can use `GooseAttack::initialize_with_config` instead, so that Goose doesn't read the command line args. The problem is that the `GooseConfiguration` object is very large. It would be nice if it had a `Default` impl so that I can send that to `GooseAttack::initialize_with_config`.

yds12:
Fix #518


LionsAd (approved):
RTBM - Looks good to me

jeremyandrews:
Thanks. I also added a line to the Changelog.

If you want to submit a followup PR, it would be great to document how you use this in https://book.goose.rs/getting-started/custom.html, as an alternative to what's currently documented there.

yds12:
Hello! I would be glad to submit a PR for the docs.

PR #520, opened 2022-09-14
Add docs about custom command line args

yds12:
Follow up on #519. As suggested by @jeremyandrews, extending the docs to cover the possibility of using custom command line arguments.


jeremyandrews (approved):
Thanks for taking the time to document this! Looks great, and I'm sure many will find this a useful alternative for custom configuration.

PR #522, opened 2022-09-15
Display times on the report in local time.

### LINKED ISSUE #521: In HTML report, the time of plan review is  UTC but not localtime
image

As is shown above, the top of the picture is the plan review table, in which the time phases are using UTC, while the local time is used in the request metrics chart. I think the time phases of a plan review should use the local time too.

slashrsm:
Closes #521


jeremyandrews (approved):
Thanks! I test before and after, and confirmed this correctly fixes the reported issue.

PR #543, opened 2023-05-01
remove dependency on num_cpus

jeremyandrews:
- remove external dependency on [`num_cpus()`](https://docs.rs/num_cpus/latest/num_cpus/), use instead built-in [`available_parallelism()`](https://doc.rust-lang.org/stable/std/thread/fn.available_parallelism.html) added in [rust 1.59.0](https://blog.rust-lang.org/2022/02/24/Rust-1.59.0.html)
 - we originally listed requiring a much earlier version of Rust, but Goose wasn't correctly compiling with such an old Rust; bumping up our dependency to something more modern


bebehei:
/ This might totally be unrelated, I'm new to the rust ecosystem /

I tried to build goose yesterday on NixOS and it already required rust version 1.65 because of the dependency on the time crate. See here:

```text
bebe@gulasch:~/goose$ cargo build --example simple
    Updating crates.io index
  Downloaded rustls-pemfile v1.0.2
  Downloaded hyper-rustls v0.23.2
  Downloaded 2 crates (50.3 KB) in 0.77s
error: package `time v0.3.21` cannot be built because it requires rustc 1.65.0 or newer, while the currently active rustc version is 1.64.0
```

Wouldn't it be appropriate to even tag goose to have rustc version 1.65?

Issue #366, opened 2021-09-28
Latency metrics resolution (microseconds)


zezic:
Hi! In our company we started using Goose to run various performance tests and one of them is the latency test. We need to control the metrics after changes in our code to detect any performance degradation.
But here is the problem – the latency of some endpoints of our service is about ~2000 **micro**seconds. Currently we use custom loadtest function to report latency as microseconds instead of milliseconds, but 

According to this procedure Goose rounds numbers by 1000 when they reach values of 1000: https://github.com/tag1consulting/goose/blob/dd8144646ae682128a9fe46decb3be6f27eb0717/src/metrics.rs#L539-L553

It makes it impossible do detect difference between, lets say, 2200 microseconds and 2500 microseconds.

Is it possible to add options to Goose to make it support measuring latency in microseconds with more precise (configurable?) rounding?

jeremyandrews:
I'm definitely open to improvements here. Early versions of Goose did not do any of this rounding, however this lead to massive data structures and horrible performance when displaying running and final metrics. I'm happy to review a PR that works to make this more flexible.

jeremyandrews:
Goose is currently tracking the response time of each request in milliseconds:
https://github.com/tag1consulting/goose/blob/main/src/metrics.rs#L305
https://github.com/tag1consulting/goose/blob/main/src/metrics.rs#L352

It also logs when the request was made with millisecond granularity:
https://github.com/tag1consulting/goose/blob/main/src/metrics.rs#L295
https://github.com/tag1consulting/goose/blob/main/src/goose.rs#L1606

Perhaps `GooseRequestMetric` and `GooseRequestMetricTimingData` could be turned into generics with a default implementation using milliseconds, allowing you to provide your own microsecond implementations. Some thought would need to go into how this would work.

nathanielc:
I looking into using a compressed histogram format like https://docs.rs/tdigest/latest/tdigest/index.html to track large numbers of request accurately without using much space. Would something like this be accepted? 

I could look into to making a generic aggregate type to allow for multiple implementations.

jeremyandrews:
Yes, if performance isn't negatively impacted this is definitely interesting. Making it generic to allow multiple implementations isn't necessary but would certainly be appreciated.

PR #546, opened 2023-05-09
Feat use tdigest for metrics 

nathanielc:
This change replaces this existing metrics collections with a tdigest implementation. This is beneficial for a few reasons:

1. It is bounded in the amount of memory space it uses
2. It is a statistically sound approach to approximating quantiles of a distribution
3. Tdigest is serializable and multiple tdigests can be merged into a single tdigest making it well suited to gaggle use cases.

This means we can get more accurate quantiles without having to use more memory that we currently do.

The previous logic would round values in order to save space, tdigest does something similar but does so in a way that minimizes the accumulated error introduced from the rounding. Instead of rounding values to predetermined values it rounds values to the closest existing point already in the distribution and only rounds when there is a need (i.e. its reached its space limit). 

This PR is still WIP but wanted to start the conversation around the contribution before continuing:

TODO:

* Update all tests
* Update all places that previously used BTreeMap<usize, usize> to use tdigest
* Rebase against main branch.  We use gaggle so this PR is a branch off  0.16.4, not really sure how much work it would need to target main. 

This PR addresses part of the issue #366

As an side we are using this branch successfully already getting good quantiles from our scenarios runs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,19 +27,18 @@ num_cpus = "1.0"
 num-format = "0.4"
 rand = "0.8"
 regex = "1"
-reqwest = { version = "0.11",  default-features = false, features = [
+reqwest = { version = "0.11", default-features = false, features = [
     "cookies",
     "gzip",
     "json",
 ] }
-serde = { version = "1.0", features = [
-    "derive",
-] }
+serde = { version = "1.0", features = ["derive"] }
 serde_cbor = "0.11"
 serde_json = "1.0"
 simplelog = "0.10"
 strum = "0.24"
 strum_macros = "0.24"
+tdigest = { version = "0.2.3", features = ["serde", "use_serde"] }
 tokio = { version = "~1.15.0", features = [
     "fs",
     "io-util",
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -650,6 +650,9 @@ pub(crate) async fn manager_main(mut goose_attack: GooseAttack) -> GooseAttack {
     // Run any configured test_stop() functions.
     goose_attack.run_test_stop().await.unwrap();
 
+    // Mark manager metrics as final.
+    goose_attack.metrics.final_metrics = true;
+
     goose_attack
 }
 
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -5,6 +5,7 @@ use std::collections::HashSet;
 use std::io::BufWriter;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time;
+use tdigest::TDigest;
 
 use crate::metrics::{
     self, GooseErrorMetricAggregate, GooseErrorMetrics, GooseRequestMetricAggregate,
@@ -126,24 +127,10 @@ fn merge_requests_from_worker(
     // Make a mutable copy where we can merge things
     let mut merged_request = parent_request.clone();
     // Iterate over user response times, and merge into global response time
-    merged_request.raw_data.times = metrics::merge_times(
-        merged_request.raw_data.times,
-        user_request.raw_data.times.clone(),
-    );
-    // Increment total response time counter.
-    merged_request.raw_data.total_time += &user_request.raw_data.total_time;
-    // Increment count of how many response counters we've seen.
-    merged_request.raw_data.counter += &user_request.raw_data.counter;
-    // If user had new fastest response time, update global fastest response time.
-    merged_request.raw_data.minimum_time = metrics::update_min_time(
-        merged_request.raw_data.minimum_time,
-        user_request.raw_data.minimum_time,
-    );
-    // If user had new slowest response time, update global slowest response time.
-    merged_request.raw_data.maximum_time = metrics::update_max_time(
-        merged_request.raw_data.maximum_time,
-        user_request.raw_data.maximum_time,
-    );
+    merged_request.raw_data.histogram = TDigest::merge_digests(vec![
+        merged_request.raw_data.histogram,
+        user_request.raw_data.histogram.clone(),
+    ]);
     // Increment total success counter.
     merged_request.success_count += &user_request.success_count;
     // Increment total fail counter.
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -32,6 +32,7 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fmt::Write;
 use std::str::FromStr;
 use std::{f32, fmt};
+use tdigest::TDigest;
 use tokio::io::AsyncWriteExt;
 
 /// Used to send metrics from [`GooseUser`](../goose/struct.GooseUser.html) threads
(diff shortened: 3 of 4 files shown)


Cargo.toml
jeremyandrews, line 1: In order to be able to merge this PR, it will need to be against the latest development version of the codebase. Please rebase. (I understand you desire the use of Gaggles, there is work underway to add them back in 0.18 (more likely to be called 1.0).

Cargo.toml
```diff
@@ -27,19 +27,18 @@ num_cpus = "1.0"
...
-serde = { version = "1.0", features = [
-    "derive",
-] }
+serde = { version = "1.0", features = ["derive"] }
 serde_cbor = "0.11"
 serde_json = "1.0"
 simplelog = "0.10"
 strum = "0.24"
 strum_macros = "0.24"
+tdigest = { version = "0.2.3", features = ["serde", "use_serde"] }
```
jeremyandrews, line 41: It looks like this library hasn't been updated in 2 years, is there anything that has replaced it? Does it matter that it's not being updated?
https://github.com/MnO2/t-digest

src/manager.rs
jeremyandrews, line 1: This file is not in the current development branch ...

jeremyandrews (changes requested):
Overall this looks like a positive step forward, but a few concerns:
 - in order to merge it in, we'll need a patch against the latest development release (yes, this temporarily does not have gaggle support)
 - are there any concerns with tdigest not having commits for a couple of years?
 - we'll need to run a 2-3 day load test and confirm that this new data structure continues to scale well without slowing anything down

Cargo.toml
nathanielc: Good question, there is another implementation here https://docs.rs/pdatastructs/latest/pdatastructs/tdigest/struct.TDigest.html however that doesn't implement serde serialize/deserialize which we would need. 

I am not concerned about the crate not getting updates as this is the kind of things that once implemented correctly doesn't need ongoing maintanence. However if you would prefer I can change out the implementation. 

Or as we discussed I could make the implementation generic and then users can bring their own solution.

Cargo.toml
nathanielc: I'll rebase the work doesn't look to be too much work.

nathanielc:
>we'll need to run a 2-3 day load test and confirm that this new data structure continues to scale well without slowing anything down

This sounds great, there is a parameter we can tune to adjust performance. When creating the tdigest you provide it with a max size, I use 100 a reasonable default, but we could potentially change it as needed. A lower value will mean less allocations an likely better performance as a result.

nathanielc:
Thanks for the hints on the failing tests. I still haven't fixed them yet but I have learned a few things:

1. The change to tdigest has caused several failures because its serialized format is large. (overflows the buffer)
    - Removing the `metrics-json` test bits is not enough. For example the step in the test that tries to set the `host` an expects a failure fails because the metrics are returned in json and it breaks the message passing logic.
    - I expect other code paths will also break 
2. Its not possible to change the serialization format to something more compact because the tdigest crate does not expose the `centroid` data directly. 

I see a few ways forward.

1. Submit a PR to the tdigest crate to improve its serialization format
2. Change to the t-digest implementation to this other crate  https://docs.rs/pdatastructs/latest/pdatastructs/tdigest/struct.TDigest.html That crate does not have a serialization format currently however they have an open issue for it https://github.com/crepererum/pdatastructs.rs/issues/61 and our implementation could ensure its compact enough.
3. Change something in Goose to allow for larger data transfer. I haven't dug into the message passing logic much yet, but my assumption is we have a fixed size buffer to keep the packet size low for telnet? Maybe there is something we can do here?
4. Something else?

My vote is option 2 as it seems like the best long term solution since that crate is more active and more likely to accept the change. 

Why is the format so large?

Under the hood a t-digest is a set of centroids, each centroid is two numeric values (i.e. the mean and the weight). With the max_size set to 100 the tdigest will store at most 100 centroids. However each centroid is serialized as `{"mean": 1.23456789, "weight": 1.23456789}`, that repeated 100x is not efficient. A more efficient implementation would be store serialize a list of means separate from a list of weights and zip them back up at serialization time, i.e `{"means": [1, 2, 3, 4, 5], "weights": [6,7,8,9,0]}`. This will save a significant amount of bytes in the serialized format. However do we still expect that two lists of 100 numbers each will still be small enough? If not we might need to consider other options like 3 to allow for larger data transfers.

The trade off with tdigest is that is guaranteed to never use more than 100 centroids where as the previous implementation was technically unbounded, however tdigest is much more likely to use all 100 pairs. Meaning if we can get the logic working for 100 centroids we know it will always work. Additionally the value 100 was choose somewhat arbitrarily,  we could choose a smaller value if needed. 

Thanks for working through these details with me. Happy to go in whichever direction.

jeremyandrews:
The size of the tdigest serialization format shouldn't matter if we remove `MetricsJson` from the tests... The only reason I'd see wanting to reduce the size of this object is if it's growing too large for Goose to sustain multi-day load tests. Have you tried extended load tests and seen how large the metrics grow?

To remove from the tests, I was thinking something like this patch:
```
diff --git a/tests/controller.rs b/tests/controller.rs
index a66ebc3..ddacd0d 100644
--- a/tests/controller.rs
+++ b/tests/controller.rs
@@ -560,19 +560,8 @@ async fn run_standalone_test(test_type: TestType) {
                     }
                 }
                 ControllerCommand::MetricsJson => {
-                    match test_state.step {
-                        // Request the running metrics in json format.
-                        0 => {
-                            make_request(&mut test_state, "metrics-json\r\n").await;
-                        }
-                        // Confirm the metrics are returned in json format.
-                        _ => {
-                            assert!(response.starts_with(r#"{"hash":0,"#));
-
-                            // Move onto the next command.
-                            test_state = update_state(Some(test_state), &test_type).await;
-                        }
-                    }
+                    // Move onto the next command.
+                    test_state = update_state(Some(test_state), &test_type).await;
                 }
                 ControllerCommand::Start => {
                     match test_state.step {
@@ -715,9 +704,9 @@ async fn update_state(test_state: Option<TestState>, test_type: &TestType) -> Te
         ControllerCommand::Config,
         ControllerCommand::ConfigJson,
         ControllerCommand::Metrics,
-        ControllerCommand::MetricsJson,
         ControllerCommand::Stop,
         ControllerCommand::Shutdown,
+        ControllerCommand::MetricsJson,
     ];
 
     if let Some(mut state) = test_state {
```

By moving `::MetricsJson` to after `::Shutdown` it effectively is disabled. (Something more/different will need to be done to address the WebSocket tests: I was looking at the Telnet tests for now.)

That said, now there's an (apparently) unrelated problem in which users aren't starting quickly enough and so we instead fail with the following error:

```
thread 'test_telnet_controller' panicked at 'assertion failed: goose_metrics.total_users == MAX_USERS', tests/controller.rs:142:5
```

At a quick glance I'm not sure why this would happen, as `--hatch-rate` seems to correctly be set to `25` and it's sleeping 1 second which should allow all 20 users to hatch. Did you already rebase against `main` where tests are working?

nathanielc:
Thanks for the diff, I did something similar and got past that part of the test failures. 

> Did you already rebase against main where tests are working?

Yes, and I ran into the same issue with `MAX_USERS`. I traced the issue down to these lines https://github.com/tag1consulting/goose/blob/main/tests/controller.rs#L608-L621 The assertion on line 614 fails and so the `user` count is never changed to `MAX_USERS`. The reason the assertion fails has something to do with the large size of the metrics in json format. When I printed out the response it was a bunch of JSON about the means and weights. I am not sure why the response to the `host` command contains JSON of the metrics but it did and that broke the test system.

PR #557, opened 2023-08-21
Speed up user initialization

Michael-Hollister:
When running goose tests with large amounts of goose users on my machine, it takes roughly 1 minute to initialize 1000 goose users. It appears that building each reqwest client for each goose user is an expensive operation. If we change the behavior to build only 1 reqwest client, and clone the object when creating the goose users, we can significantly increase initialization performance. With this change, I can create a few million goose users and only takes a couple of seconds to complete the initialization phase on my machine.

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -872,26 +872,14 @@ impl GooseUser {
         base_url: Url,
         configuration: &GooseConfiguration,
         load_test_hash: u64,
+        reqwest_client: Option<Client>,
     ) -> Result<Self, GooseError> {
         trace!("new GooseUser");
 
-        // Either use manually configured timeout, or default.
-        let timeout = if configuration.timeout.is_some() {
-            match crate::util::get_float_from_string(configuration.timeout.clone()) {
-                Some(f) => f as u64 * 1_000,
-                None => GOOSE_REQUEST_TIMEOUT,
-            }
-        } else {
-            GOOSE_REQUEST_TIMEOUT
+        let client = match reqwest_client {
+            Some(c) => c,
+            None => create_reqwest_client(configuration)?,
         };
-
-        let client = Client::builder()
-            .user_agent(APP_USER_AGENT)
-            .cookie_store(true)
-            .timeout(Duration::from_millis(timeout))
-            // Enable gzip unless `--no-gzip` flag is enabled.
-            .gzip(!configuration.no_gzip)
-            .build()?;
 
         Ok(GooseUser {
             started: Instant::now(),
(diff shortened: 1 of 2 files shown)


jeremyandrews:
Wow, that’s indeed an impressive optimization. At a first glance and with minimal testing it looks great. I’ll look a bit closer shortly, and likely will merge as is. Thanks!

jeremyandrews:
I've done a bit more testing.

On Linux, this is a massive improvement. Without the patch, trying to start too many users not only takes a very long time, but it uses very large amounts of memory. With the patch, it launches very quickly.

On Mac OS X (Silicon) there's no big difference with or without the patch. I'm still trying to understand why, but even if it's a Linux-only improvement it's still a nice improvement.

Michael-Hollister:
Ah, that is interesting. Yeah, I am primarily using Linux for running my tests. I've also tested this on Windows, and I observe similar behavior you see on Mac OS X, where user initialization is very quick without the patch and not taking up a significant amount of memory.

I also tried using the `RustLS` TLS bindings on Linux (https://book.goose.rs/config/rustls.html#rustls), but seems like that had no effect with the un-patched goose version.

jeremyandrews (approved):
This is a significant win on starting a load test on Linux. It doesn't impact the speed of starting a load test on Mac OS X, but it does consistently improve the reliability of large load tests. I hope to keep digging into why, but as there's no regressions and only improvements happy to commit as is.

jeremyandrews:
Thanks for the clean patch and impressive optimization contribution!

jeremyandrews:
Unfortunately this optimization has to be reverted. It results in the CookieJar and Headers to be shared between all clients: while this may not matter in some circumstances, it's not a good default. See this PR where I've added tests and reverted the `.clone()` optimization:
https://github.com/tag1consulting/goose/pull/575

I'd welcome a new PR if it wraps this optimization in a configuration option, so you'd have to intentionally disable cookie/header support for this quick startup.

Michael-Hollister:
Sure, I understand and thanks for letting me know. All of my Goose usage is done on Linux with large numbers of goose users, so this is essential for my use-case. I'll work on submitting a new PR making this a configuration option.

jeremyandrews:
@Michael-Hollister I took a stab at one possible solution, can you give it a try?
https://github.com/tag1consulting/goose/pull/578

I was originally thinking it could be a run-time option, but figured it was better to not even compile the related Reqwest code if it's not being used.

Michael-Hollister:
> @Michael-Hollister I took a stab at one possible solution, can you give it a try? #578
> 
> I was originally thinking it could be a run-time option, but figured it was better to not even compile the related Reqwest code if it's not being used.

Thanks @jeremyandrews! Tested the branch and the cargo feature approach works fine for me.

Issue #558, opened 2023-08-25
Potential segfault in the time crate


s-tar:
Hey! 

We use goose in our project and during `cargo deny check` we found out that goose use chrono dependency, which has security vulnerability. Here is detailed description:  https://rustsec.org/advisories/RUSTSEC-2020-0071

Solution is very simple: 
Update dependency for chrono to next
`chrono = { version = "0.4", default-features = false, features = ["clock"] }`

Do you mind to do that? Thanks.

Regards, Sergii.

PR #559, opened 2023-08-25
disable default chrono features, fix potential segfault in the time crate

jeremyandrews:
- fixes https://github.com/tag1consulting/goose/issues/558 
 - only enable required features from `chrono`, addressing https://rustsec.org/advisories/RUSTSEC-2020-0071

> A possible workaround for crates affected through the transitive dependency in chrono, is to avoid using the default oldtime feature dependency of the chrono crate by disabling its default-features and manually specifying the required features instead.

- source of the fix: https://github.com/chronotope/chrono/issues/602#issuecomment-1242149249

This fixes the following issue detailed [at the link above](https://rustsec.org/advisories/RUSTSEC-2020-0071):
> Unix-like operating systems may segfault due to dereferencing a dangling pointer in specific circumstances. This requires an environment variable to be set in a different thread than the affected functions. This may occur without the user's knowledge, notably in a third-party library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.17.2-dev
  - [#557](https://github.com/tag1consulting/goose/pull/557) speed up user initialization on Linux
+ - [#559](https://github.com/tag1consulting/goose/pull/559) disable unnecessary features in chronos, avoid potential segfault in time crate: https://rustsec.org/advisories/RUSTSEC-2020-0071
 
 ## 0.17.1 August 17, 2023
  - [#543](https://github.com/tag1consulting/goose/pull/543) remove external dependency on num_cpus(), use instead built-in available_parallelism() added in rust 1.59.0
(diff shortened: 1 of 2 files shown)


jeremyandrews:
@LionsAd I don't see the need to declare this a security release, worst-case it seems anther user on the server running the load test could cause the load test to segfault. In general, it would be unwise to run a load test on a shared server anyway. If you agree, I'll go ahead and merge as-is.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.17.2-dev
  - [#557](https://github.com/tag1consulting/goose/pull/557) speed up user initialization on Linux
+ - []() disable unnecessary features in chronos, avoid potential segfault in time crate: https://rustsec.org/advisories/RUSTSEC-2020-0071
```
slashrsm, line 5: Link missing.

slashrsm (changes requested):

CHANGELOG.md
jeremyandrews: Oh my! Thank you :)

slashrsm (approved):

PR #565, opened 2023-09-28
allow to skip cert validation for internal test situations

raffaeleragni:
Adding the possibility to setup Reqwest client with ignore certificate mode.

This feature is used in cases https is used for non terminated proxies, but when testing internal infrastructure setups eg. when self signed or dev certificates are being used.

This is what I currently need to be able to use goose in the setup I have to load test.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -195,6 +195,9 @@ pub struct GooseConfiguration {
     /// Follows base_url redirect with subsequent requests
     #[options(no_short)]
     pub sticky_follow: bool,
+    /// Allows to test internally when https validation is not required
+    #[options(no_short)]
+    pub accept_invalid_certs: bool,
 }
 
 /// Optionally defines a subset of active Scenarios to run during a load test.
@@ -332,6 +335,8 @@ pub(crate) struct GooseDefaults {
     pub websocket_host: Option<String>,
     /// An optional default for port WebSocket Controller listens on.
     pub websocket_port: Option<u16>,
+    /// Allow for internal testing when no certificates are available but still using https
+    pub accept_invalid_certs: Option<bool>,
 }
 
 /// Defines all [`GooseConfiguration`] options that can be programmatically configured with
@@ -432,6 +437,7 @@ pub enum GooseDefault {
     WebSocketHost,
     /// An optional default for port WebSocket Controller listens on.
     WebSocketPort,
+    AcceptInvalidCerts,
 }
 
 /// Most run-time options can be programmatically configured with custom defaults.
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -195,6 +195,9 @@ pub struct GooseConfiguration {
     /// Follows base_url redirect with subsequent requests
     #[options(no_short)]
     pub sticky_follow: bool,
+    /// Allows to test internally when https validation is not required
```
jeremyandrews, line 198: How about documenting this as:
`Disables validation of https certificates`

We can better document in the goose book (and it can happen in a followup PR).

src/config.rs
```diff
@@ -332,6 +335,8 @@ pub(crate) struct GooseDefaults {
     pub websocket_host: Option<String>,
     /// An optional default for port WebSocket Controller listens on.
     pub websocket_port: Option<u16>,
+    /// Allow for internal testing when no certificates are available but still using https
```
jeremyandrews, line 338: Please use the same format as the above:
`An optional default for not validating https certificates.`

src/config.rs
```diff
@@ -432,6 +437,7 @@ pub enum GooseDefault {
     WebSocketHost,
     /// An optional default for port WebSocket Controller listens on.
     WebSocketPort,
+    AcceptInvalidCerts,
```
jeremyandrews, line 440: Please add a comment, similar to the above:
` An optional default for not validating https certificates.`

src/goose.rs
```diff
@@ -2318,6 +2318,7 @@ pub(crate) fn create_reqwest_client(
         .timeout(Duration::from_millis(timeout))
         // Enable gzip unless `--no-gzip` flag is enabled.
         .gzip(!configuration.no_gzip)
+        .danger_accept_invalid_certs(configuration.accept_invalid_certs)
```
jeremyandrews, line 2321: Please include a comment, for example:

`// Validate https certificates unless `--accept_invalid_certs` is enabled.`

jeremyandrews (changes requested):
Functionally this is great, thanks!

Please add the missing comments and update as flagged.

It would be great to add test coverage as well, but is not required.

raffaeleragni:
Changed the comments

src/config.rs
jeremyandrews: Thanks!

src/goose.rs
jeremyandrews: Whoops, sorry, I think that's actually `--accept-invalid-certs` -- can you test and confirm, and fix if necessary?

jeremyandrews (changes requested):
I think I gave you a bad example, and the `_` needs to be converted to `-`?

raffaeleragni:
Ok, changed

src/goose.rs
jeremyandrews: Thanks!

jeremyandrews (approved):
Thanks for the new feature!

Issue #571, opened 2023-11-16
Something fishy about cookies managment ...


jrouaix:
Hello,

I started using your tool long time ago and just updated to v17 today ... there is something really weird.

- I use auth cookies in some scenarios.
- When I load 1 only user from goose, everything goes fine, my website auth works well, I changed nothing recently
- Then I load multiple users and some scenarios fails because the authentication does not auth them ...
- Then I switch from auth cookie managed by the reqwest client of goose side, to manually crafted from the scenario auth bearer token headers .... and I have no more problems with loads of parallel users.

So I really think something is broken in users isolation, without having any more clue ...

Does that click on your side ? Should I dig deeper ?

jeremyandrews:
Reviewing the [Changelog](https://github.com/tag1consulting/goose/blob/main/CHANGELOG.md), nothing jumps out at me as causing this regression. The only related changed is [#557 ](https://github.com/tag1consulting/goose/pull/557/files), so perhaps you can start by reverting that change and see if it solves your problem? Any further investigation would be greatly appreciated.

alecsmrekar:
I am noticing the same issue. It started popping up after upgrading from Goose 0.17.0 to 0.17.2. I am running a loadtest with 20 users, and one of those users performs two transactions:
- log into a site (runs with `set_on_start`)
- create a piece of content -> this fails due to a 403, inspecting the response, it turned out that the admin sessions wasn't there anymore

PR #575, opened 2023-12-14
test that per-user sessions are working

jeremyandrews:
- validate that with multiple users, each user has a unique and maintained session
 - validate that with multiple users, each user has their own cookie jar
 - the optimization (https://github.com/tag1consulting/goose/pull/557/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R775) in #557 caused the problem reported in #571, specifically because by cloning the reqwest client we're also cloning the cookie jar, which means all GooseUsers end up storing cookies in the same jar. Unfortunately it's not possible to clone the client and then change the cookie jar as this is only exposed when building the client.
 - fixes https://github.com/tag1consulting/goose/issues/571


alecsmrekar (approved):
This fixes the issues I encountered, thanks! Also, nice work on the test coverage.

Issue #429, opened 2022-04-08
Create new "ItemsPerSecond" data type for graph data collection


slashrsm:
At the moment we use data structures like `HashMap<String, TimeSeries<u32, u32>>`, that store per-second counters. It would be better if we would create a dedicated data type, which would also encapsulate all logic related to it. 

More info: https://github.com/tag1consulting/goose/pull/415#discussion_r842661269

PR #577, opened 2023-12-22
Issue 429: Implement ItemsPerSecond

alecsmrekar:
Solves this issue: https://github.com/tag1consulting/goose/issues/429

This PR creates a new struct `ItemsPerSecond` and some accompanying methods that encapsulate the logic of the underlying hash map.

**Note:**
I had to use `#[allow(dead_code)]` because for some reason the compiler would complain about dead code, even though the code is being used.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -23,28 +23,40 @@ impl ItemsPerSecond {
         ItemsPerSecond(Default::default())
     }
 
+    #[inline(always)]
     fn initialize_or_increment(&mut self, key: &str, second: usize, value: u32) -> u32 {
-        if !self.0.contains_key(key) {
-            self.0.insert(key.to_string(), TimeSeries::new());
+        if !self.contains_key(key) {
+            self.insert(key, TimeSeries::new());
         }
         let data = self.0.get_mut(key).unwrap();
         data.increase_value(second, value);
         data.get(second)
     }
 
+    #[inline(always)]
+    fn contains_key(&mut self, key: &str) -> bool {
+        self.0.contains_key(key)
+    }
+
+    #[inline(always)]
     #[allow(dead_code)]
     fn insert(&mut self, key: &str, time_series: TimeSeries<u32, u32>) {
         self.0.insert(key.to_string(), time_series);
     }
 
+    #[inline(always)]
+    #[allow(dead_code)]
     fn len(&self) -> usize {
         self.0.len()
     }
 
+    #[inline(always)]
+    #[allow(dead_code)]
     fn get(&self, key: &str) -> Option<TimeSeries<u32, u32>> {
         self.0.get(key).cloned()
     }
 
+    #[inline(always)]
     fn get_map(&self) -> HashMap<String, TimeSeries<u32, u32>> {
         self.0.clone()
     }
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -39,7 +39,6 @@ impl ItemsPerSecond {
     }
 
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


src/graph.rs
```diff
@@ -15,12 +15,47 @@ use std::collections::HashMap;
...
+struct ItemsPerSecond(HashMap<String, TimeSeries<u32, u32>>);
+
+impl ItemsPerSecond {
+    fn new() -> ItemsPerSecond {
+        ItemsPerSecond(Default::default())
+    }
+
+    fn initialize_or_increment(&mut self, key: &str, second: usize, value: u32) -> u32 {
+        if !self.0.contains_key(key) {
+            self.0.insert(key.to_string(), TimeSeries::new());
```
LionsAd, line 28: Maybe use the insert function here?

src/graph.rs
```diff
@@ -15,12 +15,47 @@ use std::collections::HashMap;
...
+    fn insert(&mut self, key: &str, time_series: TimeSeries<u32, u32>) {
+        self.0.insert(key.to_string(), time_series);
+    }
+
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    fn get(&self, key: &str) -> Option<TimeSeries<u32, u32>> {
+        self.0.get(key).cloned()
```
LionsAd, line 45: What happens if the key does not exist?

src/graph.rs
```diff
@@ -15,12 +15,47 @@ use std::collections::HashMap;
...
+#[derive(Clone)]
+struct ItemsPerSecond(HashMap<String, TimeSeries<u32, u32>>);
+
+impl ItemsPerSecond {
+    fn new() -> ItemsPerSecond {
+        ItemsPerSecond(Default::default())
+    }
+
+    fn initialize_or_increment(&mut self, key: &str, second: usize, value: u32) -> u32 {
+        if !self.0.contains_key(key) {
```
LionsAd, line 27: Maybe adding that as a helper function and using here, would make sense, too?

src/graph.rs
```diff
@@ -15,12 +15,47 @@ use std::collections::HashMap;
...
+        self.0.len()
+    }
+
+    fn get(&self, key: &str) -> Option<TimeSeries<u32, u32>> {
+        self.0.get(key).cloned()
+    }
+
+    fn get_map(&self) -> HashMap<String, TimeSeries<u32, u32>> {
+        self.0.clone()
+    }
```
LionsAd, line 50: Not sure how performance critical this is, but might be good to inline these four functions?

(Might already be done, too)

LionsAd (commented):
Needs some discussion

src/graph.rs
alecsmrekar: It'll just return `None`, it won't blow up

alecsmrekar:
@LionsAd Thanks for the review, I pushed a new commit with the suggested fixes

src/graph.rs
```diff
@@ -15,12 +15,59 @@ use std::collections::HashMap;
...
+        data.get(second)
+    }
+
+    #[inline(always)]
+    fn contains_key(&mut self, key: &str) -> bool {
+        self.0.contains_key(key)
+    }
+
+    #[inline(always)]
+    #[allow(dead_code)]
```
LionsAd, line 42: Don't think this line is needed still.

LionsAd (commented):
Looks great, I think we can remove one more line, but then it's RTBC

LionsAd (approved):
RTBM!

PR #579, opened 2024-01-16
Issue #531: Fix missing impl for TransactionError

alecsmrekar:
This PR addressed issue https://github.com/tag1consulting/goose/issues/531.

The original bug report provided a minimum reproducible code, but I think we're able to shrink that even further to:
```rust
async fn test_case(user: &mut GooseUser) -> TransactionResult {
    let goose = user
        .get("my/url")
        .await?;
    let _ = goose.response?;
    Ok(())
}
```

Trying to compile a loadtest with this code should give the following error:
```
the trait `From<reqwest::error::Error>` is not implemented for `Box<TransactionError>`
```

With this PR, it should compile cleanly.

### DIFF
--- a/src/goose.rs
+++ b/src/goose.rs
@@ -433,6 +433,13 @@ impl From<reqwest::Error> for TransactionError {
     }
 }
 
+/// Auto-convert Reqwest errors to boxed TransactionError.
+impl From<reqwest::Error> for Box<TransactionError> {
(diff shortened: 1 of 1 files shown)


jeremyandrews (approved):
Thanks! Closes https://github.com/tag1consulting/goose/issues/531

Issue #526, opened 2022-10-14
HTML Report adds 0ms Response Time Datapoints


mTsBucy1:
The HTML report is really nice, but one thing that bothers me is that when no request has been made in a given second, the value defaults to 0ms. This is fine for number of requests but doesn't make sense for response time.

For comparison here are the same data, first as it is now and second with the 0ms datapoints removed via [sd](https://github.com/chmln/sd): ` sd '\["\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d",0\.0\],' '' report.html`, echarts handles it just fine.
![with_0ms_datapoints](https://user-images.githubusercontent.com/85930053/195790421-29eeedbc-bf7b-4cc4-895f-e5abd2bdbb95.png)
![without_0ms_datapoints](https://user-images.githubusercontent.com/85930053/195790415-59a68946-8445-409c-8699-2c6fe352c3d4.png)

I've skimmed through the code, and I think changing `TimeSeriesValue::get_graph_value` to return a `Option<T>` instead of a `T` and then filtering based on that might be a viable approach.

mTsBucy1:
I will close this issue as I've noticed several problems with my proposed solution. Even though for some cases the graphics look much better, they tend to look worse. Thus, this shouldn't be a change in the code. Instead, I recommend using the `sd` command stated above if you desire.

jeremyandrews:
Re-opening. The reported issue sounds like something we should be able to fix.

mTsBucy1:
Here an example of what I meant with "tend to look worse". The spans between the peaks are spanned with distracting lines that interpolate data that shouldn't be interpolated. In those (common) cases, setting them to 0 gives nicer graphics while the request log metrics are unaffected.

The only solution I see is to revamp the graph to not interpolate between datapoints which are too far away, which seems overkill. There are other shortcomings of the html request that should be addressed with higher priority than this.

![image](https://user-images.githubusercontent.com/85930053/199945500-6f5167b4-1241-40be-a7c9-572596d3065f.png)

PR #581, opened 2024-01-26
Issue 526: Fix blank graph bits

alecsmrekar:
https://github.com/tag1consulting/goose/issues/526

This is a relatively simple change that prevents the chart from rendering 0 values.

The base change is in `TimeSeriesValue`, where `get_graph_value` is converted to detect zeros, if it does, it returns a `None`. Converting the return value from `U` to `Option<U>` then allows us to filter out blank values when constructing chart data in `add_timestamp_to_html_graph_data`.

To test this, just run a few different transactions, and maybe add `set_wait_time` to make sure you have some seconds where nothing happened. Then open up the HTML report and looks at the pretty chart.

It's worth noting that the person who opened the original issue complained that this solution actually makes graphs looks worse, which might very well be true in some circumstances. A solution might be to adjust the chart JS a bit so that it adds dots to where data points are, so the user is not confused where requests actually happened. Example:
image

Another alternative would be to switch to a scatter plot. Which for me personally is a lot harder to read data from at a quick glance, although is solves the issue of connecting dots that are too distant.
image

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -557,7 +557,7 @@ impl<'a, T: Clone + TimeSeriesValue<T, U>, U: Serialize + Copy + PartialEq + Par
     /// it belongs to.
     fn add_timestamp_to_html_graph_data(
         &self,
-        data: &[U],
+        data: &[Option<U>],
         started: DateTime<Utc>,
     ) -> Vec<(String, U)> {
         data.iter()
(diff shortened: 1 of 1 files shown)


jeremyandrews (approved):
This is a clean and simple improvement. We can explore your proposed improvements as followups if necessary. Thank you!

PR #595, opened 2024-06-12
Make GooseUser and GooseUserData cloneable

S7evinK:
As the title describes, this implements `Clone` and `Debug` for  `GooseUser` and `GooseUserData`.

In our use-case we need to pass `GooseUser` to a SDK which is doing all the network requests.


jeremyandrews:
Interesting, thanks for the PR.  Be sure to run through `cargo fmt` which enforce our coding standards.

In your use case, does Goose provide you any metrics? When you clone and pass off the GooseUser, is something other than Reqwest making the requests? Why not set up a proxy? I'm trying to better understand your use case, and the implications of this change.

S7evinK:
Thanks for the quick reply. :)

Goose provides metrics in our case which we need to sanitize, as having metrics for each "room" (e.g. `/_matrix/client/v3/rooms/{roomId}/join`) doesn't make sense as we care about the `/join` endpoint only.
The underlying [matrix-rust-sdk](https://github.com/matrix-org/matrix-rust-sdk) uses `reqwest` as well, but if we pass it the client provided by `GooseUser`, we're losing the metrics.

jeremyandrews (approved):
I don't see any inherent downside to making things cloneable and debug formatted, thank you!

Note that you can name requests to avoid different paths from creating a different metric. For example, see how Goose Eggs does this with static elements:
https://github.com/tag1consulting/goose-eggs/blob/428939d1634acd3224d1f635433eee56ab98e605/src/lib.rs#L1076

Issue #619, opened 2025-05-13
Is there a problem with getting errors in the report


chengguotao:
in the file metrics->common ,
pub fn prepare_data(options: ReportOptions, metrics: &GooseMetrics) -> ReportData，
the function fetch errors: metrics.errors.is_empty()
.then(|| metrics.errors.values().collect::<Vec<_>>()),
Should it be changed to errors: (!metrics.errors.is_empty())
.then(|| metrics.errors.values().collect::<Vec<_>>())？

PR #620, opened 2025-07-28
Fix error graph in HTML report

trevorbonas:
Currently, the HTML report does not display an error summary. This is due to an incorrect boolean value in `src/metrics/common.rs`. This PR fixes this, returning the error summary.

Closes: #619


trevorbonas:
@jeremyandrews `cargo fmt` run and changes pushed. Thanks for reviewing this PR!

PR #644, opened 2025-08-17
Implement the external tool PDF generation approach

LionsAd:
This was a quick test to see how it would look like.

We can certainly remove the last commit and bring back the pure linked-in crate implementation - if we really think this is better, but things like the "temporary override log level" smell still make me feel that the external tool invocation is better.

It's also more powerful as we can now also do `--margin` and we could add more -- without cluttering the main goose binary options (proper isolation of concerns).

Only the last commit actually removes something. Everything else is just adding the commands.


jeremyandrews (changes requested):
This is the very thing I'd like to avoid. It's much more complicated to install and use, and IMO will mean fewer people find and use the PDF report functionality. I feel like we're abstracting and complicating things for no real gain.

In regards to this POC, I tried simply creating a PDF report -- the load test ran but then at the end I got this error:
```
Error: InvalidOption { option: "--report-file", value: "report.pdf", detail: "PDF generation requires the 'goose-pdf' companion tool. Install with: cargo install goose --features pdf-reports" }
```

This already is a usability regression. Goose is a generally a temporary-tool that's installed to run some tests in the cloud on an instance that doesn't stick around long-term. So any manual dependencies are steps that will have to be run over and over each time a new goose server is spun up.

And, this error appears After the load test is run, which is very problematic: if this was the only report we were generating, we just lost all our data. Being decoupled as it is, this will always be a risk.

Attempting to follow the provided directions fails:
```
% cargo install goose --features pdf-reports
    Updating crates.io index
  Downloaded goose v0.18.1
  Downloaded 1 crate (5.1MiB) in 2.33s
error: there is nothing to install in `goose v0.18.1`, because it has no binaries
`cargo install` is only for installing programs, and can't be used with libraries.
To use a library crate, add it as a dependency to a Cargo project with `cargo add`.
```

Eventually I figured out how to at least build it:
```
% cargo build --release --features pdf-reports --bin goose-pdf
```

This is another big usability regression for me: you have to specify the `pdf-reports` feature flag; in the original version it auto-compiled what it needed if a report was requested. It just worked.

Running after building this again failed, so I copied the binary into my path:
```
% which goose-pdf
/usr/local/bin/goose-pdf
```

And it still failed. 

The only way I was able to generate a PDF was to first generate an HTML report, then manually run ./goose-pdf on the html report. It worked. But IMO this completely misses the point: with this PR we're optimized for abstraction (the ability to replace with a different PDF generator) at the cost of usability (it's hard to get this working; and it's something people will have to do over and over as goose tends to be used on temporary cloud vms).

Also note that the --margin feature doesn't work correctly. Yes, you can increase the margin, but when you do pieces of the report simply vanish.

I strongly prefer the implementation in https://github.com/tag1consulting/goose/pull/629 because it solves the problem I'm trying to solve: simple and reliable pdf report generation. Using external binaries that have to be built and/or installed and which may fail are not solving the problem.

PR #648, opened 2025-08-25
Fix gemini PR review workflow security with pull_request_target

LionsAd:
## Summary
- Fixes the Gemini PR integration that was "pretty much useless" by switching from `pull_request` to `pull_request_target`
- Implements secure checkout pattern to prevent malicious code execution
- Maintains Gemini's ability to analyze code context while keeping the workflow secure

## Security Improvements
- **Changed trigger**: `pull_request` → `pull_request_target` for access to secrets and write permissions
- **Secure checkout**: Only checkout base branch code, never untrusted PR code
- **Updated conditions**: Job conditions now match the new trigger type
- **Preserved functionality**: Gemini can still use `cat`, `head`, `tail`, `grep` on secure base branch files

## Why This is Safe
The workflow now follows GitHub's security best practices:
1. No untrusted PR code is ever checked out or executed
2. File access commands only work on the trusted base branch
3. PR changes are analyzed via `gh pr diff` API calls only
4. All existing author association checks remain in place

## Test Plan
- [ ] Verify workflow triggers on new PRs from collaborators
- [ ] Confirm Gemini can access base branch files for context
- [ ] Ensure PR analysis works through API calls
- [ ] Check that reviews are posted successfully

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/gemini-pr-review.yml
+++ b/.github/workflows/gemini-pr-review.yml
@@ -1,7 +1,7 @@
 name: '🧐 Gemini Pull Request Review'
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - 'opened'
       - 'reopened'
@@ -41,7 +41,7 @@ jobs:
     if: |-
       github.event_name == 'workflow_dispatch' ||
       (
-        github.event_name == 'pull_request' &&
+        github.event_name == 'pull_request_target' &&
         contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association)
       ) ||
       (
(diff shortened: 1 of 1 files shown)


jeremyandrews (approved):
I examined the workflow file differences, compared the original pull_request trigger with the new pull_request_target approach, verified that all conditional checks were consistently updated (lines 44, 84), confirmed the secure checkout implementation with base branch reference, and validated that the security model prevents execution of untrusted PR code while maintaining API-based analysis capabilities.

The implementation is technically sound and follows GitHub's recommended security patterns for pull_request_target workflows. The changes are minimal, focused, and address the core issue that rendered the original workflow non-functional due to lack of necessary permissions. The security boundary is properly maintained by analyzing PR changes only through API calls rather than executing untrusted code, while the base branch checkout ensures only trusted code is accessible to the workflow. I recommend merging this PR as it enables the intended Gemini integration functionality while implementing proper security controls.

PR #668, opened 2026-01-03
chore: update httpmock to 0.8, avoid deprecated function names

tstenner:
Basically what the title says. Since httpmock 0.8, the `.hits()` and `.assert_hits()` calls were deprecated so I replaced them.


jeremyandrews (approved):
Thank you!

PR #669, opened 2026-01-04
chore: update reqwest to 0.13

tstenner:
This PR updates reqwest to 0.13 which uses rustls by default.

I have observed the per-connection overhead to be greater than before because HTTP/2 + post-quantum cryptography is advertised / accepted depending on the remote server.
This would have likely happened with `native-tls` as well, because ALPN is enabled for it by default as well so IMHO it doesn't make sense to add another feature flag for `native-tls`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
  - [#629](https://github.com/tag1consulting/goose/pull/629) add `--pdf-print-html` option to generate printer-friendly HTML optimized for PDF conversion; provides two-step PDF workflow without requiring Chromium dependencies; add `--pdf-timeout` option for configurable Chrome timeout in direct PDF generation (10-300s, default: 60)
  - [#663](https://github.com/tag1consulting/goose/pull/663) add response time breakdowns grouped by HTTP status code in CLI and HTML/markdown reports; breakdowns only appear when a request returns multiple different status codes
  - [#657](https://github.com/tag1consulting/goose/pull/657) add comprehensive Basic Authentication example demonstrating three approaches: custom client with default headers (recommended), helper function per-request, and manual per-request; includes flexible credential configuration and complete documentation
+ - [#669](https://github.com/tag1consulting/goose/pull/669) update reqwest to 0.13, switching the default TLS backend from native-tls to rustls. HTTP/2 and post-quantum cryptography are now advertised by default, which may increase connection handshake overhead when a computationally expensive cipher suite is negotiated.
+   **Breaking changes**: Certificate validation now uses bundled rustls/webpki roots instead of the OS-native certificate store, which may affect environments with custom system CAs. `--accept-invalid-certs` now works unconditionally (previously it was silently ignored unless the `rustls-tls` feature was enabled).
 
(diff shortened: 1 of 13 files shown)


src/docs/goose-book/src/config/rustls.md
```diff
@@ -1,8 +0,0 @@
-# RustLS
```
jeremyandrews: `src/docs/goose-book/src/SUMMARY.md` still references this file on line 46:
```
- [RustLS](config/rustls.md)
```
Deleting this file without updating SUMMARY.md will leave a broken link in the goose-book. Please remove that entry from the table of contents.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
  - [#629](https://github.com/tag1consulting/goose/pull/629) add `--pdf-print-html` option to generate printer-friendly HTML optimized for PDF conversion; provides two-step PDF workflow without requiring Chromium dependencies; add `--pdf-timeout` option for configurable Chrome timeout in direct PDF generation (10-300s, default: 60)
  - [#663](https://github.com/tag1consulting/goose/pull/663) add response time breakdowns grouped by HTTP status code in CLI and HTML/markdown reports; breakdowns only appear when a request returns multiple different status codes
  - [#657](https://github.com/tag1consulting/goose/pull/657) add comprehensive Basic Authentication example demonstrating three approaches: custom client with default headers (recommended), helper function per-request, and manual per-request; includes flexible credential configuration and complete documentation
+ - [#669](https://github.com/tag1consulting/goose/pull/669) updates reqwest to 0.13. This **changes the default TLS implementation to rustls** with HTTP/2 + post-quantum cryptography support advertised to the server by default.
```
jeremyandrews: Minor: other entries in this section use imperative mood ("add ...") — this should be "update" rather than "updates" for consistency.

I'd also suggest tightening this up a bit, something like:

```suggestion
 - [#669](https://github.com/tag1consulting/goose/pull/669) update reqwest to 0.13, switching the default TLS backend from native-tls to rustls. HTTP/2 and post-quantum cryptography are now advertised by default, which may increase connection handshake overhead when a computationally expensive cipher suite is negotiated.
```

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
...
+   **Behavioral change**: `native-tls` features like certificate validation against a custom system CA store will be disabled and `--accept-invalid-certs` available. It also might increase the connection handshake overhead compared to goose 0.18 when a newly available computationally expensive cipher suite is negotiated.
```
jeremyandrews: This sentence is hard to parse — "will be disabled and `--accept-invalid-certs` available" reads as though both are being disabled. Consider something like:

```suggestion
   **Breaking changes**: Certificate validation now uses bundled rustls/webpki roots instead of the OS-native certificate store, which may affect environments with custom system CAs. `--accept-invalid-certs` now works unconditionally (previously it was silently ignored unless the `rustls-tls` feature was enabled).
```

tstenner:
Thanks, these suggestions are way better than mine. I applied them, removed the stray link in the docs and squashed the commits.

